PR #2, opened 2015-04-22
The other pieces we need to create an artifact

jennysivapalan:
build.sbt and plugins settings to get an artifact build.


tudorraul:
:+1:

PR #4, opened 2015-04-23
Frontend tools

marchibbins:
Include SASS source files, grunt to watch and compile them, bower to manage packages.


.gitignore
```diff
@@ -29,3 +28,9 @@ project/plugins/project/
 # Scala-IDE specific
 .scala_dependencies
 .worksheet
+
+# frontend
+*.css.map
+.sass-cache
+bower_components
```
davidrapson: As you're starting a new project, it might be worth talking to @OliverJAsh or @theefer about current thoughts about Bower. Currently with Bower you need to consider committing the dependencies due to lack of version locking, but that's not ideal so there might be an another option. Current preference seems to be for `jspm` but not sure how suitable that is for CSS dependencies.

PR #5, opened 2015-04-24
Migrate existing Digital pack page

marchibbins:
Styles and mark-up verbatim from existing version.

Simple object to populate edition differences. @jennysivapalan @rtyley is it worth making these more formal at this point? Presumably full models to come, just migrating templates and routes at this point.


app/assets/stylesheets/bower.json
```diff
@@ -3,11 +3,14 @@
   "dependencies": {
     "guss": "3.3.0",
     "pasteup-buttons": "0.4.10",
+    "pasteup-forms": "6.0.0",
```
davidrapson: Exciting, let's see how this goes.

davidrapson:
Any reason generate source is being committed at the moment?

marchibbins:
These PRs are migrating what we have in our current [Python GAE app](https://github.com/guardian/subscriptions) where we deploy local application files from the command line. Got little time to both get parity on existing routes and add new templates, when we've got breathing room there should be some build scripts to run the compilation etc - something to resolve along with the Bower question.

jennysivapalan:
I think the routes are fine, we can come back and add more logic around editions later.

davidrapson:
:+1: No problem, just wanted to check.

PR #6, opened 2015-04-24
Only authorised users can log on to instances

jennysivapalan:
Similar to https://github.com/guardian/membership-frontend only authorised users can log on to the instances. Theses are based on user github public keys.


rtyley:
:+1:

PR #16, opened 2015-04-29
Add nginx support for development purposes

tudorraul:


theefer:
Not sure if it applies to this, but you may want to have a look at https://github.com/guardian/dev-nginx which is a generic way to do something very similar. We use it on all editorial tools.

rtyley:
:+1: to the current config - happy to revisit `dev-nginx` later.

PR #17, opened 2015-04-29
Digital pack forms checkout javascript

marchibbins:
First layer of javascript for payment forms. Add Grunt and Require. Introduces modern browser check and JS app structure.

What this includes:
- Address lookup
- Manual address override
- Form fieldset expand/collapse
- Direct Debit legal text show/hide
- Populate details for review

Not included:
- Generic (non-checkout) form JS, i.e. validation of field requirement or type. This should be migrated from [Membership](https://github.com/guardian/membership-frontend) to [pasteup-forms](https://github.com/guardian/pasteup-forms), or temporarily duplicated here

As discussed before, compiled JS is included until build scripts are in place.


tudorraul:
:+1:

PR #21, opened 2015-05-01
Homepage static content updates

tudorraul:


rtyley:
Regarding the campaign marker links, doing a quick grep with:

```
git grep 'INTCMP.*"'
```

on  https://github.com/guardian/subscriptions-frontend & https://github.com/guardian/subscriptions, the campaign codes look to match, so :+1:

tudorraul:
what about the krux code?

rtyley:
I don't know much about Krux, so can't really speak to that...

app/views/fragments/analytics/krux.scala.html
```diff
@@ -1,9 +1,10 @@
...
 <script class="kxct" data-id="Jglpp88U" data-timing="async" data-version="1.9" type="text/javascript">
-    window.Krux || ((Krux=function() { Krux.q.push(arguments) }).q=[]);
-    (function(){
-        var k=document.createElement('script');k.type='text/javascript';k.async=true;
-        var m,src=(m=location.href.match(/\bkxsrc=([^&]+)/))&&decodeURIComponent(m[1]);
-        k.src = /^https?:\/\/([a-z0-9_\-\.]+\.)?krxd\.net(:\d{1,5})?\//i.test(src) ? src : src === "disable" ? "" : (location.protocol==="https:"?"https:":"http:")+"//cdn.krxd.net/controltag?confid=Jglpp88U";
-        var s=document.getElementsByTagName('script')[0];s.parentNode.insertBefore(k,s);
-    }());
-</script>
+  window.Krux||((Krux=function(){Krux.q.push(arguments)}).q=[]);
```
davidrapson: @tudorraul Bit late on the feedback here as it's merged, but if the new subscriptions is going to be over HTTPS you'll want to make sure you check for mixed-content warnings from Krux as we have had a lot of problems with that on Membership. It's OK once configured, but something to be aware of.

app/views/fragments/head.scala.html
```diff
@@ -24,4 +24,4 @@
     <link rel="shortcut icon" type="image/png" href="@routes.Assets.at("images/favicons/32x32.ico")"/>
 }
 
-@* <script src="//cdn.optimizely.com/js/293455139.js"></script> *@
+<script src="//cdn.optimizely.com/js/293455139.js"></script>
```
davidrapson: As this is in the head (I get why A/B testing scripts are in the head) you should definitely add `async defer` attributes to this as in the currently could cause SPOF errors. See https://github.com/guardian/membership-frontend/pull/421

app/views/fragments/analytics/krux.scala.html
tudorraul: Thanks for the heads up. It looks like all krux calls are over https which return 200. Also, there is no security warning from the browser. Thanks for flagging.

PR #24, opened 2015-05-15
Next digital pack

tudorraul:
current page: http://subscribe.theguardian.com/digital
updated page:
mobile
tablet
desktop
wider desktop


app/assets/stylesheets/modules/digital.scss
```diff
@@ -1,42 +1,163 @@
-
 /* Digital pack */
-.digital-cta {
-    padding-top: $gs-baseline*2;
-    padding-bottom: $gs-baseline*3;
+
+.digital-pack {
```
davidrapson: Is there a more component driven way of doing this? It's a good idea to avoid scoping styles based on a body class name.

app/views/main.scala.html
```diff
@@ -1,11 +1,11 @@
-@(title: String, footer: Html = Html(""))(content: Html)
+@(title: String, footer: Html = Html(""), bodyClassName: String = "")(content: Html)
```
davidrapson: I'd try and avoid setting custom classnames on the body if you can.

Gemfile
```diff
@@ -0,0 +1,8 @@
+source "https://rubygems.org"
+
+gem "sass", "~> 3.3.14"
```
davidrapson: As you are starting this fresh I'd suggest trying [node-sass](https://github.com/sass/node-sass) via [grunt-sass](https://github.com/sindresorhus/grunt-sass) rather than the default ruby sass as it uses the latest `libsass` which get much better compile times, and is the recommended version now.

app/assets/stylesheets/modules/digital.scss
```diff
@@ -1,42 +1,163 @@
...
+        margin-bottom: $gs-baseline*2;
+    }
+}
+
+// CONTAINERS
+
+$left-column: gs-span(2);
+
+.digital-pack {
+    .gs-container {
```
davidrapson: Is there a simpler way to think about this? Modifying the global `gs-container` class based on a body class feels a little awkward.

app/views/digitalpack/info.scala.html
```diff
@@ -1,88 +1,110 @@
...
 
-@main(if(edition.id == "uk") {
-    "Digital | Subscriptions | The Guardian"
-} else {
-    s"Digital - ${edition.name} | Subscriptions | The Guardian"
-}) {
+@main(
+    title = if(edition.id == "uk") "Digital | Subscriptions | The Guardian" else s"Digital - ${edition.name} | Subscriptions | The Guardian",
+    bodyClassName = "digital-pack") {
+<div class="gs-container dp-container">
```
davidrapson: What's the thinking of mixing `dp-container` with `gs-container`, feels like there's a lot of overlap / conflict. Is there a simpler way of writing this?

davidrapson:
There's a few things I'm slightly unclear about: not a big fan of using body classes to style components and I'm a bit unsure about the mixing of `gs-container` and your own guss classes `dp-container` / `dp-row`

Few other minor comments. I don't know the full context of this work so feel free to disregard if you think I'm off the mark

It might just be that screenshot but the breakpoints for the small screen version look a little off, are the benefits only meant to take up 50% of the space?

:+1: once those few things have been looked at.

app/assets/stylesheets/modules/digital.scss
tudorraul: the digital pack page needs to look different at the moment, we'll be changing this page first and rollout the change gradually to the whole site.

app/views/main.scala.html
tudorraul: it's because the digital pack page needs to look different at the moment

app/views/digitalpack/info.scala.html
tudorraul: it's because the digital pack page needs to look different at the moment

app/assets/stylesheets/modules/digital.scss
davidrapson: No problem then.

davidrapson:
:+1: LGTM

PR #26, opened 2015-05-20
Caching headers

tudorraul:
Add caching headers to all responses.

Added HTTP headers
- `Cache-Control`
- `Expires`
- `Date`

### DIFF
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -6,20 +6,16 @@ import play.api.mvc._
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
-trait CommonActions {
-
-  import CommonActions._
-
-  val noCacheAction = resultModifier(noCache)
-
-  val googleAuthAction = OAuthActions.AuthAction
-
-  val googleAuthenticatedStaffAction = noCacheAction andThen googleAuthAction
-
-  val cachedAction = resultModifier(Cached(_))
-}
-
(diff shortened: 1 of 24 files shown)


app/actions/CommonActions.scala
```diff
@@ -8,11 +9,14 @@ import scala.concurrent.Future
 trait CommonActions {
 
   import CommonActions._
+
   val noCacheAction = resultModifier(noCache)
```
rtyley: Small style nit but I think the style guide says "Constant names should be in upper camel case", ie `NoCacheAction` etc:

http://docs.scala-lang.org/style/naming-conventions.html#objects

app/controllers/DigitalPack.scala
```diff
@@ -14,23 +15,23 @@ object DigitalEdition {
...
     Ok(views.html.digitalpack.info(DigitalEdition.AU))
   }
 
-  def selectCountry = Action {
+  def selectCountry = CachedAction {
     Ok(views.html.digitalpack.country())
   }
 
-  def handleCountry = Action(parse.tolerantFormUrlEncoded) { request =>
+  def handleCountry = CachedAction(parse.tolerantFormUrlEncoded) { request =>
```
rtyley: I don't think we should be caching the response to this POST request! Unless you're sure that Fastly is varying the response based on request body.

rtyley:
The existing way of handling POST request is bizarre, but for the caching stuff :+1:

PR #28, opened 2015-05-27
Updated README notes for starting app on dev machines.

jennysivapalan:
I updated notes based on trying to start this locally. I guessed we'd need the point around AWS credentials (I already had them from Membership). I will remove this if we don't think it is necessary.


rtyley:
:+1:

PR #30, opened 2015-06-01
Added Sentry configuration.

jennysivapalan:
Based on implementation from Membership: https://github.com/guardian/membership-old/pull/876/files (N.B. note this a reference to the old repo).

There is new project / team in Sentry. @rtyley you should have access straight away but @tudorraul you'll need to become a member. I don't have powers to invite you but @rtyley might be.

I've updated the prod configuration with the sentry.dsn value.

:tanabata_tree: :page_with_curl:

### DIFF
--- a/app/Global.scala
+++ b/app/Global.scala
@@ -1,4 +1,10 @@
 import filters.CheckCacheHeadersFilter
+import monitoring.SentryLogging
+import play.api.Application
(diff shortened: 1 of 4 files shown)


rtyley:
:+1:

playframework/playframework#3780 - flexible logging of errors (ie user ids!) may be useful later.

PR #31, opened 2015-06-03
Integration with Identity and Salesforce services

tudorraul:
The two services are integrated in the checkout process, following this logic:

Firstly the user's identity is resolved using Identity API, when this succeeds, a new Salesforce user is created or an existing one is updated with the user's personal details (name, email, address).

The user's identity is resolved either by looking up the logged in user (by cookie) in the ID API, looking up the user by the email address he or she types in the checkout form or creating a new guest user if the email address is not present in the ID API.

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -1,7 +1,7 @@
 package controllers
 
 import actions.CommonActions._
-import model.{AddressData, PaymentData, SubscriptionData}
+import model.{AddressData, PaymentData, PersonalData, SubscriptionData}
 import play.api.mvc._
 import services.CheckoutService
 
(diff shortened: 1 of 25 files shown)


app/controllers/Checkout.scala
```diff
@@ -1,13 +1,67 @@
...
 
 object Checkout extends Controller {
 
-  def renderCheckout = GoogleAuthenticatedStaffAction(Ok(views.html.checkout.payment()))
+  import play.api.data.Forms._
+  import play.api.data._
+
+  val subscriptionForm = Form(mapping(
+    "" -> mapping(
+      "first" -> text,
```
jennysivapalan: Why do we have nested mapping and why does it start `"" -> mapping`?

app/controllers/Checkout.scala
```diff
@@ -1,13 +1,67 @@
...
+  )(SubscriptionData.apply)(SubscriptionData.unapply))
+
+  val renderCheckout = GoogleAuthenticatedStaffAction(Ok(views.html.checkout.payment(subscriptionForm)))
+
+  val handleCheckout = NoCacheAction.async { implicit request =>
+    val formWithData = subscriptionForm.bindFromRequest
 
-  def handleCheckout = GoogleAuthenticatedStaffAction(Redirect(routes.Checkout.thankyou()))
+    if (formWithData.hasErrors) {
+      println(s">>>${formWithData.errors}")
```
jennysivapalan: Can remove println.

app/controllers/Checkout.scala
```diff
@@ -1,13 +1,67 @@
...
+      Future.successful(Ok(views.html.checkout.payment(formWithData)))
+    }
+    else {
+      val subscriptionData: SubscriptionData = formWithData.get
+      println(s">>>$subscriptionData")
```
jennysivapalan: Can remove println.

app/services/CheckoutService.scala
```diff
@@ -0,0 +1,54 @@
...
+        })
+
+    for {
+      idUser <- lookupUser
+      sfUser <- createSFUser(idUser)
+    } yield {
+      println(s">>>userID:$idUser")
+      println(s">>>sfUser:$sfUser")
+      sfUser.right.map(_ => "AB123456")
+    }
```
jennysivapalan: Can remove printlns.
What's are we suppose to be yielding here as it looks like dummy data?

tudorraul: The end of the checkout process is the SubscriberID, to be provided by Zuora.

app/controllers/Checkout.scala
tudorraul: to group data that belongs to different model classes

app/services/CheckoutService.scala
```diff
@@ -0,0 +1,54 @@
...
+    val preserveFailureReason = (ex: CheckoutException) => Future.successful(Left(ex))
+
+    val lookupUser: Future[Either[CheckoutException, IdUser]] =
+      scGuUCookie.map(
+        identityService.userLookupByScGuU(_)
+          .map(_.toRight(InvalidLoginCookie)))
+        .orElse(
+          Some(identityService.userLookupByEmail(subscriptionData.personalData.email)
+            .filter(_.isDefined)
+            .recoverWith { case _: NoSuchElementException => identityService.registerGuest(subscriptionData.personalData) }
```
jennysivapalan: Small one but could the `case` statement be pushed to a new line. The registerGuest is an important part of the recover so I think will be easier to read on a new line.

tudorraul: I kept the case on the same line for brevity since there is only one case. I can put the `registerGuest` call on the next line for readability. Will that work in your opinion?

app/controllers/Checkout.scala
jennysivapalan: Can we split them into three separate variables and then link them in the `subscrioptionForm`? I think it will be easier to read and understand later.

app/services/CheckoutService.scala
jennysivapalan: I think a new line for the case statement onwards would be easier to read.

jennysivapalan:
@rtyley and I have looked over the PR this week. We think it would be easier to break this down into smaller parts first concentrating on just Identity to start with. Happy to use this PR as a starting point but it will be easier to create a new branch for each part.

tudorraul:
closing this as we've decided to split it into two PRs

PR #32, opened 2015-06-04
Re-instate Play v2.4.0

rtyley:
We stepped down with https://github.com/guardian/subscriptions-frontend/commit/1031c2159 because `play-googleauth` wasn't available at that time, but https://github.com/guardian/play-googleauth/pull/24 took care of that.

For unmerged stuff (ie https://github.com/guardian/subscriptions-frontend/pull/31) that depends on `membership-common`, v0.66 takes care of that with https://github.com/guardian/membership-common/pull/67.

cc @tudorraul

### DIFF
--- a/test/ApplicationSpec.scala
+++ b/test/ApplicationSpec.scala
@@ -9,11 +9,6 @@ import play.api.test.Helpers._
 class ApplicationSpec extends Specification {
 
   "Application" should {
-
-    "send 404 on a bad request" in new WithApplication {
-      route(FakeRequest(GET, "/boum")) must equalTo(None)
-    }
-
     "render the index page" in new WithApplication {
       val home = route(FakeRequest(GET, "/")).get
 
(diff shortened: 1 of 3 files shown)


test/ApplicationSpec.scala
```diff
@@ -9,11 +9,6 @@ import play.api.test.Helpers._
 class ApplicationSpec extends Specification {
 
   "Application" should {
-
-    "send 404 on a bad request" in new WithApplication {
```
tudorraul: what was wrong here?

jennysivapalan:
:+1: :up:

test/ApplicationSpec.scala
rtyley: The '404 on a bad request' test (a sample test supplied by the old version of Play) doesn't work anymore - slightly different behaviour of the Playframework - so have just deleted it.

PR #33, opened 2015-06-08
Enable JavaScript code map

tudorraul:
This adds the JS map to our deploy package and enables us to debug future possible problems on `subscribe.theguardian.com`. 

Note: The generated assets need to be removed from VS in the future and get created at the same time as the deploy package. This file is one of the generated assets.


davidrapson:
:+1: but agree with removing the compiled assets from version control soon.

PR #35, opened 2015-06-15
Improve client-side build process

afiore:
Improves client-side build process (paired with @davidrapson)

- **Moved assets from `app/assets` to `assets/`**
- Remove minified JS/CSS assets from version control
- Removed `jspm` for now as it was installed but not hooked up to anything so better to stick to working `requirejs` + `bower` combo for now. We can revisit this at a later date
- Cleans up `Gruntfile`, upgrades to newer `grunt-sass` which removes the `ruby` dependency.
- Setup grunt tasks for copying, cleaning and hashing assets
- Add `util.Asset` for reading `assets.map`
- Move Omniture JS into assets and copy it to dist
- Move images into assets
- Updated links to assets
- Adds `grunt-tc` script for TeamCity

PR #38, opened 2015-06-17
Make sure Grunt hashes source maps also for CSS files

afiore:
Fixes a minor issue spotted while verifying #35

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -71,9 +71,8 @@ module.exports = function(grunt) {
             staticfiles: {
                 files: [{
                     src: [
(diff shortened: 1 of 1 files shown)


davidrapson:
:+1: LGTM

afiore:
Checked on live site but unfortunately the issue still persists. While `main.css.map` is now there, it looks like in order to have source maps work also for CSS we need to make available the original source files under `assets/dist`.

screen shot 2015-06-17 at 18 00 03 1 2

davidrapson:
@afiore Not sure that should be the case, no need to publish the source files too. In the case of CSS sourcemaps having them in PROD is far from critical so you could happily just keep them for DEV.

PR #39, opened 2015-06-18
Identity integration

tudorraul:
Ensure Checkout creates Identity account.

- Identity account created with the correct user details - First name, Last name, Email address, address, and email preferences

- If user is already signed in the personal details should be pre-filled in the first part of the form (your details page)

PR #40, opened 2015-06-18
Add 404 and 500 error pages

afiore:
Implement 404 and 500 error pages by reusing the ErrorHandler object from  Membershops.
I have verified the 404 page works fine (see screenshot), however I haven't managed to reproduce the 500 error locally. Is there a way to run Play locally in production mode?

screen shot 2015-06-18 at 11 25 11

### DIFF
--- a/app/views/error500.scala.html
+++ b/app/views/error500.scala.html
@@ -1,0 +1,17 @@
+@(ex: Throwable)
+@main("Internal Server Error") {
+    <div class="gs-container">
+        @fragments.page.header("Error 500", Some("There was an error on this page"))
+
+        <div class="row">
+            <p>
+              Sorry, there was an error on this page. We're aware of the problem. If this error persists, please contact
+              support on <a href="mailto:subscriptions.dev@@theguardian.com">subscriptions.dev@@theguardian.com</a>.
+            </p>
+            @ex match { case err: play.api.UsefulException =>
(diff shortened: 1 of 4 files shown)


app/views/error500.scala.html
```diff
@@ -0,0 +1,17 @@
+@(ex: Throwable)
+@main("Internal Server Error") {
+    <div class="gs-container">
+        @fragments.page.header("Error 500", Some("There was an error on this page"))
+
+        <div class="row">
+            <p>
+              Sorry, there was an error on this page. We're aware of the problem. If this error persists, please contact
+              support on <a href="mailto:subscriptions.dev@@theguardian.com">subscriptions.dev@@theguardian.com</a>.
```
rtyley: We'll probably want to hcange that email address eventually...

rtyley:
:+1:

app/views/error500.scala.html
jennysivapalan: Agreed - we should get similar emails set up to membership and figure out who needs to be on the end of those.

afiore:
404 verified on prod.

screen shot 2015-06-18 at 16 42 13

rtyley:
The 500 page (thanks #39 !) looks a little rough, with the error code being pushed out over the right hand side of the page:

image

PR #42, opened 2015-06-18
Replace obsoleted `-mem` parameter with `-J-Xmx512m`

rtyley:
Apparently `sbt-native-packager` removed the `-mem` parameter with https://github.com/sbt/sbt-native-packager/commit/62362c2377 - so if you look in a subscriptions box, you see this in `/subscriptions/stdout.log`:

```
!! WARNING !! -mem option is ignored. Please use -J-Xmx and -J-Xms
```

https://github.com/sbt/sbt-native-packager/issues/82#issuecomment-76723624

### DIFF
--- a/conf/upstart.conf
+++ b/conf/upstart.conf
@@ -23,5 +23,5 @@ limit nofile 65536 65536
 respawn
 
 script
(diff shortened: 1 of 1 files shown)


tudorraul:
:+1:

PR #45, opened 2015-06-19
Passing form to the template rather than individual values

jennysivapalan:
As suggested https://github.com/guardian/subscriptions-frontend/pull/39#issuecomment-113481106 I have changed the payment template to take a subscriptions form rather than individual fields.

Also added name references to the mappings and onErrors for handleCheckout (but not being rendered).

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -3,6 +3,7 @@ package controllers
 import actions.CommonActions._
 import com.gu.identity.play.PrivateFields
 import model.{AddressData, PaymentData, PersonalData, SubscriptionData}
+import play.api.data.Form
 import play.api.mvc._
 import services.{AuthenticationService, CheckoutService, IdentityService}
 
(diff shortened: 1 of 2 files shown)


app/controllers/Checkout.scala
```diff
@@ -16,25 +17,32 @@ object Checkout extends Controller {
...
-        views.html.checkout.payment(
-          idUserData(_.firstName),
-          idUserData(_.secondName),
-          idUserOpt.map(_.primaryEmailAddress),
-          idUserData(_.billingAddress1),
-          idUserData(_.billingAddress2),
-          idUserData(_.billingAddress3),
-          idUserData(_.billingPostcode))
+      val form = SubscriptionsForm().copy(
+        data =  (
```
afiore: Wow! That `Option.++` is clever, I didn't know it existed!

I think this code could be even more compact if we were to produce the tuples directly in the `idUserData` function:

```scala
      def idUserData(keyName: String, fieldName: PrivateFields => Option[String]): Option[(String, String)] =
        for {
          idUser <- idUserOpt
          fields <- idUser.privateFields
          field  <- fieldName(fields)
        } yield keyName -> field

      val form = SubscriptionsForm().copy(
        data =  (
          idUserData("personal.first", _.firstName)                                  ++
          idUserData("personal.last", _.secondName)                                  ++
          idUserOpt.map("personal.emailValidation.email" -> _.primaryEmailAddress)   ++
          idUserOpt.map("personal.emailValidation.confirm" -> _.primaryEmailAddress) ++
          idUserData("personal.address.address1", _.address1)                        ++
          idUserData("personal.address.address2", _.address2)                        ++
          idUserData("personal.address.town", _.address3)                            ++
          idUserData("personal.address.postcode", _.postcode)
        ).toMap
      )
```

jennysivapalan: Sounds good - I'll make that change now.

jennysivapalan:
Going to merge this now as need it for next card.

PR #47, opened 2015-06-23
Ensure checkout creates salesforce account

tudorraul:
- SalesForce account created with the identity id and standard user details - First name, Last name, Email address, address

- An existing salesforce account will be amended with the users subscriptions ID

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -1,24 +1,26 @@
 package services
 
 import com.gu.identity.play.IdMinimalUser
+import com.gu.membership.salesforce.MemberId
 import com.typesafe.scalalogging.LazyLogging
 import model.SubscriptionData
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
-class CheckoutService(identityService: IdentityService) extends LazyLogging {
+class CheckoutService(identityService: IdentityService, salesforceService: SalesforceService) extends LazyLogging {
 
   def processSubscription(subscriptionData: SubscriptionData,
-                          idUser: Option[IdMinimalUser]): Future[Either[GuestUserNotCreated, IdMinimalUser]] =
-    idUser.map(user => Future(Right(user)))
+                          idUser: Option[IdMinimalUser]): Future[Option[MemberId]] =
+    idUser.map(user => Future(Some(UserId(user.id))))
       .getOrElse {
         logger.info(s"User does not have an Identity account, creating a Guest one.")
 
-        identityService.registerGuest(subscriptionData.personalData).map { guestFuture =>
-          guestFuture.right.map(id => IdMinimalUser(id, None))
-        }
+        identityService.registerGuest(subscriptionData.personalData)
       }
+      .flatMap(userId =>
+        userId.map(salesforceService.createOrUpdateUser(subscriptionData.personalData, _))
+          .getOrElse(Future(None)))
 }
 
-object CheckoutService extends CheckoutService(IdentityService)
+object CheckoutService extends CheckoutService(IdentityService, SalesforceService)
--- a/app/services/IdentityService.scala
+++ b/app/services/IdentityService.scala
@@ -6,13 +6,14 @@ import com.gu.monitoring.{AuthenticationMetrics, CloudWatch, RequestMetrics, Sta
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config
 import model.PersonalData
-import play.api.Logger
 import play.api.Play.current
 import play.api.libs.json._
 import play.api.libs.ws.{WS, WSRequest, WSResponse}
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
+
+case class UserId(id:String)
 
 case class GuestUserNotCreated(reason : String)
 
@@ -23,7 +24,7 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging 
       (response.json \ "user").asOpt[IdUser]
     }
 
-  def registerGuest(personalData: PersonalData): Future[Either[GuestUserNotCreated, String]] = {
+  def registerGuest(personalData: PersonalData): Future[Option[UserId]] = {
 
     val json = JsObject(Map(
       "primaryEmailAddress" -> JsString(personalData.email),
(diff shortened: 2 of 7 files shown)


app/services/SalesforceService.scala
```diff
@@ -0,0 +1,99 @@
...
+      Keys.MAILING_CITY -> personalData.address.town,
+      Keys.MAILING_POSTCODE -> personalData.address.postcode,
+      Keys.MAILING_COUNTRY -> "United Kingdom",
+      Keys.ALLOW_MEMBERSHIP_MAIL -> true
+    )) ++ Map(
+      Keys.ALLOW_THIRD_PARTY_EMAIL -> Some(true),
+      Keys.ALLOW_GU_RELATED_MAIL -> Some(true)
+    ).collect { case (k, Some(v)) => Json.obj(k -> v) }
+  }.reduce(_ ++ _)
+
```
afiore: Is there a specific reason why we do these intermediate transformations here (i.e. `.collect`, `.reduce`)? Why not just returning a Json object?

```scala
    Json.obj(
      Keys.EMAIL -> personalData.email,
      Keys.FIRST_NAME -> personalData.firstName,
      Keys.LAST_NAME -> personalData.lastName,
      Keys.MAILING_STREET -> personalData.address.street,
      Keys.MAILING_CITY -> personalData.address.town,
      Keys.MAILING_POSTCODE -> personalData.address.postcode,
      Keys.MAILING_COUNTRY -> "United Kingdom",
      Keys.ALLOW_MEMBERSHIP_MAIL -> true,
      Keys.ALLOW_THIRD_PARTY_EMAIL -> true,
      Keys.ALLOW_GU_RELATED_MAIL -> true)
```

app/services/IdentityService.scala
```diff
@@ -14,6 +13,8 @@ import play.api.libs.ws.{WS, WSRequest, WSResponse}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
+case class UserId(id:String)
```
afiore: Minor performance optimisation: why not use a [value class](http://docs.scala-lang.org/overviews/core/value-classes.html) here?

```scala
case class UserId(id:String) extends AnyVal
```

app/services/SalesforceService.scala
tudorraul: Sounds legit. This code was migrated across from https://github.com/guardian/membership-frontend/blob/6cdfced6cb2ddfe6c00284f6f501b9c677fb974a/frontend/app/services/MemberService.scala#L68 which uses dynamic values for the values in the Map. But for our implementations, at the moment, we can simplify further, your observation is correct.

app/services/IdentityService.scala
```diff
@@ -23,7 +24,7 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging
       (response.json \ "user").asOpt[IdUser]
     }
 
-  def registerGuest(personalData: PersonalData): Future[Either[GuestUserNotCreated, String]] = {
+  def registerGuest(personalData: PersonalData): Future[Option[UserId]] = {
```
jennysivapalan: Why is this no longer an `Either`?

app/services/SalesforceService.scala
```diff
@@ -0,0 +1,99 @@
...
+    override val apiURL = Config.Salesforce.apiURL.toString
+    override val stage = Config.Salesforce.envName
+
+    val authTask = ScheduledTask("", Authentication("", ""), 0.seconds, 30.minutes)(getAuthentication)
+
+    def authentication: Authentication = authTask.get()
+  }
+}
+
+trait ScheduledTask[T] {
```
jennysivapalan: I think this should be separated into a new file as we'll use it for Zuora too.

app/services/CheckoutService.scala
```diff
@@ -1,24 +1,26 @@
...
 
-        identityService.registerGuest(subscriptionData.personalData).map { guestFuture =>
-          guestFuture.right.map(id => IdMinimalUser(id, None))
-        }
+        identityService.registerGuest(subscriptionData.personalData)
       }
+      .flatMap(userId =>
+        userId.map(salesforceService.createOrUpdateUser(subscriptionData.personalData, _))
+          .getOrElse(Future(None)))
 }
```
jennysivapalan: Why have we removed the `Either` here too? We should be logging when things go wrong so we can alert on them.

app/services/SalesforceService.scala
```diff
@@ -0,0 +1,99 @@
...
+  val initialValue: T
+  val initialDelay: FiniteDuration
+  val interval: FiniteDuration
+
+  val name = getClass.getSimpleName
+
+  private implicit val system = Akka.system
+  lazy val agent = Agent[T](initialValue)
+
+  def refresh(): Future[T]
```
afiore: Does `refresh` need to be a `Future[T]`. `agent.sendOff` is [executed asynchronously](http://doc.akka.io/docs/akka/snapshot/scala/agents.html), so maybe I would get rid of the Future wrapper type and of the `Await.result` statement:

```scala
def refresh: T
...
agent.sendOff { _ => refresh }
```

app/services/IdentityService.scala
tudorraul: Looks like a space optimisation that the app will not benefit too much from, since these values are not piling up to an overwhelming amount, but rather created per request and GC'ed after the response is sent out.

And remember :)

afiore: All right then :)

app/services/IdentityService.scala
tudorraul: Because there is only one left case here and the handling of this scenario (logging) is already done within the identity client. Options are also easier to reason about.

app/services/SalesforceService.scala
tudorraul: Sounds like a refactoring to be made later. At the moment of writing the only client was the Salesforce integration code.

app/services/CheckoutService.scala
tudorraul: We do this here https://github.com/guardian/subscriptions-frontend/blob/salesforce-integration/app/services/IdentityService.scala#L125

app/services/SalesforceService.scala
tudorraul: I think it's to do with the fact that we use a library method that returns a future, `getAuthentication ` here https://github.com/guardian/subscriptions-frontend/blob/salesforce-integration/app/services/SalesforceService.scala#L51

afiore: Cool, that makes sense

afiore:
:+1: fine by me

PR #48, opened 2015-06-23
Check email addresses already in use, and handle signed in users

afiore:
- If the user enters an email address already present in Identity, encourage her/him to login or to continue sign up using another email.
- Pre-populate form with relevant data when the user is signed in.

@rtyley @davidrapson


app/configuration/Config.scala
```diff
@@ -35,9 +35,15 @@ object Config {
     val keys = if (idConfig.getBoolean("production.keys")) new ProductionKeys else new PreProductionKeys
 
     val webAppUrl = idConfig.getString("webapp.url")
-    
-    def webAppSigninUrl(uri: String): String =
-      (webAppUrl / "signin") ? ("returnUrl" -> s"$subscriptionsUrl$uri") ? ("skipConfirmation" -> "true")
+
+    def webAppSigninUrl(path: String): String = {
```
tudorraul: outer brackets redundant

app/views/checkout/payment.scala.html
```diff
@@ -1,16 +1,21 @@
-@(form: Form[model.SubscriptionData])
+@(form: Form[model.SubscriptionData], userIsSignedIn: Boolean)(implicit request: Request[_])
```
tudorraul: I would prefer to inject only the necessary data in this template, as the request has a wide scope that might make understanding the purpose of the unit harder.

afiore: Well spotted! Turns out that the `request` parameter [is unnecessary](https://www.playframework.com/documentation/2.4.x/api/scala/index.html#play.api.mvc.Call@path():String) here.

app/views/checkout/payment.scala.html
```diff
@@ -1,16 +1,22 @@
...
 
+
 <div class="gs-container">
     @fragments.page.header("Checkout")
 
-
-    <div class="page-section__content">
+    <div class="page-section__content" data-user-signedin="@userIsSignedIn">
+    @if(userIsSignedIn) {
+        <h3 class="h-section h-section--lead">Not you?</h3>
```
davidrapson: You don't have any of these classes in your styles, e.g., `h-section-*`, `.page-section__content`. Can you streamline this markup at all?. `page-section__content` should definitely go as you don't appear to have a parent `page-section` anywhere.

assets/javascripts/modules/form/email-check.js
```diff
@@ -0,0 +1,45 @@
...
+    function init() {
+
+        var YOUR_DETAILS_SUBMIT_ELEM = document.querySelector('.js-checkout-your-details-submit');
+        var EMAIL_ID_INPUT_ELEM      = document.querySelector('.js-checkout-email');
+        var USER_SIGNED_IN           = document.querySelector('data-user-signedin="true"]')
+
+        // do nothing if the user is already signed in
+        if (USER_SIGNED_IN) { return; }
+
+        ajax.init({page: {ajaxUrl: ''}});
```
davidrapson: Even though this is the only time you are using this at the moment I'd recommend initialising this upfront in `main.js`, similar to https://github.com/guardian/membership-frontend/blob/master/frontend/assets/javascripts/src/main.js#L56

assets/javascripts/modules/form/email-check.js
```diff
@@ -0,0 +1,45 @@
...
+
+        bean.on(YOUR_DETAILS_SUBMIT_ELEM, 'click', function (event) {
+            if (EMAIL_ID_INPUT_ELEM) {
+                var email = textUtils.removeWhitespace(EMAIL_ID_INPUT_ELEM.value);
+
+                if (email.length) {
```
davidrapson: This is going to be a string right? If so you don't need to check `.length`

assets/javascripts/modules/form/email-check.js
```diff
@@ -0,0 +1,45 @@
...
+                    ajax({
+                        url: '/checkout/check-identity?email=' + email
+                    }).then(function (response) {
+                        if (!response.emailInUse) {
+                            bean.fire(YOUR_DETAILS_SUBMIT_ELEM, "submit");
```
davidrapson: Are you submitting a form here? If so I'd rather keep track of the form element itself an trigger an submit on that, rather than simulating a click.

assets/javascripts/modules/form/email-check.js
```diff
@@ -0,0 +1,45 @@
...
+                        } else {
+                            alert("Your email is already in use! Please sign in or use another email address");
+                        }
+                    }).fail(function (_, msg) {
+                        console.error("Error reaching endpoint /checkout/check-identity:", msg);
```
davidrapson: I'm guessing this `console` and the above `alert` are just for debugging purposes at the moment? Will need to remove them afterwards.

assets/javascripts/modules/form/email-check.js
```diff
@@ -0,0 +1,45 @@
...
+                    });
+                } else {
+                    alert("Please enter an email address");
```
davidrapson: See above

app/views/checkout/payment.scala.html
```diff
@@ -1,16 +1,22 @@
...
 
 @main("Checkout | Subscriptions | The Guardian") {
 
+
 <div class="gs-container">
     @fragments.page.header("Checkout")
 
-
-    <div class="page-section__content">
+    <div class="page-section__content" data-user-signedin="@userIsSignedIn">
```
davidrapson: Not sure I follow the logic here. Not a massive fan of using the DOM to model data on single div on a single template. Can you expose this as a config object globally on uncached pages?

app/views/checkout/payment.scala.html
davidrapson: General request here to review the classes you've got on the markup to make sure you need them in.

davidrapson:
Had my pedantic :tophat: on here so quite a few comments. Probably need to review the signed-in check but :+1: aside from that.

afiore:
Thanks for the generous feedback @davidrapson, I will try addressing all the recommendations before merging this

assets/javascripts/modules/form/email-check.js
jennysivapalan: Yes - validation / errors back to the user is another card so felt like we'd duplicate work here.

assets/javascripts/modules/form/email-check.js
jennysivapalan: :eyes: :arrow_up:

davidrapson: :ok_hand: :eyes:

afiore:
@davidrapson @rtyley Hopefully I have addressed most of the comments you guys raised yesterday. 
Additionally, I have refactored the logic of the `email-check.js` module so that it returns a promise, and leaves the caller module (`checkout.js`) in charge of deciding how to best handle errors.

davidrapson:
:+1: Looks good to me, assuming exposing `guardian.user.isSignedIn` is OK if you have a combination of cached / uncached pages. You don't want that being exposed from the server on cached pages.

afiore:
@davidrapson It shouldn't be the case. Relevant controller actions use the  `GoogleAuthenticatedStaffAction` combinator, which should disable caching.

jennysivapalan:
:+1: Looks good :email:

afiore:
Verified on prod and works as expected.

PR #50, opened 2015-06-24
Configuration for Touchpoint services - Salesforce and Zuora.

jennysivapalan:
This PR creates a TouchpointBackend class that we can use to encapsulate the Zuora and Salesforce services. I haven't included it as part of this PR but from here we can add test users quite simply and switch between DEV and UAT (or PROD and UAT).

I've done this now as it will make it easier to implement the Zuora code. I tried to make use of classes in Membership common, specifically: https://github.com/guardian/membership-common/blob/master/src/main/scala/com/gu/membership/touchpoint/TouchpointBackendConfig.scala and https://github.com/guardian/membership-common/blob/master/src/main/scala/com/gu/membership/zuora/ZuoraApiConfig.scala however these are very coupled to Membership. So for now I've duplicated these and removed the elements we do not need.

 ZuoraService, it currently just pings an empty auth request. My next PR will address this.

Before merging I will need to update the PROD private properties (I'll keep PROD pointing to DEV until I have the correct credentials in place).


conf/DEV.conf
```diff
@@ -13,4 +13,6 @@ subscriptions.url="https://sub.thegulocal.com"
 
 play.ws.acceptAnyCertificate=true
 
+touchpoint.backend.default=DEV
+
 sentry.dns = ""
```
afiore: Just spotted a typo (pre-dating this PR): this should be "sentry.dsn" not sentry.dns!

jennysivapalan: ok - i can change that here

afiore:
I think it would be good to perform an actual authentication with Zuora as part of this PR, as this would provide us with a way to verify that all the different components are wired up correctly.

Also, while it is all right for now to duplicate some code from Membership, later on we should make some time to refactor these shared components into a single library.

jennysivapalan:
Sounds good to me - I was just working on the auth part so can add it to this PR.
Yes as part of the Auth I'm going to migrate some of the code from membership-frontend to membership-common. However there will be some duplication as I think some parts will be harder to migrate. We should definitely look into it as tech debt card.

jennysivapalan:
Build fails as it's pointing to a local snapshot. This will be fixed once https://github.com/guardian/membership-common/pull/69 has been merged a released. @afiore @tudorraul please could you be able to look over that PR too.

https://github.com/guardian/membership-common/pull/69 contains the common code for Zuora that membership-frontend and subscriptions-frontend that will be used here. 

This PR now makes use of the library to make an authentication request as suggested by @afiore

jennysivapalan:
I've updated the dependency and the properties in the private bucket (Prod is currently pointed to Zuora DEV instance). @tudorraul @afiore please could you let me know if you're happy for me to merge this.

afiore:
I am have little knowledge of Zuora, but it looks good at a first sight :+1:!

app/monitoring/Metrics.scala
```diff
@@ -0,0 +1,11 @@
+package monitoring
+
+import com.amazonaws.regions.{Region, Regions}
+import com.gu.monitoring.CloudWatch
+import configuration.Config
+
+trait Metrics extends CloudWatch {
+  val region = Region.getRegion(Regions.EU_WEST_1)
+  val stage = Config.stage
+  val application = "SubscriptionsFrontend"
```
tudorraul: we have an application constant that holds the application name, I think it's better to keep it unified https://github.com/guardian/subscriptions-frontend/blob/bbbba800417d8d2c2a0648f23898635733d9efce/app/configuration/Config.scala#L13

app/touchpoint/TouchpointBackendConfig.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  object BackendType {
+
+    object Default extends BackendType("default")
+
+    object Testing extends BackendType("test")
+
+  }
+
+  def byType(typ: BackendType = BackendType.Default, config: com.typesafe.config.Config) = {
```
tudorraul: perhaps `backendType` might be a better name here

app/monitoring/Metrics.scala
jennysivapalan: :+1: Done

app/touchpoint/TouchpointBackendConfig.scala
jennysivapalan: :+1: Done

PR #51, opened 2015-06-25
Rewire checkout service

afiore:
I had introduced two regressions in my previous PR: @tudorraul @jennysivapalan 

- `CheckoutService` had been accidentally unwired from the controller
- I was using the `disabled` attribute for form inputs, which has the undesirable effect of prevent the field value from being submitted. Replacing it with 'readonly', which seems to work fine.

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -2,6 +2,7 @@ package controllers
 
 import actions.CommonActions._
 import com.gu.identity.play.{IdUser, PrivateFields}
+import com.typesafe.scalalogging.LazyLogging
 import model.{AddressData, PaymentData, PersonalData, SubscriptionData}
 import play.api.libs.json._
 import play.api.mvc._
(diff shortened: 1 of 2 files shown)


app/controllers/Checkout.scala
```diff
@@ -45,6 +45,7 @@ object Checkout extends Controller {
           yield BadRequest(views.html.checkout.payment(formWithErrors, userIsSignedIn = idUserOpt.isDefined))
       },
       userData => {
+        CheckoutService.processSubscription(userData, AuthenticationService.authenticatedUserFor(request))
```
tudorraul: It looks like if `processSubscription` fails, the user is still taken to the "thank you" page. Please map this to the redirect, potentially with a fallback on rendering the checkout form in case of error.

afiore: Ok, I am fine in doing this as a temporary measure, but I am starting to think that `CheckoutService.processSubscription`, as well as most of the other calls we do to external applications should not return values of type `Future[Option[T]]`, but just of `Future[T]` (or maybe Either[Throwable, T]).

jennysivapalan:
:+1:

PR #52, opened 2015-06-26
Add client-side validations to checkout form

tudorraul:
screen shot 2015-06-26 at 14 59 13
screen shot 2015-06-26 at 15 00 30


davidrapson:
@tudorraul Not too sure about this. As you are already sharing the form markup from Membership it might be worth looking into how form validation is done there: https://github.com/guardian/membership-frontend/tree/master/frontend/assets/javascripts/src/modules/form

Happy to run through how that works if it helps?

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -0,0 +1,151 @@
...
+                }
+            });
+        }
+
+        if($EDIT_PAYMENT_DETAILS.length > 0){
+            bean.on($EDIT_PAYMENT_DETAILS[0], 'click', function (e) {
+                e.preventDefault();
+                collapseFieldsets();
+                $FIELDSET_PAYMENT_DETAILS.removeClass(FIELDSET_COLLAPSED);
+                $EDIT_PAYMENT_DETAILS.addClass(IS_HIDDEN);
```
davidrapson: Fair amount of boilerplate here, these three steps seem to be done a few times (albeit with different elements):

```js
e.preventDefault();
collapseFieldsets();
$FIELDSET_PAYMENT_DETAILS.removeClass(FIELDSET_COLLAPSED);
$EDIT_PAYMENT_DETAILS.addClass(IS_HIDDEN);
```

assets/javascripts/modules/checkout/email-check.js
```diff
@@ -18,7 +18,6 @@ define([
                 }).then(function (response) {
                     if (response.emailInUse) {
                         var msg = "Your email is already in use! Please sign in or use another email address";
-                        alert(msg);
                         reject(Error(msg));
```
davidrapson: Not 100% sure about this pattern, but at the lest you might want `new` here.

assets/javascripts/modules/checkout/form-elements.js
```diff
@@ -1,36 +1,52 @@
...
-    var formElements = {
-        $FIRST_NAME: $('.js-checkout-first'),
-        $LAST_NAME: $('.js-checkout-last'),
-        $EMAIL: $('.js-checkout-email'),
-        $ADDRESS1: $('.js-checkout-house'),
-        $ADDRESS2: $('.js-checkout-street'),
-        $ADDRESS3: $('.js-checkout-town'),
-        $POSTCODE: $('.js-checkout-postcode'),
-        $ACCOUNT: $('.js-checkout-account'),
+    return {
```
davidrapson: This object looks large? Is there opportunity to narrow this to a core set of selectors, i.e., can the error handling be generalised so you don't have to handle each field on a case-by-case basis.

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -0,0 +1,151 @@
...
+        $REVIEW_SORTCODE = form.$REVIEW_SORTCODE,
+        $REVIEW_HOLDER = form.$REVIEW_HOLDER,
+        $SMALLPRINT = form.$SMALLPRINT,
+        $FIELDSET_YOUR_DETAILS = form.$FIELDSET_YOUR_DETAILS,
+        $FIELDSET_PAYMENT_DETAILS = form.$FIELDSET_PAYMENT_DETAILS,
+        $FIELDSET_REVIEW = form.$FIELDSET_REVIEW,
+        $EDIT_YOUR_DETAILS = form.$EDIT_YOUR_DETAILS,
+        $EDIT_PAYMENT_DETAILS = form.$EDIT_PAYMENT_DETAILS,
+
+        FIELDSET_COLLAPSED = 'fieldset--collapsed',
```
davidrapson: My Preference is towards adding a state class here, e.g., `is-collapsed` as I find it a nice way of signifying that this has been added by some JS / is a behaviour of the component.

assets/javascripts/modules/checkout/checkout.js
tudorraul: Fair enough, this represents the behaviour of existing form workflow and it's not the main subject of this PR. Will consider your suggestion for future refactorings.

assets/javascripts/modules/checkout/form-elements.js
tudorraul: I thought about it. I'd prefer to keep the selector definition static for to make it straight forward to refactor the form markup.

davidrapson:
After conversation the approach seems OK for the moment, doesn't sound like the Membership approach  is quite the right approach at this point. :+1: with bit of refactoring to remove some of the repetition.

Might want to run the error message wording past someone as well

PR #55, opened 2015-06-26
Eslint

davidrapson:
This adds some client side linting to Subscriptions using [eslint](http://eslint.org/docs/rules/). I've added the smallest `.eslintrc` config file I could get away with; you'll probably want to tweak/define these rules. The rules are quite strict by default but as the app is new it's easy to start strict and relax as needed.

This PR also fixes a bunch of errors reported by the linter. Many are syntax / code-style changes but it also caught a number of errors in the code.

The important part is to get this running as part of the CI step to help catch these errors as part of a build step. I'll set that up with this branch next week before this is merged.

@tudorraul @afiore

### DIFF
--- a/assets/javascripts/modules/checkout/regex.js
+++ b/assets/javascripts/modules/checkout/regex.js
@@ -1,7 +1,9 @@
 define([], function () {
+    'use strict';
+
     return {
         isNumber: function isNumber(s) {
-            return ! /[^\d]+/.test(s);
+            return !/[^\d]+/.test(s);
         },
         isValidEmail: function isValidEmail(s) {
             return /^[^@\s]+@[^@\s]+\.[^@\s]+$/.test(s);
@@ -9,5 +11,5 @@ define([], function () {
         isPostcode: function isPostcode(s) {
             return /^[\sA-Z\d]{5,}$/i.test(s);
         }
-    }
+    };
 });
--- a/assets/javascripts/modules/checkout/validations.js
+++ b/assets/javascripts/modules/checkout/validations.js
@@ -1,4 +1,5 @@
 define(['$', 'modules/checkout/form-elements', 'modules/checkout/regex'], function ($, form, regex) {
+    'use strict';
 
     var ERROR_CLASS = 'form-field--error',
         EMAIL_ERROR_DEFAULT = 'Please enter a valid Email address';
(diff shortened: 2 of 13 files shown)


assets/javascripts/modules/checkout/regex.js
```diff
@@ -1,7 +1,9 @@
 define([], function () {
+    'use strict';
+
     return {
         isNumber: function isNumber(s) {
```
davidrapson: For these kinds of tests, it might be worth loading `https://github.com/lodash/lodash-amd` (which is very useful to have around anyway) as it includes https://lodash.com/docs#isNumber . The AMD build allows you to include single functions so you only get the minimum required in your build. cc/ @tudorraul

assets/javascripts/modules/checkout/validations.js
```diff
@@ -46,12 +47,12 @@ define(['$', 'modules/checkout/form-elements', 'modules/checkout/regex'], functi
 
 
     var validatePaymentDetails = function () {
-        var accountNumberValid = form.$ACCOUNT.val() != ''
+        var accountNumberValid = form.$ACCOUNT.val() !== ''
             && form.$ACCOUNT.val().length <= 10
-            && isNumber(form.$ACCOUNT.val());
```
davidrapson: This function didn't exist.

assets/javascripts/modules/checkout/validations.js
```diff
@@ -46,12 +47,12 @@ define(['$', 'modules/checkout/form-elements', 'modules/checkout/regex'], functi
...
+            && regex.isNumber(form.$ACCOUNT.val());
         toggleError(form.$ACCOUNT_CONTAINER, !accountNumberValid);
 
-        var holderNameValid = form.$HOLDER.val() != ''
+        var holderNameValid = form.$HOLDER.val() !== ''
```
davidrapson: @tudorraul You might want to give this a look to check if you do want strict equality. You nearly always do, but you might have reason to not do that.

tudorraul:
First read looks ok, I'll followup on the rest of the comments with some tests. Thanks for taking the time to do this.

davidrapson:
Just added linting to the build step so it will be picked up by teamcity. Also ran through a few pages and all looks good, validation still works etc.

screen shot 2015-06-29 at 10 34 47

tudorraul:
:+1:

PR #56, opened 2015-06-30
Display error message if email is taken

tudorraul:
This integrates the email check functionality with the checkout form validations.
screen shot 2015-06-30 at 10 56 03

### DIFF
--- a/assets/javascripts/modules/checkout/email-check.js
+++ b/assets/javascripts/modules/checkout/email-check.js
@@ -20,12 +20,14 @@ define([
                     url: '/checkout/check-identity?email=' + email
                 }).then(function (response) {
                     if (response.emailInUse) {
-                        reject(new Error('EMAIL_IN_USE'));
+                        reject('EMAIL_IN_USE');
                     } else {
                         resolve();
                     }
                 }).fail(function (_, msg) {
-                    console.error('Error reaching endpoint /checkout/check-identity:' + msg);
+                    if(console){
+                        console.error('Error reaching endpoint /checkout/check-identity:' + msg);
+                    }
                     reject(new Error('NETWORK_FAILURE'));
                 });
 
--- a/assets/javascripts/modules/checkout/email-check.js
+++ b/assets/javascripts/modules/checkout/email-check.js
@@ -20,7 +20,7 @@ define([
                     url: '/checkout/check-identity?email=' + email
                 }).then(function (response) {
                     if (response.emailInUse) {
-                        reject(Error('EMAIL_IN_USE'));
+                        reject(new Error('EMAIL_IN_USE'));
                     } else {
                         resolve();
                     }
@@ -30,7 +30,7 @@ define([
                 });
 
             } else {
-                reject(Error('Email is blank'));
+                reject(new Error('Email is blank'));
             }
         });
     }
--- a/assets/javascripts/modules/checkout/email-check.js
+++ b/assets/javascripts/modules/checkout/email-check.js
@@ -17,16 +17,16 @@ define([
                     url: '/checkout/check-identity?email=' + email
                 }).then(function (response) {
                     if (response.emailInUse) {
-                        var msg = "Your email is already in use! Please sign in or use another email address";
-                        reject(Error(msg));
+                        reject(Error('EMAIL_IN_USE'));
                     } else {
                         resolve();
                     }
                 }).fail(function (_, msg) {
-                    reject(Error("Error reaching endpoint /checkout/check-identity:" + msg))
+                    console.error('Error reaching endpoint /checkout/check-identity:' + msg);
+                    reject(new Error('NETWORK_FAILURE'));
                 });
 
-            } else { return reject(Error("Email is blank")) }
+            } else { return reject(Error('Email is blank')) }
         });
     }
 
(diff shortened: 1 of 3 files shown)


assets/javascripts/modules/checkout/email-check.js
```diff
@@ -17,17 +20,17 @@ define([
                     url: '/checkout/check-identity?email=' + email
                 }).then(function (response) {
                     if (response.emailInUse) {
-                        var msg = 'Your email is already in use! Please sign in or use another email address';
-                        reject(new Error(msg));
+                        reject(Error('EMAIL_IN_USE'));
```
davidrapson: Not actually sure why you are using an `Error` object unless you specifically want the stack trace. Probably still need `new` here though.

assets/javascripts/modules/checkout/email-check.js
```diff
@@ -17,17 +20,17 @@ define([
...
                     } else {
                         resolve();
                     }
                 }).fail(function (_, msg) {
-                    reject(new Error('Error reaching endpoint /checkout/check-identity:' + msg));
+                    console.error('Error reaching endpoint /checkout/check-identity:' + msg);
```
davidrapson: Remove this?

assets/javascripts/modules/checkout/email-check.js
```diff
@@ -17,17 +20,17 @@ define([
...
+                    reject(new Error('NETWORK_FAILURE'));
                 });
 
             } else {
-                reject(new Error('Email is blank'));
+                reject(Error('Email is blank'));
```
davidrapson: How come you removed `new`?

assets/javascripts/modules/checkout/validations.js
```diff
@@ -28,21 +38,49 @@ define(['$', 'modules/checkout/form-elements', 'modules/checkout/regex'], functi
...
+            var emptyFields = mandatoryFieldsPersonalDetails.filter(function (field) {
+                var isEmpty = field.input.val() === '';
+                toggleError(field.container, isEmpty);
+                return isEmpty;
+            });
+            var noEmptyFields = emptyFields.length === 0;
+
+            var validEmail = regex.isValidEmail(form.$EMAIL.val());
+            renderEmailError(!validEmail);
+            toggleError(form.$EMAIL_CONTAINER, !validEmail);
```
davidrapson: Another option might be, instead on performing side-effects inside the promise you pass an object to the `resolve` method with validation information about each field?

assets/javascripts/modules/checkout/validations.js
```diff
@@ -28,21 +38,49 @@ define(['$', 'modules/checkout/form-elements', 'modules/checkout/regex'], functi
...
+            else {
+                onFailure(new Error('Personal details incorrect.'));
+            }
         });
-        var noEmptyFields = emptyFields.length === 0;
 
-        var validEmail = regex.isValidEmail(form.$EMAIL.val());
-        renderEmailError(!validEmail);
-        toggleError(form.$EMAIL_CONTAINER, !validEmail);
+        var validationResult = staticValidations.then(function(){
```
davidrapson: If you want to use a Promise pattern, rather than nesting promises you might want to try [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)

tudorraul: the only thing with that is that the all the promises are started concomitantly and I wanted them to be chained.

assets/javascripts/modules/checkout/validations.js
```diff
@@ -28,21 +38,49 @@ define(['$', 'modules/checkout/form-elements', 'modules/checkout/regex'], functi
...
+
+            var warnIfEmailTaken = emailCheck.warnIfEmailTaken();
+
+            warnIfEmailTaken.catch(function (error) {
```
davidrapson: I'm not sure I'd class the email being taken as an error, as that's the point of the function. Makes sense the network failure (or just failure in general) being an error, but would I'd expect the email being taken to be part of the `resolve`. Either a boolean or an object with validation information.

davidrapson:
It's worth getting Sentry set up client-side to make sure that the `Error` pattern you are using doesn't end up sending errors to sentry for non-error cases. e.g. email already take.

assets/javascripts/modules/checkout/validations.js
davidrapson: Good point. If you were using a different pattern with errors you could chain them.

assets/javascripts/modules/checkout/validations.js
```diff
@@ -1,8 +1,18 @@
...
+             form,
+             emailCheck,
+             regex) {
     'use strict';
 
     var ERROR_CLASS = 'form-field--error',
-        EMAIL_ERROR_DEFAULT = 'Please enter a valid Email address';
+        EMAIL_ERROR_DEFAULT = 'Please enter a valid Email address',
+        EMAIL_ERROR_TAKEN = 'Your email is already in use! Please sign in or use another email address.',
+        EMAIL_ERROR_NETWORK = 'Your email could not be verified, please check your Internet connection and try again later.';
```
jennysivapalan: Have the validation messages been written? If not can we get someone to provide the copy? I don't think the `EMAIL_ERROR_NETWORK` is one we want to display to the user.

davidrapson: Seconded.

PR #57, opened 2015-06-30
Create subscription in Zuora

jennysivapalan:
First pass at creating a subscription in Zuora. 

There are a few caveats to this PR:
* ~~The build will fail until we merge and release https://github.com/guardian/membership-common/pull/71~~
* The product rate plan is hardcoded to Zuora DEV Monthly plan. The config requires a fair bit of change and given we're doing the different billing cycles as separate card and we don't have Prod set up I think we ca separate this out to a separate PR.

Other considerations:
* I've put Zuora in the TouchpointBackend class meaning that when we implement test users we should be able to configure this quite quickly. However as we don't have test users I'm currently directing the calls straight to `TouchpointBackend.Normal.zuoraService` methods which is not ideal as I'm we may miss updating this later on.
* I've changed Salesforce create / update call and Identity register guest user to return non Options. This helped to then create the flow of calls and also we should fail if one of these calls fail. 

To test this PR out we need to use the Go Cardless dummy details:
https://developer.gocardless.com/#dummy-bank-details

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -46,12 +46,8 @@ object Checkout extends Controller with LazyLogging {
           yield BadRequest(views.html.checkout.payment(formWithErrors, userIsSignedIn = idUserOpt.isDefined))
       },
       userData => {
-        CheckoutService.processSubscription(userData, AuthenticationService.authenticatedUserFor(request)).map {
-          case Some(memberId) =>
-            Redirect(routes.Checkout.thankyou())
-          case None =>
-            Redirect(routes.Checkout.renderCheckout())
-        }
+        for (_ <- CheckoutService.processSubscription(userData, AuthenticationService.authenticatedUserFor(request)))
+        yield  Redirect(routes.Checkout.thankyou())
       }
     )
   }
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -2,25 +2,33 @@ package services
 
 import com.gu.identity.play.IdMinimalUser
 import com.gu.membership.salesforce.MemberId
+import com.gu.membership.zuora.soap.Zuora.SubscribeResult
 import com.typesafe.scalalogging.LazyLogging
 import model.SubscriptionData
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
+case class Subscription(memberId: MemberId, zuoraId: SubscribeResult)
+
 class CheckoutService(identityService: IdentityService, salesforceService: SalesforceService) extends LazyLogging {
 
   def processSubscription(subscriptionData: SubscriptionData,
-                          idUser: Option[IdMinimalUser]): Future[Option[MemberId]] =
-    idUser.map(user => Future(Some(UserId(user.id))))
+                          idUser: Option[IdMinimalUser]): Future[Subscription] = {
+    def idUserF = idUser.map(user => Future(UserId(user.id)))
       .getOrElse {
-        logger.info(s"User does not have an Identity account, creating a Guest one.")
+      logger.info(s"User does not have an Identity account, creating a Guest one.")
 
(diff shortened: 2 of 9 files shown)


app/services/CheckoutService.scala
```diff
@@ -2,25 +2,33 @@ package services
...
 import com.gu.identity.play.IdMinimalUser
 import com.gu.membership.salesforce.MemberId
+import com.gu.membership.zuora.soap.Zuora.SubscribeResult
 import com.typesafe.scalalogging.LazyLogging
 import model.SubscriptionData
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
+case class Subscription(memberId: MemberId, zuoraSubscription: SubscribeResult)
```
jennysivapalan: @afiore this may conflict with the case class you are creating but we should be able to merge both.

app/services/CheckoutService.scala
```diff
@@ -2,25 +2,33 @@ package services
...
-        logger.info(s"User does not have an Identity account, creating a Guest one.")
+      logger.info(s"User does not have an Identity account, creating a Guest one.")
+
+      identityService.registerGuest(subscriptionData.personalData)
+    }
+
+    for {
+      idUser <- idUserF
+      memberId <- salesforceService.createOrUpdateUser(subscriptionData.personalData, idUser)
+      subscription <- TouchpointBackend.Normal.zuoraService.createSubscription(memberId, subscriptionData)   //TODO when implementing test-users this requires updating to supply data to correct location
```
jennysivapalan: We use TouchpointBackend.Normal.zuoraService as later we can update this to point to Normal (DEV / PROD) or Test (UAT). However this PR does not update Salesforce to distinguish between instances. @rtyley is it worth changing here or waiting until we implement test-users?

app/services/CheckoutService.scala
afiore: It should be fairly easy to reconcile. We should just agree on the name of the case class :)

app/services/zuora/Subscribe.scala
```diff
@@ -0,0 +1,83 @@
...
+            <ns2:ContractEffectiveDate>{effectiveDate}</ns2:ContractEffectiveDate>
+            <ns2:ContractAcceptanceDate>{contractAcceptanceDate}</ns2:ContractAcceptanceDate>
+            <ns2:InitialTerm>12</ns2:InitialTerm>
+            <ns2:RenewalTerm>12</ns2:RenewalTerm>
+            <ns2:TermStartDate>{effectiveDate}</ns2:TermStartDate>
+            <ns2:TermType>TERMED</ns2:TermType>
+          </ns1:Subscription>
+          <ns1:RatePlanData>
+            <ns1:RatePlan xsi:type="ns2:RatePlan">
+              <ns2:ProductRatePlanId>2c92c0f84bbfec8b014bc655f4852d9d</ns2:ProductRatePlanId>
```
jennysivapalan: This is a hard coded value of Zuora DEV - Digital Monthly pack which will need changing.

app/controllers/Checkout.scala
```diff
@@ -46,12 +46,8 @@ object Checkout extends Controller with LazyLogging {
           yield BadRequest(views.html.checkout.payment(formWithErrors, userIsSignedIn = idUserOpt.isDefined))
       },
       userData => {
-        CheckoutService.processSubscription(userData, AuthenticationService.authenticatedUserFor(request)).map {
-          case Some(memberId) =>
-            Redirect(routes.Checkout.thankyou())
-          case None =>
-            Redirect(routes.Checkout.renderCheckout())
-        }
+        for (_ <- CheckoutService.processSubscription(userData, AuthenticationService.authenticatedUserFor(request)))
```
tudorraul: shorter with `map`

TesterSpike:
:+1:

PR #58, opened 2015-06-30
Convert guest users

afiore:
When a guest user subscribes, display a form in the thank you page inviting her to complete her account by setting password. This PR implements only a basic password validation, I plan on adding the same password strength widget used in Identity-frontend and Membership in a separate one.

**Note:** In order to be able to set a password for an Identity guest user account the subscription-frontend token needs to have the `user.password.c` permission enabled (I have [updated](https://github.com/guardian/identity/pull/283) the corresponding PR on identity).

screen shot 2015-06-30 at 14 05 56

On successful submission of the password form the user gets redirected to another page
linking to the Identity frontend website.

screen shot 2015-06-30 at 14 02 05

@rtyley @jennysivapalan @tudorraul

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -40,7 +40,6 @@ object Config {
 
     val webAppUrl = idConfig.getString("webapp.url")
 
-
     val webAppProfileUrl = webAppUrl / "account" / "edit"
 
     def webAppSigninUrl(path: String): String =
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -15,6 +15,8 @@ object Config {
 
   val config = ConfigFactory.load()
 
+  val playSecret = config.getString("play.crypto.secret")
+
   val googleAuthConfig = {
     val con = ConfigFactory.load().getConfig("google.oauth")
     GoogleAuthConfig(
@@ -37,6 +39,9 @@ object Config {
     val keys = if (idConfig.getBoolean("production.keys")) new ProductionKeys else new PreProductionKeys
 
     val webAppUrl = idConfig.getString("webapp.url")
+
+
+    val webAppProfileUrl = webAppUrl / "account" / "edit"
 
     def webAppSigninUrl(path: String): String =
       (webAppUrl / "signin") ? ("returnUrl" -> absoluteUrl(path)) ? ("skipConfirmation" -> "true")
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -3,10 +3,12 @@ package controllers
 import actions.CommonActions._
 import com.gu.identity.play.{IdUser, PrivateFields}
 import com.typesafe.scalalogging.LazyLogging
-import model.{AddressData, PaymentData, PersonalData, SubscriptionData}
+import forms.{FinishAccountForm, SubscriptionsForm}
 import play.api.libs.json._
 import play.api.mvc._
-import services.{AuthenticationService, CheckoutService, IdentityService}
+import services.CheckoutService
+import services.CheckoutService.CheckoutResult
+import services._
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
@@ -46,13 +48,29 @@ object Checkout extends Controller with LazyLogging {
           yield BadRequest(views.html.checkout.payment(formWithErrors, userIsSignedIn = idUserOpt.isDefined))
       },
       userData => {
-        for (_ <- CheckoutService.processSubscription(userData, AuthenticationService.authenticatedUserFor(request)))
-        yield  Redirect(routes.Checkout.thankyou())
+        val idUserOpt = AuthenticationService.authenticatedUserFor(request)
+        CheckoutService.processSubscription(userData, idUserOpt).map {
+          case CheckoutResult(_, guestUser: GuestUser, _) =>
+            Some(FinishAccountForm().bind(guestUser.toFormParams))
+          case CheckoutResult(_, _:MinimalIdUser, _) =>
+            None
+        }.map(formOpt => Ok(views.html.checkout.thankyou(formOpt)))
       }
     )
   }
 
   def thankyou = GoogleAuthenticatedStaffAction(Ok(views.html.checkout.thankyou()))
+
+  def processFinishAccount = GoogleAuthenticatedStaffAction.async { implicit request =>
+    FinishAccountForm().bindFromRequest.fold( formWithErrors => {
+      Future {
+        BadRequest(views.html.checkout.thankyou(Some(formWithErrors)))
+      }
+    }, guestAccountData => {
+      IdentityService.convertGuest(guestAccountData.guestUser, guestAccountData.password)
+        .map(_ => Ok(views.html.checkout.alldone()))
+    })
+  }
 
   def checkIdentity(email: String) = GoogleAuthenticatedStaffAction.async { implicit request =>
     for {
@@ -66,45 +84,3 @@ object Checkout extends Controller with LazyLogging {
     }
 }
 
-object SubscriptionsForm {
-  import play.api.data.Forms._
-  import play.api.data._
-
-  private val addressDataMapping = mapping(
-    "address1" -> text,
-    "address2" -> text,
-    "town" -> text,
-    "postcode" -> text
-  )(AddressData.apply)(AddressData.unapply)
-
-  private val emailMapping = tuple(
-    "email" -> email,
-    "confirm" -> email)
-    .verifying("Emails don't match", email => email._1 == email._2)
-    .transform[String](
-      email => email._1, // Transform to a single field
-      email => (email, email) // Reverse transform from a single field to multiple
-    )
-
-  private val personalDataMapping = mapping(
-    "first" -> text,
-    "last" -> text,
-    "emailValidation" -> emailMapping,
-    "address" -> addressDataMapping
-  )(PersonalData.apply)(PersonalData.unapply)
-
-  private val paymentDataMapping = mapping(
-    "account" -> text(8, 8),
-    "sortcode1" -> text(2, 2),
-    "sortcode2" -> text(2, 2),
-    "sortcode3" -> text(2, 2),
-    "holder" -> text
-  )(PaymentData.apply)(PaymentData.unapply)
-
-  private val subsForm = Form(mapping(
-    "personal" -> personalDataMapping,
-    "payment" -> paymentDataMapping
-  )(SubscriptionData.apply)(SubscriptionData.unapply))
-
-  def apply() = subsForm
-}
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -67,7 +67,7 @@ object Checkout extends Controller with LazyLogging {
         BadRequest(views.html.checkout.thankyou(Some(formWithErrors)))
       }
     }, guestAccountData => {
-      IdentityService.convertGuest(guestAccountData.guestUser, guestAccountData.password)
+      IdentityService.convertGuest(guestAccountData.password, IdentityToken(guestAccountData.token))
         .map(_ => Ok(views.html.checkout.alldone()))
     })
   }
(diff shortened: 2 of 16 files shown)


app/controllers/Checkout.scala
```diff
@@ -46,18 +47,30 @@ object Checkout extends Controller with LazyLogging {
...
+          case CheckoutResult(_, guestUser: GuestUser) =>
+            Some(FinishAccountForm().bind(guestUser.toFormParams))
+          case CheckoutResult(_, _:MinimalIdUser) =>
+            None
+        }.map(formOpt => Ok(views.html.checkout.thankyou(formOpt)))
       }
     )
   }
 
   def thankyou = GoogleAuthenticatedStaffAction(Ok(views.html.checkout.thankyou()))
```
afiore: Perhaps the Thankyou page doesn't need to exist anymore as a separate route as we render as part the `handleCheckout` action. Should I delete it?

app/services/IdentityService.scala
```diff
@@ -29,28 +27,38 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging
...
+
+    identityApiClient.createGuest(json).map(response => response.json.as[GuestUser])
+  }
 
+  def convertGuest(gu: GuestUser, password: String): Future[Unit] = {
+    val json = JsObject(Map("password" -> JsString(password)))
 
-    identityApiClient.createGuest(json).map(response => (response.json \ "guestRegistrationRequest" \ "userId").asOpt[String].map(UserId))
+    identityApiClient.convertGuest(json, gu).map { response =>
+      if (!response.status.toString.matches("""2\d\d""")) {
```
jennysivapalan: What are the response codes we can get from Identity for this request? Can we match on those rather than use a regex?

app/services/IdentityService.scala
```diff
@@ -130,6 +140,17 @@ object IdentityApiClient extends IdentityApiClient with LazyLogging {
       .withWSFailureLogging(endpoint)
       .withCloudwatchMonitoringOfPost
   }
+
+  override def convertGuest: (JsValue, GuestUser) => Future[WSResponse] = (json, gu) => {
+    val endpoint = authoriseCall(WS.url(s"$identityEndpoint/guest/${gu.id}"))
+
+    val ep = endpoint
```
jennysivapalan: In stead of `ep` perhaps `request`?

app/services/IdentityService.scala
afiore: We could match on 201, but I think it would be a bit more robust if we match on something like `2xx` (i.e. any success status code)

app/services/IdentityService.scala
```diff
@@ -29,28 +27,38 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging
...
-      "statusFields" -> JsObject(Map("receiveGnmMarketing" -> JsString("true")).toSeq)
-    ).toSeq)
+      )),
+      "statusFields" -> JsObject(Map("receiveGnmMarketing" -> JsString("true")))
+    ))
+
+    identityApiClient.createGuest(json).map(response => response.json.as[GuestUser])
+  }
 
+  def convertGuest(gu: GuestUser, password: String): Future[Unit] = {
```
jennysivapalan: Maybe we don't need to define the return type.

app/services/package.scala
```diff
@@ -0,0 +1,59 @@
...
+        case _            => JsError("Cannot parse IdentityToken")
+      }
+    }
+  }
+
+  sealed trait UserIdData {
+    def id: UserId
+  }
+
+  case class MinimalIdUser(user: IdMinimalUser) extends UserIdData {
```
jennysivapalan: I'm not sure of  the purpose of this case class? Can we just use `IdMiminalUser`? It's case class is defined as
`case class IdMinimalUser(id: String, displayName: Option[String])`

app/views/checkout/thankyou.scala.html
```diff
@@ -1,4 +1,10 @@
-@()
+@import model.GuestAccountData
+@import services.UserId.{paramName => userIdParam}
+@import services.IdentityToken.{paramName => identityTokenParam}
+
+@import routes.Checkout.processFinishAccount
+
+@(formOpt: Option[Form[GuestAccountData]] = None)
```
jennysivapalan: Can we name `formOpt` to be something more related to `GuestAccountData` e.g. `guestAccountForm` ?

app/views/checkout/thankyou.scala.html
```diff
@@ -41,9 +47,29 @@ <h3 class="section-header">Download your apps</h3>
...
+            <p>Enter a password to finish creating your account:</p>
+            <form action="@processFinishAccount" method="POST" class="form">
+                <fieldset class="fieldset">
+                    <div class="fieldset__fields">
+                        <div class="form-field js-checkout-finish-account-password">
+                            <label class="label" for="last">Password</label>
+                            <input name="@userIdParam" type="hidden" value="@form.data.get(userIdParam)">
+                            <input name="@identityTokenParam" type="hidden" value="@form.data.get(identityTokenParam)">
+                            <input name="password" type="password" class="input-text js-input">
+                            @checkout.errorMessage("Password is too short")
```
jennysivapalan: We should list what the requirements for password are somewhere (here or an FAQ page?). cc @rtyley 
@afiore Do we get error messages from Identity as perhaps we can pass this through here?

assets/javascripts/modules/checkout/validations.js
```diff
@@ -13,6 +13,10 @@ define(['$', 'modules/checkout/form-elements', 'modules/checkout/regex'], functi
         {input: form.$POSTCODE, container: form.$POSTCODE_CONTAINER}
     ];
 
+    function isNumber (n) {
+        return !isNaN(parseFloat(n)) && isFinite(n);
+    }
```
jennysivapalan: @davidrapson fixed this (I think).

app/controllers/Checkout.scala
jennysivapalan: We should still have it as it exists as a route and a user can book mark it. We need to update the page to include order subscription number ( @rtyley can you confirm?) and so we'll have to make a call to Zuora and / or Salesforce to get the data. 
I think `handleCheckout` should redirect to this instead?

app/services/package.scala
afiore: The idea here is to define an [algebraic data type](http://twitter.github.io/effectivescala/#Functional programming-Case classes as algebraic data types) to represent both authenticated and guest users (both having an `id` method).

app/views/checkout/thankyou.scala.html
afiore: good point!

app/views/checkout/thankyou.scala.html
afiore: We don't as the identity token expires after one request and we do the client side validation before submitting the form. What we could do is to specify in the error message that the password should have n characters at least.

afiore: Also, we plan on adding the password widget used in Identity frontend, which would communicate to the user whether a password is strong enough

app/services/package.scala
jennysivapalan: Okay I see - I didn't spot where it was being used in the Controller. I think this is okay but we need to use IdMinimalUser when implementing `test-users` and I'm a bit concerned we're going to get ourselves confused with the type of User. @rtyley what do you think?

app/controllers/Checkout.scala
afiore: This is currently a mock page (if you look into it, order reference number and subscription code are hardcoded). Happy to handle that as a redirect and pass the form params as query strings if we really have a requirement for bookmarking.

jennysivapalan: Yeah we need to update the mock up to have correct values - I hope there is card for this! 
I don't know if bookmarking is a requirement but we caveat for this in Membership so I think we should follow suit.

app/configuration/Config.scala
```diff
@@ -37,6 +39,9 @@ object Config {
 
     val webAppUrl = idConfig.getString("webapp.url")
 
+
```
tudorraul: one line too many

app/services/IdentityService.scala
tudorraul: We use maths when logging response status code type (i.e. 2xx, 5xx) in CloudWatch metrics. We can use something similar here I reckon. https://github.com/guardian/membership-common/blob/a891156ed37ee1873cc5c13c79b45588293a6ce0/src/main/scala/com/gu/monitoring/Metrics.scala#L6

app/services/IdentityService.scala
tudorraul: I think the name can be a bit more specific here i.e. `convertGuestToUser`

app/services/package.scala
```diff
@@ -0,0 +1,59 @@
...
+  case class GuestUser(id: UserId, token: IdentityToken) extends UserIdData {
+    def toFormParams = Map(
+      UserId.paramName -> id.toString,
+      IdentityToken.paramName -> token.toString
+    )
+  }
+
+  object GuestUser {
+    implicit val guestUserTokenReads: Reads[GuestUser] = (
+      (JsPath \  "guestRegistrationRequest" \ "userId").read[UserId] and
```
tudorraul: indentation (+1)

app/controllers/Checkout.scala
```diff
@@ -46,14 +48,30 @@ object Checkout extends Controller with LazyLogging {
...
       userData => {
-        for (_ <- CheckoutService.processSubscription(userData, AuthenticationService.authenticatedUserFor(request)))
-        yield  Redirect(routes.Checkout.thankyou())
+        val idUserOpt = AuthenticationService.authenticatedUserFor(request)
+        CheckoutService.processSubscription(userData, idUserOpt).map {
+          case CheckoutResult(_, guestUser: GuestUser, _) =>
+            Some(FinishAccountForm().bind(guestUser.toFormParams))
+          case CheckoutResult(_, _:MinimalIdUser, _) =>
+            None
+        }.map(formOpt => Ok(views.html.checkout.thankyou(formOpt)))
```
jennysivapalan: So do we want to change this to a redirect?

conf/routes
```diff
@@ -22,6 +22,7 @@ GET         /checkout                        controllers.Checkout.renderCheckout
 POST        /checkout                        controllers.Checkout.handleCheckout
 GET         /checkout/thankyou               controllers.Checkout.thankyou
 GET         /checkout/check-identity         controllers.Checkout.checkIdentity(email: String)
+POST        /checkout/finish-account         controllers.Checkout.processFinishAccount
```
jennysivapalan: Maybe this endpoint could be called `/checkout/register-guest-user`

jennysivapalan:
A couple small naming comments but otherwise :+1:

app/services/IdentityService.scala
```diff
@@ -29,31 +27,36 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging
...
 
+    identityApiClient.createGuest(json).map(response => response.json.as[GuestUser])
+  }
 
-    identityApiClient.createGuest(json).map { response =>
-      UserId((response.json \ "guestRegistrationRequest" \ "userId").as[String])
+  def convertGuest(gu: GuestUser, password: String): Future[Unit] = {
+    val json = JsObject(Map("password" -> JsString(password)))
+    identityApiClient.convertGuest(json, gu).map { response =>
+      if (!response.status.toString.matches("""2\d\d""")) {
```
rtyley: I'm with @jennysivapalan on this one - I think just match against the Play constant `CREATED` (== `201`). If the return status code of this endpoint changes, I'd like to know why.

davidmcdowell:
Looks great, but if we want users to go to the trouble of creating a password we should sell in the benefits a little stronger.  I know this is a challenge at this stage as the benefits won't be available until future releases. maybe something like:

Complete your registration
Create a password to access your account online.

tudorraul:
Code looks good! :+1:

afiore:
Verified on prod, but form password form fails due to missing CSFR token (see PR #66).

PR #61, opened 2015-06-30
Fix json serialisation in IdentityService.registerUser

afiore:
@markjamesbutler @rtyley @tudorraul


tudorraul:
LGTM :+1: bar the call passes the ID API validations -- if any

afiore:
Verified on my local instance of Identity and the `receiveGnmMarketing ` field in the corresponding db record is now stored as a boolean.

PR #62, opened 2015-07-01
Load in rate plan from Configuration. 

jennysivapalan:
This relies on https://github.com/guardian/membership-common/pull/72

Currently this loads in just one ProductRatePlan but we should be able to extend this to a list when we bring in the other billing frequencies and products.

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -22,10 +22,14 @@ class CheckoutService(identityService: IdentityService, salesforceService: Sales
       identityService.registerGuest(subscriptionData.personalData)
     }
 
+    //TODO when implementing test-users this requires updating to supply data to correct location
+    val touchpointBackend = TouchpointBackend.Normal
+    val ratePlan = touchpointBackend.ratePlan.ratePlanId
+
     for {
       idUser <- idUserF
       memberId <- salesforceService.createOrUpdateUser(subscriptionData.personalData, idUser)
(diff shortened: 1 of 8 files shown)


app/services/CheckoutService.scala
```diff
@@ -22,10 +22,14 @@ class CheckoutService(identityService: IdentityService, salesforceService: Sales
       identityService.registerGuest(subscriptionData.personalData)
     }
 
+    //TODO when implementing test-users this requires updating to supply data to correct location
+    val touchpointBackend = TouchpointBackend.Normal
+    val ratePlan = touchpointBackend.ratePlan.ratePlanId
+
     for {
       idUser <- idUserF
       memberId <- salesforceService.createOrUpdateUser(subscriptionData.personalData, idUser)
```
rtyley: I guess `salesforceService` should sit under TouchpointBackend too eventually?

jennysivapalan: Yes - I didn't want to do it in this PR as it looks like a bit of refactoring is required.

jennysivapalan: I'm working on making this change as a separate PR

jennysivapalan: Turns out we don't need to do this as the service calls TouchpointBackend: https://github.com/guardian/subscriptions-frontend/blob/master/app/services/SalesforceService.scala#L34-36

PR #63, opened 2015-07-02
Setup client-side Sentry

davidrapson:
- Setup raven-js for Sentry
- Add build version to guardian global object
- Move javascript -> javascriptFirstSteps, as the `isModernBrowser` / global vars should be in the `<head>`

As well as generally tracking client-side errors, this also helps validate if the pattern [`onFailure(new Error('Personal details incorrect.')); `](https://github.com/guardian/subscriptions-frontend/blob/95909983e925a73aa3e50b41c0051a0478b6a01c/assets/javascripts/modules/checkout/validations.js#L61) results in Sentry noise and console errors.

Fairly confident it will as the current pattern does this in the console:

screen shot 2015-07-02 at 12 48 33

// @rtyley @tudorraul

### DIFF
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -7,7 +7,6 @@
 <head>
     @fragments.head(title)
     @fragments.javascriptFirstSteps(jsVars)
-    <script src="//cdn.optimizely.com/js/293455139.js"></script>
 </head>
     <body class="@bodyClassName">
         @fragments.header()
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -6,6 +6,8 @@
 <html lang="en-GB">
 <head>
     @fragments.head(title)
+    @fragments.javascriptFirstSteps(jsVars)
+    <script src="//cdn.optimizely.com/js/293455139.js"></script>
 </head>
     <body class="@bodyClassName">
         @fragments.header()
@@ -13,8 +15,6 @@
         @content
 
         @fragments.footer.footer(footer)
-
-        @fragments.javascript(jsVars)
 
         @fragments.analytics.google()
         @fragments.analytics.omniture()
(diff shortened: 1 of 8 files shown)


app/views/main.scala.html
```diff
@@ -6,6 +6,8 @@
 <html lang="en-GB">
 <head>
     @fragments.head(title)
+    @fragments.javascriptFirstSteps(jsVars)
+    <script src="//cdn.optimizely.com/js/293455139.js"></script>
```
tudorraul: I think we can remove this for now.

assets/javascripts/bower.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "subscriptions javascripts",
-  "version": "0.1",
```
tudorraul: is this removed because we don't publish this module?

assets/javascripts/main.js
```diff
@@ -1,17 +1,28 @@
...
+    digitalpack
 ) {
     'use strict';
 
-    // Global
+    /**
+     * Set up Raven, which speaks to Sentry to track errors
+     */
+    /*global Raven */
+    Raven.config('https://6dd79da86ec54339b403277d8baac7c8@app.getsentry.com/47380', {
```
tudorraul: should the url not be made secret? I think this can be passed in from the app `Config` object.

tudorraul:
:+1: bar the comments

app/views/main.scala.html
davidrapson: Agreed!

assets/javascripts/bower.json
davidrapson: Correct

assets/javascripts/main.js
davidrapson: Nope. This is the public version url, so doesn't need to be secret.

afiore: With the current setup, do we capture errors also in our development environments?

davidrapson: Correct, the whitelist limits to PROD, and only under `/assets/`

afiore:
:+1:

PR #64, opened 2015-07-02
Set delay in payment request by Zuora upon creating a subscription

tudorraul:
A delay is achieved by setting a future date in the `ContractAcceptanceDate` field when a subscription is created in Zuora

This is how a next day (dev configuration) payment request is set in Zuora
screen shot 2015-07-02 at 15 55 09

:sparkles:


jennysivapalan:
I'm not sure it's worth a config value, I don't think we expect to see anything different for dev over prod. Otherwise :+1: :date:

TesterSpike:
Having a 1 day free period in dev will be useful for testing as we can check if subscription status is correctly changed after the free period is completed if the payment has failed.

PR #65, opened 2015-07-02
Read-only email address

afiore:
Marks the email address as read-only and hide email confirmation when pre-populating user fields.
screen shot 2015-07-02 at 16 38 16

@rtyley @davidrapson

### DIFF
--- a/assets/stylesheets/_form.scss
+++ b/assets/stylesheets/_form.scss
@@ -1,0 +1,4 @@
+input[readonly] {
+  background: $c-neutral7;
+  color: $c-neutral2;
+}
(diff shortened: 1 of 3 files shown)


assets/stylesheets/_form.scss
```diff
@@ -0,0 +1,4 @@
+input[readonly] {
+  background: $c-neutral7;
```
davidrapson: `background-color` is preferable but it doesn't really matter.

davidrapson:
:+1: LGTM

afiore:
Verified on prod

PR #66, opened 2015-07-03
Add CSFR token input to finish-account form

afiore:
Adds CSFR token as a hidden input in the Thank you page finish account form. This is now required in all forms, otherwise Play will just reject the request with an "invalid token" error.


jennysivapalan:
:+1:

afiore:
Verified on prod and form is now working as expected

PR #68, opened 2015-07-03
Password strength indicator

afiore:
Implement a password strength indicator reusing code from membership @rtyley @davidrapson

screen shot 2015-07-03 at 12 06 07


app/views/fragments/javascriptFirstSteps.scala.html
```diff
@@ -19,8 +20,18 @@
...
+            newElm.src = '@hashedPathFor("javascripts/main.min.js")';
             document.getElementsByTagName("head")[0].appendChild(newElm);
         }
     })(guardian.isModernBrowser);
+
+    // Initialise requireJS loader module Curl
+    var curl = {
+        apiName: "require",
+        paths: {
+            zxcvbn: '@hashedPathFor("javascripts/vendor/zxcvbn-2014-08-04.js")'
```
davidrapson: Not installing via a package manger?

app/views/fragments/javascriptFirstSteps.scala.html
```diff
@@ -19,8 +20,18 @@
...
+        }
+    };
 </script>
+
+<script src='@hashedPathFor("javascripts/vendor/curl-2014-04-02.js")'></script>
```
davidrapson: Why `2014-04-02`?

assets/stylesheets/main.scss
```diff
@@ -23,6 +23,7 @@
 @import 'form';
 @import 'type';
 @import 'helpers';
+@import 'form';
```
davidrapson: Double import? Also on line 23

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -144,8 +146,8 @@ define(['$',
                     bean.fire($FINISH_ACCOUNT_SUBMIT[0], 'click');
                 }
             });
-
         }
+        password.init();
```
davidrapson: Why is this being initialised inside `finishAccount`?

davidrapson:
Couple of comments that would be good to chat through first, but then :+1:

app/views/fragments/javascriptFirstSteps.scala.html
afiore: That's the date of the [latest change](https://github.com/cujojs/curl/blob/master/dist/curl-with-js-and-domReady/curl.js) on that file. I thought that `assets/vendor/omniture-2014-11-19.js` was named following this convention.

assets/stylesheets/main.scss
afiore: Well spotted! Git automerge side-effect...

afiore:
@davidrapson I believe I have addressed all of you comments. Let me know if you spot any other potential issue.

app/views/fragments/javascriptFirstSteps.scala.html
davidrapson: Omniture is a bit of a special case I think as that's a custom build. Just `curl.js` is fine here.

assets/javascripts/modules/components/password.js
```diff
@@ -0,0 +1,68 @@
...
+
+
+    /**
+     * load in zxcvbn lib as it is ~700kb!
+     * setup listener for length check
+     * setup listener for zxcvbn.score check
+     */
+    var addListeners = function () {
+        require(['js!zxcvbn'], function() {
+            STRENGTH_INDICATOR_ELEM.classList.toggle('is-hidden');
```
davidrapson: One other thing, this assumes that you have a `classList` polyfill available which is the case for Membership, might not be the case here.

davidrapson:
Sorry, there's one more thing to cover that I missed, which is the need for a `classList` polyfill or to rewrite it to use some other method if you'd rather not do that.

afiore:
Added the `ClassList` Polyfill. I have also renamed `curl-2014-xx-xx.js` into `curl.js`

davidrapson:
:+1: Nice. Thanks for sticking with me.

afiore:
Verified on prod and the password strength indicator is being displayed as expected

PR #70, opened 2015-07-03
Use CODE ID API locally

tudorraul:


conf/DEV.conf
```diff
@@ -10,7 +10,7 @@ identity {
   apiToken = "subscriptions-dev-client-token"
 }
 
-subscriptions.url="https://sub.thegulocal.com"
+subscriptions.url="https://idapi.code.dev-theguardian.com"
```
jennysivapalan: I don't think this is not the url you want to change. You probably want identity.baseUri

afiore:
:+1: This is a more sensible default than running off a local and potentially out of sync instance of Identity. However, we should always double check that API methods work in the same way when we hit the production instance of Identity.

PR #71, opened 2015-07-03
Partial refactor of validation JS

davidrapson:
- Removes Promise `reject` pattern for false values.
- Removes using `new Error` to handle false values, causes console errors.
- Only `reject` on actual failures (e.g., network failure).

Prevents this:
cecafbdc-20b8-11e5-8a43-e86a60a43bc7-1

This PR only refactors the promise pattern in places where it causes errors.
I feel there is still opportunity to split things out more to avoid side effects in some functions and to make the validation logic more generic but that's outside the scope of this PR.

A new PR separating DOM code from validation would help us add some tests around the validation more easily.

validity mov


afiore:
:+1 on both refactoring the promise and on the idea of separating DOM manipulation from of the actual validation logic.

assets/javascripts/modules/checkout/emailCheck.js
```diff
@@ -0,0 +1,12 @@
+define(['utils/ajax'], function (ajax) {
+    'use strict';
+
+    return function(email) {
+        return ajax({ url: '/checkout/check-identity?email=' + email }).then(function (response) {
+            return response.emailInUse;
+        }).fail(function (err) {
```
tudorraul: How will the network failures going to be handled?

assets/javascripts/modules/checkout/validations.js
```diff
@@ -1,18 +1,24 @@
...
 
-    var ERROR_CLASS = 'form-field--error',
-        EMAIL_ERROR_DEFAULT = 'Please enter a valid Email address',
-        EMAIL_ERROR_TAKEN = 'Your email is already in use! Please sign in or use another email address.',
-        EMAIL_ERROR_NETWORK = 'There\'s been a problem. Please try again later.';
+    var ERROR_CLASS = 'form-field--error';
+    var MESSAGES = {
+        emailInvalid: 'Please enter a valid email address.',
+        emailTaken: 'Your email is already in use. Please sign in or use another email address.',
+        emailFailure: 'There has been a problem. Please try again later.'
```
tudorraul: This message, initially displayed mostly in network failure cases, is not used

davidrapson: Whoops, good catch meant to add it here https://github.com/guardian/subscriptions-frontend/blob/validation-refactor/assets/javascripts/modules/checkout/validations.js#L87 will sort that.

assets/javascripts/modules/checkout/emailCheck.js
davidrapson: See below, should be handled in `validations.js`, just forgot to re-add that part…

davidrapson:
@TesterSpike Would you be able to go over this one? Just confirm that client-side validation is all still OK on checkout pages.

TesterSpike:
:+1:

PR #72, opened 2015-07-03
Toggle js

davidrapson:
- Move out toggle code, doesn't need to be part of digitalpack.js
- Remove digitalpack.js

toggle mov


tudorraul:
:+1:

PR #74, opened 2015-07-06
Css cleanup

davidrapson:
This PR does a bunch of cleanup on the Subscriptions CSS to bring it more inline with Membership and .com in terms of tools and versions used.

- Add [autoprefixer](https://github.com/guardian/membership-frontend/pull/500) and [px_to_rem](https://github.com/guardian/membership-frontend/pull/604) in line with Membership and .com
- Move base styles into base/, save top-level for config files only.
- Merge components & modules. There's not difference between the two.
- Prefix all partials with `_` as per sass conventions.
- Add a dedicated mixins file to make it explicit where these come from (mixins should be kept to a minimum, so it's good to keep them all in one place for clarity)
- Upgrades guss components, depends on individual components e.g. `guss-webfonts` rather than entirety of `guss`

There's a lot of changes in here but it's *mostly* just a tidy up alongside a couple of tooling upgrades, will post a diff of the compiled output just to confirm that the output doesn't actually change that much. Happy  (and probably worth) chatting through the rationale of theses changes, and if needed to break it out into smaller PRs.

@afiore


davidrapson:
Build is failing but I can't see TC from where I am at the moment, think It's probably issues with the `bower.json` upgrades as I had a couple of version resolutions issues locally so I'm not sure I've got the versions set up quite right.

davidrapson:
Going to close this for now as some other work I'm doing might allow me to split this out into smaller PRs and may supersede some of this work.

PR #75, opened 2015-07-07
Remove ID dev token

tudorraul:
The token for Identity API CODE environment should be placed in local configuration file, `/etc/gu/subscription-frontend.conf`


davidrapson:
:+1: Does this the original API code need revoking now though?

tudorraul:
The CODE token is present in the S3 conf file which needs to be placed in the above mentioned location

PR #77, opened 2015-07-07
Css cleanup

davidrapson:
This PR is a simplified version of https://github.com/guardian/subscriptions-frontend/pull/74

- Move base styles into base/, save top-level for config files only.
- Merge components & modules. There's no difference between the two folders.
- Prefix all partials with `_` as per sass conventions.
- Add a dedicated mixins file to make it explicit where these come from (mixins should be kept to a minimum, so it's good to keep them all in one place for clarity)

This doesn't upgrade any guss components, I'll save that for another PR. It just cleans up the structure a bit.

@tudorraul @afiore


afiore:
:+1: fine by me :)

tudorraul:
:ship:

PR #79, opened 2015-07-07
Fix IE 8 Layout

davidrapson:
Fixes a major layout regression in IE 8 :disappointed_relieved: .

**Current**
screen shot 2015-07-07 at 15 13 11

**Fixed**
screen shot 2015-07-07 at 15 12 13


davidrapson:
Confirmed on PROD

screen shot 2015-07-07 at 15 35 23

PR #81, opened 2015-07-07
Update user details upon successful checkout

tudorraul:
Once the user has purchased a subscription, first name, last name and billing address are updated in ID API. 

This requires that the ID API grants `subscription-frontend` CRUD permissions to users':
`privateFields.firstName`,
`privateFields.secondName`,
`privateFields.billingAddress1`,
`privateFields.billingAddress2`,
`privateFields.billingAddress3`,
`privateFields.billingAddress4`,
`privateFields.billingCountry`,
`privateFields.billingPostcode`


app/services/IdentityService.scala
```diff
@@ -56,10 +42,37 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging
...
+      userId.id)
+    .map(_ => Unit)
+
 }
 
 object IdentityService extends IdentityService(IdentityApiClient) {
   class IdentityGuestPasswordError(jsonMsg: String) extends RuntimeException(s"Cannot set password for Identity guest user. Json response form service: $jsonMsg")
+
+  implicit class PersonalDataJsonSerialiser(personalData: PersonalData) {
+    implicit def convertToPrivateFields(): JsObject = JsObject(Map(
```
afiore: Why marking these as implicit if we call the method directly?

afiore: Never-mind.. I see this is an implicit class

afiore: ...still you can probably lose the implicit keyword next to each method definition

tudorraul:
@afiore @jennysivapalan can you have a look?

jennysivapalan:
:+1: :postbox:

tudorraul:
will wait for ID team to make the scope changes to our token on PROD and will be merging this then. cc @markjamesbutler

app/controllers/Checkout.scala
```diff
@@ -47,12 +48,13 @@ object Checkout extends Controller with LazyLogging {
...
-      userData => {
+      subscriptionData => {
         val idUserOpt = AuthenticationService.authenticatedUserFor(request)
-        CheckoutService.processSubscription(userData, idUserOpt).map {
+        CheckoutService.processSubscription(subscriptionData, idUserOpt).map {
           case CheckoutResult(_, guestUser: GuestUser, _) =>
             Some(FinishAccountForm().bind(guestUser.toFormParams))
-          case CheckoutResult(_, _:MinimalIdUser, _) =>
+          case CheckoutResult(_, registeredUser:MinimalIdUser, _) =>
+            updateStoredRegisteredUserDetails(subscriptionData.personalData, registeredUser.id, request.cookies.find(_.name == "SC_GU_U"))
```
afiore: Why not moving this bit of logic into `CheckoutService.processSubscription`? You could pass authCookie along with the other two arguments. I think it would help making the controller code a bit more concise.

tudorraul: Agreed. This can also work. Will refactor.

app/services/CheckoutService.scala
```diff
@@ -34,7 +35,19 @@ class CheckoutService(identityService: IdentityService, salesforceService: Sales
       userData <- userOrElseRegisterGuest
       memberId <- salesforceService.createOrUpdateUser(subscriptionData.personalData, userData.id)
       subscribeResult <- touchpointBackend.zuoraService.createSubscription(memberId, subscriptionData, ratePlan, paymentDelay)
-    } yield CheckoutResult(memberId, userData, subscribeResult)
+    } yield {
+      updateAuthenticatedUserDetails(subscriptionData.personalData, idUserOpt, authCookieOpt)
```
afiore: Should we move this IO call between line 35 and 36? This would allow to execute the request in parallel with the Zuora and salesforce call, perhaps gaining a bit of performance

afiore:
:+1: Fine by me

PR #82, opened 2015-07-07
[WIP] Update thank you page with relevant fields

afiore:
- Display personal data and subscription id in the thank you page
- Handle the form submission through an AJAX call (as per clickable wireframes).
- Also refactor checkout controller so that it handles the subscription form in a slightly more compact way.

out

@jennysivapalan @davidrapson @rtyley 

**Update**:  We looked into the possibility of making the thank you page persistent, but it looks like in order to do so we need to do some more analysis, and potentially change a few things in the Identity API. While we sort out these details, it would be probably a good idea to ship the thank you page as it is now.


conf/routes
```diff
@@ -20,7 +20,6 @@ POST        /digital/country                 controllers.DigitalPack.handleCount
 # Checkout
 GET         /checkout                        controllers.Checkout.renderCheckout
 POST        /checkout                        controllers.Checkout.handleCheckout
-GET         /checkout/thankyou               controllers.Checkout.thankyou
```
jennysivapalan: I think we should handle this route and order id could be a query parameter here.

app/controllers/Checkout.scala
```diff
@@ -45,35 +45,24 @@ object Checkout extends Controller with LazyLogging {
...
-    FinishAccountForm().bindFromRequest.fold( formWithErrors => {
-      Future {
-        BadRequest(views.html.checkout.thankyou(Some(formWithErrors)))
-      }
+    FinishAccountForm().bindFromRequest.fold({ _ =>
+      throw new RuntimeException("Could not set a password for guest account. Validation should happen on the client side!")
     }, guestAccountData => {
       IdentityService.convertGuest(guestAccountData.password, IdentityToken(guestAccountData.token))
-        .map(_ => Ok(views.html.checkout.alldone()))
+        .map(_ => Ok(view.alldone()))
```
jennysivapalan: I'm not a fan of the `alldone`. What are we displaying on the page? If it's thank you then I would opt for `views.html.checkout.thankyou`

conf/routes
afiore: I am afraid query parameters wouldn't cut it, as we now have a number of (nested) fields to display:

- subscription id
- personal data (i.e. name with surname, postal address, email address)
- when we create a guest user, also a password form (i.e. Identity user id and token)

If we really have a requirement for this page to have a permanent URL, than I am afraid we need to store the data above in a database record, or add significant code to fetch all the relevant bits of information from their respective source based on a single user identifier  (i.e email). Either ways, it seems to me that at this stage that would involve a fair amount of work and add little business value.

jennysivapalan: The subscription ID should be enough. We can use that and make a call to Zuora and Salesforce to pull back the relevant information. Most of the groundwork is done for making calls to Zuora and Salesforce so it should not be a great deal of work. 

It's a similar approach to Membership. Happy to pair on this tomorrow or later next week.

app/controllers/Checkout.scala
afiore: Yeah, that is due to some ambiguity in the visual specification of the guest user flow (I had referred to wireframes in Digital Pack v2, not to the more recent clickable wireframes). @davidrapson has agreed to pick up this bit and to replace the All done page with an Ajax endpoint.

conf/routes
afiore: Had a quick chat about this with @blackyjnz and @davidmcdowell, and they don't think there is a business requirement for a bookmarkable thank you page. Happy to implement this if you guys think this would add value from the technical standpoint. @rtyley @davidrapson @tudorraul any opinions?

davidrapson:
Just FYI you are likely to get some merge conflicts with this PR https://github.com/guardian/subscriptions-frontend/pull/88. @afiore Happy to pair to resolve them if needed.

conf/routes
rtyley: > The subscription ID should be enough. We can use that and make a call to Zuora and Salesforce to pull back the relevant information. Most of the groundwork is done for making calls to Zuora and Salesforce so it should not be a great deal of work.

What @jennysivapalan says makes sense to me - it's not too difficult to read subscription objects from Zuora, given the code we've already got. We _would_ need to verify that the subscription was for the current user, and not some other arbitrary user!

davidrapson:
Merged in latest changes from master and resolved conflicts form https://github.com/guardian/subscriptions-frontend/pull/88

conf/routes
afiore: Cool. @jennysivapalan an introduction to our Zuora setup would definitely help then...

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -135,13 +138,25 @@ define(['$',
...
+                    ajax({
+                        url: action,
+                        method: 'POST',
+                        data: formData
+                    }).then(function () {
+                        $FINISH_ACCOUNT_FORM.hide();
+                        $ALL_DONE.show();
+
+                    }).catch(function (jsonError) {
+                        console && console.error("Error while submitting guest user password form:", jsonError);
```
davidrapson: Can we do something other than console log here? Display something to the user maybe? Probably good to capture the exception with Raven too.

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -135,13 +138,25 @@ define(['$',
...
+                    var action = $FINISH_ACCOUNT_FORM.attr('action');
+                    var formData = ajax.reqwest.serialize($FINISH_ACCOUNT_FORM[0]);
+
+                    ajax({
+                        url: action,
+                        method: 'POST',
+                        data: formData
+                    }).then(function () {
+                        $FINISH_ACCOUNT_FORM.hide();
+                        $ALL_DONE.show();
```
davidrapson: Can we toggle a `is-hidden` class here instead? `show()/hide()` typically has some performance implications I'd rather avoid.

app/views/checkout/thankyou.scala.html
```diff
@@ -81,9 +82,15 @@ <h3 class="section-divider">Register for an account</h3>
                     <div class="actions">
                         <button class="button button--primary button--large js-checkout-finish-account-submit">Confirm</button>
                     </div>
-
                 </form>
-            }
+                <div class="js-all-done" style="display: none">
```
davidrapson: Use `is-hidden`. (If we need to do kind of contextual content more often we should possibly consider using client-side templates instead)

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -135,13 +138,25 @@ define(['$',
...
-                    bean.fire($FINISH_ACCOUNT_SUBMIT[0], 'click');
+                    var action = $FINISH_ACCOUNT_FORM.attr('action');
+                    var formData = ajax.reqwest.serialize($FINISH_ACCOUNT_FORM[0]);
+
+                    ajax({
+                        url: action,
+                        method: 'POST',
+                        data: formData
+                    }).then(function () {
+                        $FINISH_ACCOUNT_FORM.hide();
```
davidrapson: Can we toggle a is-hidden class here instead?

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -1,12 +1,13 @@
 define(['$',
         'bean',
         'modules/checkout/form-elements',
-        'modules/checkout/validations'
-    ],
+        'modules/checkout/validations',
+        'utils/ajax'],
```
davidrapson: Minor coding style point, probably move this up to after `bean` but before `modules/`

assets/javascripts/modules/checkout/form-elements.js
```diff
@@ -47,10 +47,12 @@ define(['$'], function ($) {
...
         $SMALLPRINT: $('.js-checkout-smallprint'),
         $FINISH_ACCOUNT_SUBMIT: $('.js-checkout-finish-account-submit'),
+        $FINISH_ACCOUNT_FORM: $('.js-finish-account'),
         $FIELDSET_YOUR_DETAILS: $('.js-fieldset-your-details'),
         $FIELDSET_PAYMENT_DETAILS: $('.js-fieldset-payment-details'),
         $FIELDSET_REVIEW: $('.js-fieldset-review'),
         $EDIT_YOUR_DETAILS: $('.js-edit-your-details'),
-        $EDIT_PAYMENT_DETAILS: $('.js-edit-payment-details')
+        $EDIT_PAYMENT_DETAILS: $('.js-edit-payment-details'),
+        $ALL_DONE: $('.js-all-done')
```
davidrapson: This selector name feels fairly ambiguous?

assets/javascripts/modules/checkout/checkout.js
```diff
@@ -135,13 +138,25 @@ define(['$',
...
-                    bean.off($FINISH_ACCOUNT_SUBMIT[0], 'click');
-                    bean.fire($FINISH_ACCOUNT_SUBMIT[0], 'click');
+                    var action = $FINISH_ACCOUNT_FORM.attr('action');
+                    var formData = ajax.reqwest.serialize($FINISH_ACCOUNT_FORM[0]);
+
+                    ajax({
+                        url: action,
+                        method: 'POST',
+                        data: formData
+                    }).then(function () {
```
davidrapson: Do we not need to check anything about the response here to make sure the registration was actually successful?

afiore: The `reqwest` library will automatically call the promise's `catch` method if our endpoint returns a non success status code. We don't need to get any information from the response except from knowing that the response was successful.

davidrapson:
:+1: With the latest changes

afiore:
Verified on prod and password form submission works fine. However, we now have a regression with the password strength indicator which is failing to display. Also Sentry error reporting seems to be broken or misconfigured.

screen shot 2015-07-13 at 12 08 11

PR #73, opened 2015-07-06
Fix label/form field association on payment form

davidrapson:
Spotted whilst working on https://github.com/guardian/subscriptions-frontend/pull/71, some form labels had no or broken association with their respective fields. This PR adds missing ids.

labels mov

@tudorraul

PR #85, opened 2015-07-08
Display address line 2 as optional

davidrapson:
This displays address line 2 as optional and fixes some missing label associations originally highlighted in https://github.com/guardian/subscriptions-frontend/pull/73

screen shot 2015-07-08 at 14 13 18

**Note:** This assumes that an empty value is posted to the service as an empty string. Membership does this the same way i.e., address2 not an Option. Current assumption is that this passes through to the service OK.

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -101,24 +101,24 @@
                         </div>
                         <div class="js-checkout-full-address">
                             <div class="form-field js-checkout-house">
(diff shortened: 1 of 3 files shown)


TesterSpike:
:+1: 

though the <edit> links are no longer showing

davidrapson:
@TesterSpike Seems this is already the case in master so fixed in another PR https://github.com/guardian/subscriptions-frontend/pull/89 . Thanks for spotting it though!!

PR #86, opened 2015-07-08
Global cleanup

davidrapson:
Bit of global cleanup.

## Footer

The way the footer messages were being modelled meant that HTML had to be passed around on all pages and also meant that some pages had no FAQ information in the footer. This PR just flags the few pages that need is as being international and then shows the appropriate FAQ message. Also means that the default message is shown on all pages.

screen shot 2015-07-08 at 16 19 57
screen shot 2015-07-08 at 16 20 08

## Normalize.css

- Adds base styles for HTML5 elements, now that footer uses, erm,  `<footer>`
- Also, just good to include as a decent baseline

## Old IE scripts

- Updates the check for old IE scripts. Previously was `<!--[if IE 9]>` (only IE9) which is true for Membership but should really be `<!--[if lte IE 9]>` here.
- Adds `html5shiv` for <IE9
- Removes the `.min` from `pollyfils`, just to match convention of other files. All are minified anyway.

@afiore


davidrapson:
@jennysivapalan

app/configuration/Links.scala
```diff
@@ -0,0 +1,10 @@
+package configuration
+
+object Links {
+  val guardianSubscriptionFaqs = LinkConfig(
+    "http://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions",
+    "Frequently Asked Questions"
+  )
+}
+
+case class LinkConfig(href: String, title: String)
```
afiore: You could expose `guardianSubscriptionFaqs` as a static method on Link (this would allow us to create a single type instead of two).

```scala

case class Link(href: Stirng, title: String)

object Link {
   val guardianSubscriptionFaqs = Link(
      "http://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions",
      "Frequently Asked Questions")
}
```

davidrapson: @afiore Nice thanks!

app/views/main.scala.html
```diff
@@ -6,7 +6,12 @@
     Currently being used for digital pack page and to help transition to a consistent
     body width throughout the site. The extra classes will be removed when this is the case.
 *@
-@(title: String, jsVars: JsVars = JsVars.default, footer: Html = Html(""), bodyClasses: Seq[String] = Nil)(content: Html)
+@(
+    title: String,
+    jsVars: JsVars = JsVars.default,
+    isInternational: Boolean = false,
```
afiore: Not crucial, but as we keep adding new arguments to this function, it probably would be good to group related parameters into a single object.

```scala
case class PageConfig(
         title: String, 
         isInternation: Boolean,
         bodyClasses: Seq[String] = Seq.empty
) 

object PageConfig {
   def withTitle(title: String) = ???
    val default = ???
}
```

Something like this would help us keep the signature of `@main` short and simple: 

```
   @main(pageConfig: PageConfig  = PageConfig.default,
                jsVars: JsVars = JsVars.default)
```

davidrapson: Yep definitely :smile: . It's on my list of things to do, we do the same on Membership fwiw.

afiore:
:+1:

PR #87, opened 2015-07-09
Fix csrf bug on country selection form

afiore:
Avoid caching [page]( https://subscribe.theguardian.com/digital/country) with country selection form.

@jennysivapalan


jennysivapalan:
:+1:

afiore:
Verified on prod and the CSRF bug seems to be fixed!

PR #90, opened 2015-07-09
Form styling refinements

davidrapson:
Refines form styling a little further to be closer to the designs. The one outstanding area is Review and confirm section which is still in need of some love.

screen shot 2015-07-09 at 12 15 04
screen shot 2015-07-09 at 12 16 24

@afiore @tudorraul


jennysivapalan:
:bowtie: :pencil2: :+1:

PR #92, opened 2015-07-09
Billing frequencies 

jennysivapalan:
Display option for different billing options and sign user up to appropriate plan based on their choice.

This isn't ready to merge as I need to work with someone on the layout. @davidrapson if you have time tomorrow that would be great. Currently I've just added a select drop down but it's needs to display the options as a pop up in the You've chosen section.

~~I've added as a config the price of each plan. This isn't great as if the prices in Zuora change it won't be reflected here. The other option is to call Zuora to get the pricing on start up of an instance, is it worth doing @rtyley ?~~

Prices is now loaded in on start up (refresh time is daily). Product plans are restricted to Digital Product, rate plans are restricted to British pounds and billing frequency of monthly, quarterly and annually. I've added a healthcheck to ensure there are products fetched otherwise users will not be able to sign up to a plan.

Finally the thankyou page needs to display the rate plan choice but I'd prefer to wait for https://github.com/guardian/subscriptions-frontend/pull/82


rtyley:
> I've added as a config the price of each plan. This isn't great as if the prices in Zuora change it won't be reflected here. The other option is to call Zuora to get the pricing on start up of an instance, is it worth doing @rtyley ?

The first option is basically what we do in Membership, but the 2nd option is better. In Membership the app servers did stuff that _wasn't_ dependent on Zuora (sell events), so it made sense to allow app servers to start up without having to connect to Zuora first - but subscribe.theguardian.com is all about Zuora, so reading prices at startup sounds legit to me.

jennysivapalan:
Okay - I'll work on that now

jennysivapalan:
Closing this until I sort the prices out

jennysivapalan:
Have now added in the work to load the Digital product prices and billing frequencies on start up. Refresh time is set to a day.

app/services/ZuoraService.scala
```diff
@@ -27,8 +30,69 @@ class ZuoraService(zuoraApiConfig: ZuoraApiConfig) extends ZuoraApi {
...
+
+        if (billingFreq.isDefined && productRatePlanChargeTier.isDefined)
+          SubscriptionProduct(Digital, billingFreq.get, ratePlan.id, productRatePlanChargeTier.get.price)
+        else throw new ZuoraServiceError(s"Could not find valid ratePlan details for Digital Pack. Rate plan ID requested ${ratePlan.id}")
+      }
+      subscriptionProducts
+    }
+  }
+
+  val productsTask = ScheduledTask[Seq[SubscriptionProduct]]("Loading rate plans", Nil, 5.seconds, 1.day)(getProducts())
```
jennysivapalan: We have to wait for Zuora to authenticate so I've set this to fetch 5 seconds after. We expect the plans to rarely vary so I've set the refresh time to 1 day.

app/services/ZuoraService.scala
```diff
@@ -27,8 +30,69 @@ class ZuoraService(zuoraApiConfig: ZuoraApiConfig) extends ZuoraApi {
...
+        val productRatePlanChargeTier = billingFreqProductRatePlan.flatMap { p =>
+          productRatePlanChargesTierPounds.find(_.productRatePlanChargeId == p.id)
+        }
+
+        val billingFreq = billingFreqProductRatePlan.flatMap(a => BillingFrequency.all.find(a.billingPeriod.toLowerCase == _.lowercase))
+
+        if (billingFreq.isDefined && productRatePlanChargeTier.isDefined)
+          SubscriptionProduct(Digital, billingFreq.get, ratePlan.id, productRatePlanChargeTier.get.price)
+        else throw new ZuoraServiceError(s"Could not find valid ratePlan details for Digital Pack. Rate plan ID requested ${ratePlan.id}")
+      }
```
jennysivapalan: Is there a better way to do the above check?

app/controllers/Healthcheck.scala
```diff
@@ -1,12 +1,15 @@
...
 object Healthcheck extends Controller {
 
-  def index = NoCacheAction {
-    Ok("""{"status":"OK"}""").withHeaders(CONTENT_TYPE -> "application/json")
+  def index = Action {
+    Cached(1) {
+      if(TouchpointBackend.Normal.zuoraService.products.size > 0) Ok("OK")
+      else ServiceUnavailable("Service Unavailable")
+    }
   }
```
jennysivapalan: I've followed membership-frontend for the healthcheck. We need to ensure there are product plans fetched from Zuora or the user won't be able to subscribe

davidrapson:
@jennysivapalan 

mirror mov

screen shot 2015-07-13 at 16 11 47

Updated the payment options to live in the sidebar and added toggle / live update behaviour.

jennysivapalan:
Oooh nice, thanks @davidrapson

TesterSpike:
The front end looks okay but I notice that the amounts are not showing all of the digits, £140.0 rather than £140.00
I'll be checking the back end integration with Jason this afternoon

app/controllers/Healthcheck.scala
```diff
@@ -1,12 +1,15 @@
...
 import play.api.mvc._
+import services.TouchpointBackend
 
 object Healthcheck extends Controller {
 
-  def index = NoCacheAction {
-    Ok("""{"status":"OK"}""").withHeaders(CONTENT_TYPE -> "application/json")
+  def index = Action {
+    Cached(1) {
+      if(TouchpointBackend.Normal.zuoraService.products.size > 0) Ok("OK")
```
afiore: Slight style improvement here could be to write: `zuoraService.products.nonEmpty`. Also, are we sure it is OK to drop the `application/json` content type?

app/services/ZuoraService.scala
```diff
@@ -4,18 +4,21 @@ import com.gu.membership.salesforce.MemberId
...
 import org.joda.time.Period
 import services.zuora.Subscribe
 import utils.ScheduledTask
 
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.concurrent.duration._
 
-class ZuoraService(zuoraApiConfig: ZuoraApiConfig) extends ZuoraApi {
+class ZuoraService(zuoraApiConfig: ZuoraApiConfig, digitalProductId: String) extends ZuoraApi {
```
afiore: Is there a reason why `digitalProductId` is a string and not a `ProductPlan`? If they are the same thing, it would be probably safer and more informative to use the latter

jennysivapalan: Good point - the digitalProductId should be assigned to the ProductPlan. That would be nicer.

app/controllers/Healthcheck.scala
jennysivapalan: I think so, the only think checking the healthcheck is the ELB which checks for a OK response. 
Good spot - will change to `nonEmpty`

app/services/ZuoraService.scala
```diff
@@ -27,8 +30,69 @@ class ZuoraService(zuoraApiConfig: ZuoraApiConfig) extends ZuoraApi {
...
+  def createSubscription(memberId: MemberId, data: SubscriptionData, paymentDelay: Option[Period]): Future[SubscribeResult] = {
+    request(Subscribe(memberId, data, paymentDelay))
   }
+
+  private def getProducts(): Future[Seq[SubscriptionProduct]] = {
+
+    def productRatePlans: Future[Seq[ProductRatePlan]] =
+      query[ProductRatePlan](s"ProductId='$digitalProductId'")
+
+
```
afiore: Do we need to execute several HTTP calls to Zuora to extract a list of `SubscriptionProduct`? Is there a way to do this in a single task? Perhaps it is worth spending some time figuring out a better way of doing this using the Zuora query language

app/services/ZuoraService.scala
afiore: I would push both the initial delay and the refresh frequency into the configuration settings, so we can easily tweak them to different environments

app/services/ZuoraService.scala
jennysivapalan: Yeah maybe but I think it's a fair bit of work. We'd have to understand what queries we need to make to Zuora and then also how to extract it to something meaningful. I don't think this is too much overload but perhaps it will be when more products are introduced.

jennysivapalan: Although thinking about it might be possible to pass a range of ID's for one select query and get a list of results. I will test it out.

app/touchpoint/TouchpointBackendConfig.scala
```diff
@@ -3,9 +3,8 @@ package touchpoint
 import com.gu.membership.salesforce.SalesforceConfig
 import com.gu.membership.zuora.ZuoraApiConfig
 import com.typesafe.scalalogging.LazyLogging
-import touchpoint.Product.Digital
 
-case class TouchpointBackendConfig(salesforce: SalesforceConfig, zuora: ZuoraApiConfig, productRatePlan: ProductRatePlan)
+case class TouchpointBackendConfig(salesforce: SalesforceConfig, zuora: ZuoraApiConfig, digitalProductId: String)
```
afiore: Also here we could use the `ProductPlan` type instead of string.

davidrapson:
Updated the price formatting (copied the Prices implicits from Membership) and removed a bit of repetition in the template.

screen shot 2015-07-14 at 15 46 05

app/services/ZuoraService.scala
tudorraul: I see two possible issues with setting a delay: 1/ setting this too high will make it likely that the application will start serving requests before this data is available and 2/ if it is too low, the credentials would not have been obtained when this call is attempted which will result in the application failing for a whole day.

Perhaps it's worthwhile looking into how this could be chained to the retrieval of credentials (that this call depends on).

jennysivapalan: I haven't looked into this yet but mainly as I want to see if this will work as-is in production. 

We can alter the configuration times, I've set a 5 second delay from the auth request. I've added to the healthcheck that the list of products must be nonEmpty so an instance won't be able to serve requests until it passes this check.

app/services/ZuoraService.scala
jennysivapalan: Done!

app/touchpoint/TouchpointBackendConfig.scala
jennysivapalan: Done

app/services/ZuoraService.scala
jennysivapalan: Done!

TesterSpike:
:+1:

davidrapson:
:shipit:

jennysivapalan:
Not true @prout-bot - I had to revert :cry:

PR #93, opened 2015-07-09
Country select page design

davidrapson:
Update country select to match latest designs. There is still some work to do around responsive styling for the current selection panel, but I want to pick this up with Ben W on Monday.

screen shot 2015-07-09 at 16 49 14

screen shot 2015-07-09 at 16 41 50
screen shot 2015-07-09 at 16 42 33

From a technical point of view, the most controversial part is the custom select styling. I typically don't recommend custom select styling but Filament group [have a solution](https://github.com/filamentgroup/select-css) which looks pretty solid across browsers and fallsback to default selects on unsupported browsers so it feels OK to use. Does rely on complex CSS browser hacks but they are well documented and tests so it's worth considering if it's worth the complexity.

@tudorraul @afiore

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -12,13 +12,11 @@
     bodyClasses = List("is-wide")
 ) {
 
-<div class="gs-container">
+<main class="gs-container">
 
-    <div class="checkout-container">
+    <section class="checkout-container">
 
-        <div class="checkout-header">
-            <h1 class="checkout-header__title">Checkout</h1>
-        </div>
+        @fragments.checkout.checkoutHeader("Checkout")
 
         <div class="checkout-container__sidebar">
             @fragments.checkout.basketPreview()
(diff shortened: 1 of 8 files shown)


afiore:
:+1: I like the new select input look and feel. How big is the original css file from Filament group? Wouldn't be better to manage it via bower? While adding some extra bytes to our CSS, perhaps it would protect us from styling regressions that would otherwise be hard to spot.

davidrapson:
@afiore Ah, it's not really bowerable kind of CSS, it's just a pattern that can be used. It was more just a comment about the potential complexity of the CSS, feels OK though.

davidrapson:
Ran this through a cross browser test and everything looks pretty good. Things fallback pretty well. Managed to get the browser hacks down to an absolute minimum so I'm pretty happy with this.

Chrome: Looks lovely naturally

chrome

iOS: Good

ios-8

Android 4.4: Good

android-4 4

Firefox: Good

firefox

IE 10: OK, icon a little fuzzy but fine

ie10

IE 9: OK, select styled but shows native select arrow

ie-9

@TesterSpike You happy with this a representative set of devices?

PR #94, opened 2015-07-09
Isolate digital pack landing page

davidrapson:
The digital pack landing page is in need of refactor. I'm aware this page was built quickly so completely understandable.

Unfortunately this means that the template and styling is very tightly coupled to that page and is pretty divorced from the rest of the sites patterns, lots of specific overriding going on, and the page is visually broken on IE. So definitely in need of some love.

**IE 9**
screen_shot_2015-07-09_at_14 01 24

This PR just merges fragments (some in fragments/ others just in views/...) for the digital pack landing page and puts them inline in the template, as prep for future refactor.

@rtyley


rtyley:
LGTM :+1:

PR #95, opened 2015-07-09
Implement email marketing preferences

afiore:
Add an opt-in email marketing checkbox to the checkout page.

 
screen shot 2015-07-09 at 18 30 13

Also pushes some data transformation code from the Checkout controller into the model package.

~~@tudorraul some further wiring needs to be done once #81 gets merged~~

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -22,7 +22,6 @@ object Checkout extends Controller with LazyLogging {
     getIdentityUserByCookie.map { idUserOpt =>
       val form = idUserOpt.map { idUser =>
         val data = SubscriptionData.fromIdUser(idUser)
-        logger.debug(s"validation failed. Rendering data: $data")
         SubscriptionsForm().fill(data)
       } getOrElse {
         SubscriptionsForm()
@@ -36,6 +35,7 @@ object Checkout extends Controller with LazyLogging {
     SubscriptionsForm().bindFromRequest.fold(
       formWithErrors => {
         getIdentityUserByCookie.map { idUserOpt =>
+          logger.error(s"Backend form validation failed. Please make sure that the front-end and the backend validations are in sync (validation errors: ${formWithErrors.errors}})")
           BadRequest(view.payment(formWithErrors, userIsSignedIn = idUserOpt.isDefined))
         }
       },
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -1,43 +1,32 @@
 package controllers
 
 import actions.CommonActions._
-import com.gu.identity.play.{IdUser, PrivateFields}
+import com.gu.identity.play.IdUser
 import com.typesafe.scalalogging.LazyLogging
 import forms.{FinishAccountForm, SubscriptionsForm}
-import model.PersonalData
+import model.SubscriptionData
 import play.api.libs.json._
 import play.api.mvc._
 import services.CheckoutService.CheckoutResult
 import services.{CheckoutService, _}
 import views.html.{checkout => view}
 import configuration.Config.Identity.webAppProfileUrl
-
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
 object Checkout extends Controller with LazyLogging {
+  def identityCookieOpt(implicit request: Request[_]): Option[Cookie] =
+    request.cookies.find(_.name == "SC_GU_U")
 
   def renderCheckout = GoogleAuthenticatedStaffAction.async { implicit request =>
-    for (idUserOpt <- getIdentityUserByCookie(request)) yield {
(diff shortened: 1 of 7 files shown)


app/views/checkout/payment.scala.html
```diff
@@ -120,23 +120,20 @@ <h2 class="fieldset__headline">
...
-                                    <input type="checkbox" class="js-input">
-                                </span>
-                                <span class="checkbox__label">Email</span>
-                            </label>
-                            <label class="checkbox">
-                                <span class="checkbox__input">
-                                    <input type="checkbox" class="js-input">
-                                </span>
-                                <span class="checkbox__label">Post</span>
+                            <div class="checkbox">
```
davidrapson: Can you use the same markup pattern used above for this?

jennysivapalan:
:+1: :email:

afiore:
@ostapneko and I have added a few more updates to PR. @tudorraul mind having a final look?

app/views/checkout/payment.scala.html
```diff
@@ -140,23 +140,16 @@ <h2 class="fieldset__headline">
...
-                                    <input type="checkbox" class="js-input">
-                                </span>
-                                <span class="checkbox__label">Email</span>
-                            </label>
-                            <label class="checkbox">
-                                <span class="checkbox__input">
-                                    <input type="checkbox" class="js-input">
-                                </span>
-                                <span class="checkbox__label">Post</span>
+                            <label class="label checkbox" for="receive-gnm-marketing">
```
davidrapson: Original comment has been outdated:

Can you use the same markup pattern used above for this?

```html
<label class="checkbox">
  @defining(form.data.get("personal.receiveGnmMarketing")) { gnmMarketingOpt =>
    <span class="checkbox__input">
      <input type="checkbox"
             name="personal.receiveGnmMarketing"
             @(if (gnmMarketingOpt.contains("true")) "checked" else "")
             value="@gnmMarketingOpt">
    </span>
  }
  <span class="checkbox__label">
    I'd like to be kept up to to date by email with offers and developments from Guardian News &amp; Media Limited
  </span>
</label>

```

app/controllers/Checkout.scala
```diff
@@ -1,43 +1,32 @@
...
-          idUserData("personal.address.address1", _.billingAddress1) ++
-          idUserData("personal.address.address2", _.billingAddress2) ++
-          idUserData("personal.address.town", _.billingAddress3) ++
-          idUserData("personal.address.postcode", _.billingPostcode)
-        ).toMap
-      )
+    getIdentityUserByCookie.map { idUserOpt =>
+      val form = idUserOpt.map { idUser =>
+        val data = SubscriptionData.fromIdUser(idUser)
+        logger.debug(s"validation failed. Rendering data: $data")
```
tudorraul: is this right? if the execution reaches this point it must mean there has been a validation error?

app/model/SubscriptionData.scala
```diff
@@ -1,15 +1,50 @@
 package model
 
+import com.gu.identity.play.IdUser
+
 case class PaymentData(account: String, sortCode1: String, sortCode2: String, sortCode3: String, holder: String) {
   val sortCode = s"$sortCode1$sortCode2$sortCode3"
 }
+object PaymentData {
+  val blank = PaymentData("", "", "", "", "")
```
tudorraul: this value is only used once; can its definition be moved where it is being used?

app/controllers/Checkout.scala
afiore: This is for cases where input passes front-end validation but not back-end validation (hence `BadRequest`). Perhaps we should log as warn instead of debug

app/services/IdentityService.scala
```diff
@@ -43,35 +41,42 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging
...
+      createOnlyFields.foldLeft(personalData.convertToUser()) { (map, field) => map - field }
+
+    identityApiClient.updateUserDetails(updatedFields, userId, authCookie).map(_ => Unit)
+  }
 }
 
 object IdentityService extends IdentityService(IdentityApiClient) {
+  val primaryEmailAddress = "primaryEmailAddress"
+  val publicFields = "publicFields"
+  val createOnlyFields = Seq(primaryEmailAddress, publicFields)
```
tudorraul: these fields can be converted to local variables, since they are only used once. It can make the code more readable if everything is one place.

tudorraul:
:+1: bar the last comment

app/services/IdentityService.scala
afiore: We can make them private, but they are actually used twice (in the definition of  `createOnlyFields` and in `convertToPrivateFields `)

afiore:
Verified on prod and the email preference are persisted correctly in Identity

PR #97, opened 2015-07-10
Improve JavaScript fallbacks

davidrapson:
Improves how JavaScript loading is handled and adds unsupported messages

- Splits `<head`> JavaScript into firstSteps and laterSteps, we want to be loading `main.js` later than CSS, so this introduces that separation between the two stages of scripts.
- Adds no-js and browser unsupported messages.

screen shot 2015-07-10 at 15 33 21
screen shot 2015-07-10 at 15 33 42

@rtyley


rtyley:
Love it :+1:

PR #98, opened 2015-07-10
fix prepopulation of last name field on the checkout page

tudorraul:
restored :sparkles:


tudorraul:
@afiore

afiore:
:+1: Well spotted!

PR #102, opened 2015-07-13
Update production to point to UAT

jennysivapalan:
This requires an update to the private config file in S3 before merging.

I've renamed `touchpoint.PROD.conf` to `touchpoint.UAT.conf`. Once we have the production credentials we can create the relevant file.


dominickendrick:
:+1:

PR #103, opened 2015-07-14
Fix Direct Debit copy, clean up notices

davidrapson:
Direct Debit copy supplied had a duplicated item it, this PR removes that and cleans up the notice markup a bit to make the template more readable.

screen shot 2015-07-14 at 11 55 27

screen shot 2015-07-14 at 11 59 11

@afiore @tudorraul


tudorraul:
:+1:

PR #106, opened 2015-07-15
Extract colours from gs-container, use separate class

davidrapson:
`.gs-container` classes should ideally not have any non-layout styles on them, this was causing some inflexibility in how we do layouts so it's worth sorting.

This PR adds a separate container to determine colour and leaves the `.gs-container` class for constraining layout only. I've also updated all wrapping container to use `<main>`.

*(Also removes an unused view)*

@TesterSpike Would you mind giving this one a once through. If i've missed a page then you will see a grey background for the whole page instead of a white container.

@afiore @tudorraul


app/views/checkout/alldone.scala.html
```diff
@@ -1,17 +0,0 @@
-@import configuration.Config.Identity.webAppProfileUrl
```
tudorraul: was this template not used?

davidrapson: Not anymore, no.

tudorraul:
looked to be a bigger change initially, but I think that was due to the diffing algorithm.
:+1: LGTM, bar the comment

TesterSpike:
:+1:

PR #107, opened 2015-07-15
Checkout tests

ostapneko:
The plan is to add unit tests to 
- [x] `IdentityService`
- ~~[ ] `SalesforceService`~~

And then mock these services to test
- [x] `CheckoutService`

### DIFF
--- a/app/services/IdentityService.scala
+++ b/app/services/IdentityService.scala
@@ -27,14 +27,12 @@ class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging 
       (response.json \ "user").asOpt[IdUser]
     }
 
- def registerGuest(personalData: PersonalData): Future[GuestUser] = {
-    val json = personalData.convertToUser()
-    identityApiClient.createGuest(json).map(response => response.json.as[GuestUser])
+  def registerGuest(personalData: PersonalData): Future[GuestUser] = {
+    identityApiClient.createGuest(personalData).map(response => response.json.as[GuestUser])
   }
 
   def convertGuest(password: String, token: IdentityToken): Future[Unit] = {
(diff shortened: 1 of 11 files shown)


app/services/IdentityService.scala
```diff
@@ -14,7 +14,7 @@ import play.api.libs.ws.{WS, WSRequest, WSResponse}
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
-class IdentityService(identityApiClient: IdentityApiClient) extends LazyLogging {
+class IdentityService(identityApiClient: => IdentityApiClient) extends LazyLogging {
```
afiore: Is there a reason why we are passing in `IdentityApiClient` by name?

test/services/CheckoutServiceSpec.scala
```diff
@@ -0,0 +1,112 @@
...
+      def process(service: CheckoutService): Future[CheckoutResult] =
+        service.processSubscription(subscriptionData, None, None)
+
+      val checkoutResult = Await.result(
+        process(service),
+        2.seconds
+      )
+
+      "does not update the user details in the Identity service" in {
+        var updated = false
```
afiore: Checking for the effect by mutating a var is not great, but I guess we have no better way of doing it...

afiore:
:+1: Looking good!

app/services/IdentityService.scala
ostapneko: To be able to replace it by `???` in the tests

test/services/CheckoutServiceSpec.scala
ostapneko: I'm open to alternatives

PR #108, opened 2015-07-15
Review details styling, add terms links

davidrapson:
Refines review details section styling and adds links to terms of service and privacy policy.

screen shot 2015-07-15 at 12 31 49

@tudorraul @afiore

### DIFF
--- a/app/configuration/Links.scala
+++ b/app/configuration/Links.scala
@@ -3,6 +3,14 @@ package configuration
 case class Links(href: String, title: String)
 
 object Links {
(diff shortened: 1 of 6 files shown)


jennysivapalan:
:+1: :scissors: :scroll:

TesterSpike:
Clicking on the links take you away from the subscription page which clears the details already entered

davidrapson:
@TesterSpike Good point about the links. 2nd point I agree with, but I'm going to look at improving that in another PR as I think that JS needs a rework.

PR #109, opened 2015-07-15
Bypass Google staff Auth via cookie

afiore:
~~@TesterSpike Can you try this locally? This will switch off google auth in dev, so should give you a way to get started writing end-to-end tests. @rtyley @tudorraul Would you mind having a look at this; I am not really sure about the values I am setting in `devUserIdentity`.~~

**Update:** 

After some discussion with @rtyley we have decided to go the whole way and implement a cookie-based mechanism to bypass pre-release pages. Together with implementing a custom Play action, this PR also updates the acceptance tests suite with a new feature demonstrating how to use the QA cookie.

**Todo**:

- [x] Add confidential key `qa.passthrough-cookie-value`
- [x] Add travis env var


app/actions/CommonActions.scala
```diff
@@ -12,7 +15,7 @@ object CommonActions {
 
   val GoogleAuthAction = OAuthActions.AuthAction
 
-  val GoogleAuthenticatedStaffAction = NoCacheAction andThen GoogleAuthAction
+  val GoogleAuthenticatedStaffAction = NoCacheAction andThen ProdOnlyAuth
```
tudorraul: Should this auth method be instead of or in addition to Google auth?

afiore: This replaces and wraps the old google auth action, which in turn was just calling `OAuthActions.AuthAction`

jennysivapalan:
This looks like a good solution to me

TesterSpike:
:+1: 
I did a very quick test and a web driver chrome browser was able to get to the checkout page without requiring a cookie or google auth

rtyley:
:+1:

PR #110, opened 2015-07-15
Fix password strength indicator

afiore:
Fix a recent regression whereby zxcvbn was not being imported in the module scope. 

- Drop `curl.js` in favour of a direct `require` call to include the library. 
- Extend `JsVars` so that a map of relevant hashed paths can percolate from the backend into the front-end code. 

@davidrapson


davidrapson:
@afiore Not hugely keen on this. We're you able to repeat the issues reliably before this change? Can we perhaps take another look at it together just in case we've missed something?

afiore:
I have definitely reproduced the bug locally and I am pretty sure the version of this code currently deployed on prod is not working (@TesterSpike can you confirm this). Happy to explore alternative ways of solving this, this is a solution which was simple enough for me to implement and didn't involve adding other libraries

afiore:
Superseded by #111

PR #111, opened 2015-07-16
Fix zxcvbn version in bower file

afiore:
This fixes a regression whereby the password strength indicator failed to load. The regression was caused by a recent release of  the zxcvbn library which wouldn't play well with RequireJS.

Investigated and solved in collaboration with @davidrapson


davidrapson:
:+1:

PR #112, opened 2015-07-16
Update configuration so PROD points to UAT endpoints

jennysivapalan:
A repeat of #102 . I've done some more testing and uploaded the config again so going to give this a whirl. :dash:


app/services/CheckoutService.scala
```diff
@@ -39,12 +40,26 @@ class CheckoutService(identityService: IdentityService, salesforceService: Sales
         identityService.registerGuest(subscriptionData.personalData)
       }
 
+    userOrElseRegisterGuest.onComplete(t => Logger.info("userOrElseRegisterGuest result "+t))
```
tudorraul: is this logging statement still needed? If so, perhaps we should put a more descriptive message.

jennysivapalan: No sorry - I should close the PR. This was me debugging all morning.

jennysivapalan:
Still doesn't work but mainly because Zuora is being slow to respond.

app/services/CheckoutService.scala
```diff
@@ -39,12 +40,26 @@ class CheckoutService(identityService: IdentityService, salesforceService: Sales
...
-      subscribeResult <- zuoraService.createSubscription(memberId, subscriptionData)
+      memberId <- {
+        val mi = salesforceService.createOrUpdateUser(subscriptionData.personalData, userData.id)
+        mi.onComplete(t => Logger.info("salesforce result "+t))
+        mi
+      }
+      subscribeResult <- {
+        val sr = zuoraService.createSubscription(memberId, subscriptionData)
+        sr.onComplete(t => Logger.info("zuora result " + t))
+        sr
```
tudorraul: are these logging statement still needed? if so this pattern can be simplified by using a map that does logging and returns the original object.

app/services/CheckoutService.scala
```diff
@@ -39,12 +40,26 @@ class CheckoutService(identityService: IdentityService, salesforceService: Sales
...
+      }
     } yield {
+      Logger.info(s"Identity user : ${userData.id}")
+      Logger.info(s"Salesforce user: ${memberId.salesforceContactId}")
+      Logger.info(s"Zuora: ${subscribeResult.id}")
```
tudorraul: loggings need?

PR #116, opened 2015-07-16
Refactor Digital pack layout

davidrapson:
This PR refactors the digital pack template to bring it in line with the rest of the site in terms of patterns and layout, and fixes some rendering issues.

- Extends page to widest breakpoint, avoids slowly increasing page-width as you go through the flow
- Uses common design patterns rather than unique styles for this template
- Models the benefits to avoid markup repetition
- Actually works in IE 9

screen shot 2015-07-16 at 12 45 01
screen shot 2015-07-16 at 12 45 06

**IE 9 before**
screen shot 2015-07-16 at 12 46 56

**IE 9 after**
screen shot 2015-07-16 at 12 46 32

@tudorraul @jennysivapalan


jennysivapalan:
:+1: :game_die:

PR #117, opened 2015-07-16
Use ScalaTest's whenReady instead of Await when testing futures

ostapneko:
The checkout service tests where using Await and were also causing timeouts in CI.

This PR

  1. Increase the timeout to avoid new failures
  1. Uses the ScalaTest `Futures` trait to handle future results instead of `Await`

### DIFF
--- a/test/services/CheckoutServiceSpec.scala
+++ b/test/services/CheckoutServiceSpec.scala
@@ -6,24 +6,28 @@ import com.gu.membership.zuora.soap.Zuora.{Authentication, SubscribeResult}
 import model.zuora.SubscriptionProduct
 import model.{PaymentData, PersonalData, SubscriptionData}
 import org.scalatest.FreeSpec
-import services.CheckoutService.CheckoutResult
+import org.scalatest.concurrent.{Futures, ScalaFutures}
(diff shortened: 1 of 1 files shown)


test/services/CheckoutServiceSpec.scala
```diff
@@ -5,24 +5,28 @@ import com.gu.membership.salesforce.{BasicMember, MemberId}
...
 
-class CheckoutServiceSpec extends FreeSpec {
+class CheckoutServiceSpec extends FreeSpec with Futures with ScalaFutures {
+  implicit override val patienceConfig =
+    PatienceConfig(timeout = scaled(Span(10, Seconds)), interval = scaled(Span(50, Millis)))
 
-  def makeIdentityService(onUserUpdate: => Unit = {}): IdentityService=
+  class UpdateFlag(var updated: Boolean = false)
+
+  def makeIdentityService(updateFlag: UpdateFlag): IdentityService=
```
afiore: The `updateFlag` argument could probably have a default; this would make the tests a bit more dry

afiore:
:+1: Looks good, needs rebasing though

PR #114, opened 2015-07-16
display subscription name on the checkout confirmation page

tudorraul:
### Before 
screen shot 2015-07-16 at 11 50 00

### After
screen shot 2015-07-16 at 11 47 54

PR #118, opened 2015-07-16
fix tests and build

tudorraul:


rtyley:
This is to fix the break included in https://github.com/guardian/subscriptions-frontend/pull/114 ?

rtyley:
:+1: once we've seen the build succeed!

PR #120, opened 2015-07-16
Account number server-side validation

tudorraul:
:hammer:


tudorraul:
waiting for further details

PR #121, opened 2015-07-16
Subscribe options

davidrapson:
Update subscribe pages to:

- Use links instead of forms
- Model subscription details rather than having large views
- All actions can now be cached

Updated functional tests as we are now using plain links instead of a form.

screen shot 2015-07-16 at 16 34 06

@rtyley @TesterSpike


afiore:
:+1: Well done for simplifying both controller and view code. I find that `model.SubscriptionDetails` and `model.SubscriptionOptions` could be modelled a bit better in terms of types (e.g. strings used for numerical values), but given that we have no logic relying on them other than rendering in the view is probably all right.

davidrapson:
@afiore Cheers. Yeah I wanted to take this a bit further (not a big fan of `packageType` and `method` being strings, but wanted to limit the scope a little bit.

davidrapson:
@afiore @rtyley Pushed even more into the model and changed the prices to numerical values. Mind taking another look?

app/views/shipping/shipping.scala.html
```diff
@@ -0,0 +1,57 @@
...
+            <ol class="steps">
+                @for(step <- subscription.steps) {
+                    <li class="steps__item">@step</li>
+                }
+            </ol>
+        </section>
+
+        <section class="section-slice">
+
+            @if(subscription.insideM25) {
```
davidrapson: I'd love to be able to do this based on whether the `subscription` is a `CollectionSubscriptionProduct` or a `DeliverySubscriptionProduct` but I couldn't work out how to do that. @rtyley

afiore: I guess what you are trying to do is to reuse the markup while parameterising the different values. You could do that by defining a [function](https://www.playframework.com/documentation/2.2.x/ScalaTemplates#Declaring-reusable-blocks) within your view

```scala
@displayProduct(sp: SubscriptionProduct) = {
  val x = if (sp.insideM25) "a" else "b"
  <div>
    <span x="$x">
  </div>

} 
```
However, looks a bit overkill given that we repeat the same block on once and within the same file.

davidrapson: Yeah, less concerned about the repetition of markup (I think that's fine), just about what the if statement was. Not a big thing though.

TesterSpike:
:+1:

davidrapson:
@afiore These additional changes look OK to you?

afiore:
:+1: fine by me

PR #123, opened 2015-07-16
Add git commit hooks for CSS/JS linting

davidrapson:
Add git commit hooks for CSS/JS linting, helps catch syntax errors early rather than in a build. Will need to run `setup.sh` again to copy the hooks over.

@tudorraul @rtyley


tudorraul:
:+1:

PR #126, opened 2015-07-17
Billing frequency with call to fetch product changes

jennysivapalan:
A repeat of #92 but we've change the call to fetch products to happen after a successful authentication to Zuora. https://github.com/guardian/subscriptions-frontend/commit/22a626e8a3958cccd7362a58a32d32e793f2c955

@rtyley could you give a quick look over https://github.com/guardian/subscriptions-frontend/commit/22a626e8a3958cccd7362a58a32d32e793f2c955 the rest of the PR was signed off as working okay by @TesterSpike on #92


rtyley:
In order to unblock this story, go ahead and merge it, but it really is using [a really nasty hack](https://github.com/guardian/subscriptions-frontend/commit/22a626e8a3958cccd7362a58a32d32e793f2c955#commitcomment-12227046) which I really don't want to see used in any code again! I'll try to get round to doing an alternative implementation when I can - I realise it is a tricky problem.

PR #128, opened 2015-07-20
Clean up password.js

davidrapson:
- Add to pattern library
- Remove from unneeded `components/` directory
- As we just pull out the actual DOM nodes anyway to use with `zxcvbn`, we can remove the need for `bean` or `$` here

password-strength mov

@afiore 

As an aside is there any reason we set a limit of `20` characters for passwords? @jennysivapalan


afiore:
:+1:

PR #129, opened 2015-07-20
Form semantic improvements

davidrapson:
- Fieldset titles should be `<legend>`s
- Move `errorMessage` out of views/ and into fragments, because it's,
  erm, a fragment
- Fix option styling for checkboxes  (classnames had changed)

@tudorraul @afiore


afiore:
:+1:

PR #130, opened 2015-07-20
Term start date should match ContractAcceptanceDate for the Subscribe call to Zuora

tudorraul:


tudorraul:
cc @jennysivapalan

jennysivapalan:
:+1:

PR #131, opened 2015-07-20
Checkout cleanup

davidrapson:
This code was handling two separate forms. Once for the payment flow, the other for the confirmation page. These have been split out into separate components.

In addition:

- Splits out checkout JS into smaller modules
- Removes long var alias list in checkout.js
- Adds some initial tests for validations (more to come)
- Small steps to making validations.js fully testable
- Improves the way the review details information is generated, includes tests.
- Removes `manualAddress` code, presumably for future address lookup?

Sorry for the rewrite, but these changes hopefully go some way towards making this checkout JS testable in unit tests so I feel is worth it.

Here's a GIF:

checkout mov

@afiore @rtyley @TesterSpike


TesterSpike:
:+1:

davidrapson:
@afiore Mind giving this the once over?

afiore:
:+1:

PR #132, opened 2015-07-20
Make payment details validation testable

davidrapson:
- Makes validations#validatePaymentDetails testable
- Adds unit test suite for validatePaymentDetails
- Refactors toggleFieldsets to handle error display there, only data validation in validatePaymentDetails

Based on https://github.com/guardian/subscriptions-frontend/pull/131

@afiore


assets/javascripts/modules/confirmation.js
```diff
@@ -0,0 +1,61 @@
...
+    var IS_HIDDEN = 'is-hidden';
+
+    function validatePassword(password) {
+        /**
+         * We have a component which explicitly
+         * does password strength validation.
+         * Why not use that to determine if we shoule submit??
+         * e.g., if zxcvbn says it's not weak?
+         */
+        return password.length >= 6;
```
afiore: Agree. Maybe we can extend the password module to expose a function that calls zxcvbn and returns a boolean

davidrapson: Yep, hopefully. Little bit more to refactor before that's worth doing, which is why I just left it as a comment.

assets/javascripts/modules/checkout/toggleFieldsets.js
```diff
@@ -0,0 +1,116 @@
...
+                    swap(formEls.$EDIT_YOUR_DETAILS, formEls.$EDIT_PAYMENT_DETAILS);
+                }
+            });
+        });
+    }
+
+    function toggleFieldsetsPayment() {
+        submitHelper(formEls.$PAYMENT_DETAILS_SUBMIT, function() {
+
+            var validity = validations.validatePaymentDetails({
```
afiore: Perhaps we could name these object keys so that they are consistent with the actual form input fields. These way we could just call `reqwest.serialize` on the relevant field set, and pass in the appropriate data to the validation function.

test/js/spec/test-main.js
```diff
@@ -1,18 +1,45 @@
...
 requirejs.config({
-    // Karma serves files from '/base'
-    baseUrl: '/base/test/js/spec',
+    /**
+     * Karma serves files from '/base'
+     */
+    baseUrl: '/base/assets/javascripts/',
 
-    // Keep these in sync with the paths found in the requireJs paths
+    /**
```
afiore: Is there anyway to factor these aliases out and include them in both `Gruntfile.js` and `test-main.js`. This would save some code duplication

assets/javascripts/modules/checkout/toggleFieldsets.js
davidrapson: :wink: You read my mind, already started this yesterday. Slowly moving that way.

test/js/spec/test-main.js
davidrapson: Yeah, had the same thought on Membership. Not a great way that I've  found so far, unless I can find a way to share a json file between the two.

afiore:
:+1:

PR #133, opened 2015-07-21
Implement test users endpoint

afiore:
Allow to switch touchpoint backends at runtime. This uses use a signed cookie as the main mechanism to distinguish test users from ordinary ones. 

Note that some of the acceptance tests fail as they cover functionality that is part of this PR and has not been deployed on prod yet.

Pair programmed with @jennysivapalan 

@rtyley

### DIFF
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -1,7 +1,7 @@
 package actions
 
 import com.typesafe.scalalogging.LazyLogging
-import configuration.QA.{passthroughCookie => qaCookie}
+import configuration.QA.passthroughCookie
 import controllers.{Cached, NoCache}
 import play.api.mvc._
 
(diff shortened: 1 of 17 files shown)


jennysivapalan:
This needs the PROD config in the S3 updating before merging.

app/actions/CommonActions.scala
```diff
@@ -22,17 +27,43 @@ object CommonActions {
...
       } else {
-        cookie.foreach(_ => logger.warn("Invalid QA Cookie supplied"))
+        qaCookie.foreach(_ => logger.warn("Invalid QA Cookie supplied"))
         OAuthActions.AuthAction.authenticate(request, block)
       }
     }
   }
 
+  class RequestWithCookies[A](request: Request[A]) extends WrappedRequest[A](request) {
+    val testUserCookie = request.cookies.get(testUserCookieName)
```
rtyley: The 'test user cookie' is troubling to me. The problem is that there may be other parts of the system (for instance CAS) that aren't driven by the user's browser, and will _still_ need to know if a user is a test user or not, and they won't have access to the user's cookie. I guess we're driven to this because rendering the Checkout page requires us to know the list of Zuora services, which requires knowing which Touchpoint backend we're talking to. So some kind of session state (e.g. this cookie) is necessary. It gets troubling when the user-identity (ie first name/email) gets out of sync with the value of the testUserCookie.

afiore: I am afraid we have totally overlooked CAS proxy while implementing this. I would park it for today as I need a bit more context in order to go ahead. Hopefully we can look into this together tomorrow.

rtyley:
@TesterSpike  this (commit 237cd6e) is my re-cut of this PR, with a quite a few changes from the work that @jennysivapalan & @afiore originally did.

There are few rough edges, but I think it can be merged, if anyone is brave enough to do it while I'm not around :-)

TesterSpike:
This appears to create test users in UAT zuaora (as the subscription codes start with 0005 rather than 0001) however the test user details are not automatically filled into the relevant forms when selecting either register user or guest user.

rtyley:
Have walked through this with @TesterSpike on my local dev machine, the branch performed fine when doing a Quarterly purchase with a 'Normal' user.

Will merge - note that we will  still be pointing Normal users to UAT in Prod when this goes out : https://github.com/guardian/subscriptions-frontend/blob/75242150/conf/PROD.conf#L15-L17

PR #127, opened 2015-07-17
Run acceptance tests as part of the travis build

afiore:
Update `.travis.yml` to include the new `acceptance-test` task

@rtyley

PR #134, opened 2015-07-21
Don't run acceptance-tests before an environment is available

rtyley:
The default Travis build is triggered when a PR is made or merged, but at that point there is no deployed environment that is actually running that code, so we don't want to run the acceptance tests (which drive a web browser) at that point. This change reverts part of https://github.com/guardian/subscriptions-frontend/pull/127.

There's an upcoming feature in @prout-bot which will trigger a Travis build post-deploy to run the acceptance-tests... cc @afiore

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ cache:
   directories:
   - $HOME/.sbt
   - $HOME/.ivy2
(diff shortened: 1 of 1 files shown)


afiore:
:+1: Agree. We will work out a way for Travis to execute the full test suite (including acceptance tests) for builds triggered by Prout.

PR #135, opened 2015-07-21
Fix basket preview displaying on country select screen

davidrapson:
Fix basket preview displaying on country select screen

**Before**
screen shot 2015-07-21 at 15 19 08

**After**
screen shot 2015-07-21 at 15 23 56

@afiore


ostapneko:
:+1: on the code front

TesterSpike:
:+1:

PR #137, opened 2015-07-21
Use buttons for country choice, not select

davidrapson:
In the spirit of of [turning things that shouldn't forms be into links](https://github.com/guardian/subscriptions-frontend/pull/121). It was really bugging me that we had a two option select box, having discussed with David M and we've agreed to turn this in to links. It was a short discussion. Means we can cache this page, and also means we can remove the gnarly custom select styling.

screen shot 2015-07-21 at 16 53 45

screen shot 2015-07-21 at 16 23 25

~~I need to confirm design with Ben W before this goes out, but the principal makes me happy.  Simplicity always wins.~~ Confirmed

screen shot 2015-07-21 at 17 05 55

cool-corgis

@rtyley @ostapneko


dominickendrick:
This is the best pull request I've seen in ages !!!! Simplicity for the win

jennysivapalan:
Agreed :+1: :scissors:

TesterSpike:
:+1: 
The only niggle is that the buttons touch when viewed on an android phone with chrome which is different from the other browsers.  This is still functional and looks acceptable.

PR #138, opened 2015-07-21
Pdf mandate

ostapneko:
Create a link to a mandate preview PDF, as described on the [Trello task](https://trello.com/c/5ohvsm8t/75-display-the-name-that-will-appear-on-the-customer-s-bank-statement-and-add-a-link-to-a-pdf-of-the-customers-direct-debit-mandate).

The link is currently unstyled:

mandate-1

It opens the PDF in a separate tab:

mandate

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -1,13 +1,14 @@
 package configuration
 
 import com.github.nscala_time.time.Imports._
+import com.gocardless.GoCardlessClient
+import com.gocardless.GoCardlessClient.Environment
 import com.gu.googleauth.GoogleAuthConfig
 import com.gu.identity.cookie.{PreProductionKeys, ProductionKeys}
 import com.gu.membership.salesforce.SalesforceConfig
 import com.netaporter.uri.dsl._
 import com.typesafe.config.ConfigFactory
 import net.kencochrane.raven.dsn.Dsn
-import com.github.nscala_time.time.Imports._
 
 import scala.util.Try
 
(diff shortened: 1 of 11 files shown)


app/configuration/Config.scala
```diff
@@ -63,4 +64,9 @@ object Config {
   val sentryDsn = Try(new Dsn(config.getString("sentry.dsn")))
 
   lazy val Salesforce =  SalesforceConfig.from(config.getConfig("touchpoint.backend.environments").getConfig(stage), stage)
+
+  object GoCardless {
+    private val token = config.getString("gocardless.token")
+    val client = GoCardlessClient.create(token, Environment.SANDBOX)
```
afiore: Fine for now to connect to sandbox by default, however it would be good to connect to the right environment here as we initialise a client that might be used later for other operations

app/controllers/Checkout.scala
```diff
@@ -59,15 +59,20 @@ object Checkout extends Controller with LazyLogging {
     )
   }
 
+  def mandatePDF = GoogleAuthenticatedStaffAction.async { implicit request =>
+    SubscriptionsForm.paymentDataForm.bindFromRequest.fold(
+      handleWithBadRequest,
+      paymentData =>
+        goCardlessService.mandatePDFUrl(paymentData).map(url =>
```
afiore: I would call this `generatePDFUrl`

afiore:
:+1: Fine for the backend code

davidrapson:
@ostapneko JS looked good too. Apologies, I'd refactored this code a lot in the meantime though so it moved around quite a bit.

I've fixed the merged conflicts and updated the styling.

screen shot 2015-07-21 at 17 36 51

I wasn't able to see the PDF though as I don't have the correct credentials (guessing there's a config that I'll need). It would be great to have that as I'd like to add some loading/error states to the mandate button. Nothing that should stop this going out though :+1:

ostapneko:
@davidrapson We updated both the production and the development configuration files, so you should be able to pull it from S3

davidrapson:
@ostapneko Great, ta.

TesterSpike:
:+1:

PR #140, opened 2015-07-22
README updates

davidrapson:


ostapneko:
It still works to use 1. for each point I think, and it's more extensible as well should you need to insert a point in the middle.

Otherwise I agree that the AWS should be swapped!

davidrapson:
@ostapneko Good point about the markdown, will swap that back.

PR #141, opened 2015-07-23
Get Prout to run acceptance tests after build is seen as deployed

rtyley:
This follows on from https://github.com/guardian/subscriptions-frontend/pull/134

Incidentally, @afiore mentioned that rather than specifying the full travis script (which includes the https://github.com/guardian/gu-who/commit/83d29a150 hack for better build times) we could maybe update the default `.travis.yml` to use an environment variable to identify the sbt task (`fast-test` or `acceptance-test`) which would result in less duplication. That would be nice - though getting Travis to understand by default that certain files shouldn't trigger a repack of the cached dependencies would be even better.

### DIFF
--- a/.prout.json
+++ b/.prout.json
@@ -1,5 +1,15 @@
 {
-  "checkpoints": {
-    "PROD": { "url": "https://subscribe.theguardian.com/", "overdue": "14M" }
-  }
+    "checkpoints": {
+        "PROD": {
+            "url": "https://subscribe.theguardian.com/",
+            "overdue": "14M",
+            "afterSeen": {
+                "travis": {
+                    "config": {
+                        "script": "sbt ++$TRAVIS_SCALA_VERSION acceptance-test && find $HOME/.sbt -name \"*.lock\" -type f -delete && find $HOME/.ivy2/cache -name \"*[\[\]\(\)]*.properties\" -type f -delete"
+                    }
+                }
+            }
+        }
+    }
 }


rtyley:
> getting Travis to understand by default that certain files shouldn't trigger a repack of the cached dependencies would be even better.

Hey, this commit might be relevant to that!

https://github.com/travis-ci/travis-build/commit/d2a1f8fd626b83fbb3f3ee15f5f5da559c30c53c

http://docs.travis-ci.com/user/caching/#before_cache-phase

afiore:
@rtyley Couldn't we do both? 

We could move the `$HOME/.sbt` and the `$HOME/.ivy2/cache` cleanup snippets into the `before_cache` section of `travis.yml`. At the same time, we can also update `travis.yml` `script` section with something like:

```bash
test_task="${SBT_TEST_TASK:-"fast-test"}"
sbt ++$TRAVIS_SCALA_VERSION $test_task
```

At this point, I assume that Prout could override the default `travis.yml` with something like 

```json
"config": {
   "env": [
      {"SBT_TEST_TASK": "test"}
   ]
}
```

Probably that is enough scope for another PR, but still I think it would be great to avoid duplicating travis built script in two different places.

rtyley:
> I think it would be great to avoid duplicating travis built script in two different places.

Well, the repeated section just becomes: `sbt ++$TRAVIS_SCALA_VERSION fast-test` - which personally I can live with repeating!

afiore:
All right then :+1:

PR #145, opened 2015-07-24
Use gu.com as email address domain

afiore:
This should allow Identity to cleanup test records from their db and remove them from their reporting

@rtyley


rtyley:
:+1: & gonna merge...

PR #150, opened 2015-07-24
Review BUT DON'T MERGE: Send data extension row to ExactTarget upon subscription

ostapneko:
- [x] Update production configuration with exact target configuration
- [ ] Update membership common once it is merged
- [x] Review the business logic with Renaldo?

Upon subscription, this trigger the creation of an ExactTarget Data Extension row, that should in turn send a confirmation email.

This requires some extra config:

```
exact-target {
  client-id=xxx
  client-secret=xxx
  data-extension-keys {
    thank-you=xxx
  }
}
```

Limitation: The next payment date is not yet calculated, and set to a date set absurdly far in the future.

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -18,8 +18,7 @@ import scala.concurrent.Future
 
 object Checkout extends Controller with LazyLogging {
   private val zuoraService = TouchpointBackend.Normal.zuoraService
-  private val exactTargetService = ExactTargetService
-  private lazy val checkoutService = new CheckoutService(IdentityService, SalesforceService, zuoraService, exactTargetService)
+  private lazy val checkoutService = new CheckoutService(IdentityService, SalesforceService, zuoraService, ExactTargetService)
   private val goCardlessService = GoCardlessService
 
   def identityCookieOpt(implicit request: Request[_]): Option[Cookie] =
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -5,7 +5,7 @@ import com.gu.identity.play.IdUser
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config.Identity.webAppProfileUrl
 import forms.{FinishAccountForm, SubscriptionsForm}
-import model.{GuestAccountData, SubscriptionData}
+import model.SubscriptionData
 import play.api.data.Form
 import play.api.libs.json._
 import play.api.mvc._
@@ -18,7 +18,8 @@ import scala.concurrent.Future
 
 object Checkout extends Controller with LazyLogging {
   private val zuoraService = TouchpointBackend.Normal.zuoraService
-  private lazy val checkoutService = new CheckoutService(IdentityService, SalesforceService, zuoraService)
+  private val exactTargetService = ExactTargetService
+  private lazy val checkoutService = new CheckoutService(IdentityService, SalesforceService, zuoraService, exactTargetService)
   private val goCardlessService = GoCardlessService
 
   def identityCookieOpt(implicit request: Request[_]): Option[Cookie] =
@@ -33,7 +34,6 @@ object Checkout extends Controller with LazyLogging {
         SubscriptionsForm()
       }
       //TODO when implementing test-users this requires updating to supply data to correct location
-      val touchpointBackend = TouchpointBackend.Normal
       Ok(views.html.checkout.payment(form, userIsSignedIn = idUserOpt.isDefined, zuoraService.products))
     }
   }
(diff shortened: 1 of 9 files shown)


app/controllers/Checkout.scala
```diff
@@ -18,7 +18,8 @@ import scala.concurrent.Future
 
 object Checkout extends Controller with LazyLogging {
   private val zuoraService = TouchpointBackend.Normal.zuoraService
-  private lazy val checkoutService = new CheckoutService(IdentityService, SalesforceService, zuoraService)
+  private val exactTargetService = ExactTargetService
```
rtyley: Can probably inline this val?

app/model/exactTarget/SubscriptionDataExtensionRow.scala
```diff
@@ -0,0 +1,76 @@
...
+      case '1' => "st"
+      case '2' => "nd"
+      case '3' => "rd"
+      case _   => "th"
+    }
+    val month = dateTime.monthOfYear.getAsText
+
+    val year = dateTime.year.getAsString
+
+    s"$day$daySuffix $month $year"
```
rtyley: This could probably done with a bit less code by using `DateTimeFormat`?

http://joda-time.sourceforge.net/apidocs/org/joda/time/format/DateTimeFormat.html

ostapneko: My main problem was to ordinalize (like 29th). Any idea how to achieve it?

rtyley: The [Guardian style guide](http://www.theguardian.com/guardian-observer-style-guide-d) says:

```
dates
Our style is: 21 July 2014 (day month year; no commas).
```

...so you probably don't need to ordinalize.

app/model/exactTarget/package.scala
```diff
@@ -0,0 +1,29 @@
+package model
+
+package object exactTarget {
+  trait DataExtensionColumn
+
+  case object `SubscriberKey` extends DataExtensionColumn
```
rtyley: I think I'd have been okay with leaving these as simple string constants - you only have to change them back into strings again at https://github.com/guardian/subscriptions-frontend/blob/91699a3/app/services/ExactTargetService.scala#L20

It's kind of nice to have strongly-typed keys, but then relying on the `toString` method of case objects to do the right thing is a bit stressful.

ostapneko: True, wanted to do more at first, but turned out to be useless

app/services/ZuoraService.scala
```diff
@@ -17,10 +17,33 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
+case class SimpleFilter(key: String, value: String) extends ZuoraFilter {
+  override def toFilterString = s"$key='$value'"
+}
+
+case class AndFilter(clauses: (String, String)*) extends ZuoraFilter {
+  override def toFilterString =
+    clauses.map(p => SimpleFilter.tupled(p).toFilterString).mkString(" AND ")
+}
+
+case class OrFilter(clauses: (String, String)*) extends ZuoraFilter {
```
rtyley: Was just wondering if maybe this should just take a `Seq`, as it mainly seems to be called with Seq and ':_*'?

build.sbt
```diff
@@ -23,12 +23,26 @@ lazy val root = (project in file(".")).enablePlugins(
 
 scalaVersion := "2.11.6"
 
+val cxfV = "2.7.5"
+val cxfUtilsV = "2.7.0"
+val exactTargetDeps = Seq(
+  "com.exacttarget" % "fuelsdk" % "1.0.3" from "https://github.com/salesforcefuel/FuelSDK-Java/releases/download/v1.0.3/fuel-java-1.0.3.jar",
```
rtyley: They don't even upload to Maven Central?! Gah! Come on! https://github.com/salesforcefuel/FuelSDK-Java/issues/13

app/services/ZuoraService.scala
ostapneko: It's best to use the same interface for `AndFilter` and `OrFilter`, and in the first case we splat. I find it a bit nicer, and I assume the general case to be splats. But if you prefer I can change that to `Seq`

build.sbt
```diff
@@ -23,12 +23,26 @@ lazy val root = (project in file(".")).enablePlugins(
...
+  "commons-beanutils" % "commons-beanutils" % "1.9.2",
+  "log4j" % "log4j" % "1.2.17",
+  "com.google.code.gson" % "gson" % "2.3.1",
+  "org.apache.cxf" % "cxf-bundle-minimal" % cxfV,
+  "org.apache.cxf" % "cxf-tools-common" % cxfV,
+  "org.apache.cxf" % "cxf-tools-common" % cxfV,
+  "org.apache.cxf.xjc-utils" % "xjc-utils" % cxfUtilsV,
+  "org.apache.cxf.xjc-utils" % "cxf-xjc-runtime" % cxfUtilsV
```
rtyley: This list of dependencies is pretty long! Do we actually have to declare all of it?!

rtyley:
Replaced by #164

PR #151, opened 2015-07-27

ostapneko:
@davidrapson Should I also do the front end validations or are yo uhappy to do them?

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -58,12 +58,12 @@
                         <div class="form-field js-checkout-first">
                             <label class="label" for="first">First name</label>
                             <input type="text" class="input-text js-input" name="personal.first" id="first" value="@form.data.get("personal.first")">
-                            @fragments.forms.errorMessage("This field is required")
+                            @fragments.forms.errorMessage("")
                         </div>
                         <div class="form-field js-checkout-last">
                             <label class="label" for="last">Last name</label>
                             <input type="text" class="input-text js-input" name="personal.last" id="last" value="@form.data.get("personal.last")">
-                            @fragments.forms.errorMessage("This field is required")
+                            @fragments.forms.errorMessage("")
                         </div>
                         <div class="form-field js-checkout-email">
                             <label class="label" for="email">Email address</label>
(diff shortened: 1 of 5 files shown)


davidrapson:
@ostapneko If you are OK to update the validations on the client-side too that would be great.

ostapneko:
@davidrapson Will do

ostapneko:
@davidrapson I did my best on the front end side to make it feature complete without duplication, but I'll leave the polishing to you I guess as I comply to the JS coding standards.

assets/javascripts/modules/checkout/personalDetails.js
```diff
@@ -13,25 +15,65 @@ define([
...
+        return {
+            el: containerEl,
+            maxLength: maxLength,
+            input: $('.js-input', containerEl)[0],
+            error: $('.js-error-message', containerEl)[0]
+        };
+    }
+
+    var firstNameField = new TextField(formEls.$FIRST_NAME_CONTAINER, MAX_NAME_LENGTH);
+    var lastNameField = new TextField(formEls.$LAST_NAME_CONTAINER, MAX_NAME_LENGTH);
```
davidrapson: You don't need the `new` here.

assets/javascripts/modules/checkout/personalDetails.js
```diff
@@ -13,25 +15,65 @@ define([
     var FIELDSET_COMPLETE = 'data-fieldset-complete';
     var HIDDEN_CLASS = 'is-hidden';
 
+    var MAX_NAME_LENGTH = 50;
+    var MAX_EMAIL_LENGTH = 240;
+    var MAX_ADDRESS_LENGTH = 255;
+
+    var ERROR_MESSAGES = {
+        FIELD_TOO_LONG: 'This field is too long',
```
davidrapson: We should tell the user how long is too long

app/views/checkout/payment.scala.html
```diff
@@ -83,23 +83,23 @@
...
                                 <label class="label" for="address-line-1">Address line 1</label>
                                 <input type="text" class="input-text js-input" id="address-line-1" name="personal.address.address1" value="@form.data.get("personal.address.address1")">
-                                @fragments.forms.errorMessage("This field is required")
+                                @fragments.forms.errorMessage("")
                             </div>
                             <div class="form-field js-checkout-street">
                                 <label class="label optional-marker" for="address-line-2">Address line 2</label>
                                 <input type="text" class="input-text js-input" id="address-line-2" name="personal.address.address2" value="@form.data.get("personal.address.address2")">
-                                @fragments.forms.errorMessage("This field is required")
+                                @fragments.forms.errorMessage("")
```
davidrapson: Is it worth still providing a default message, even if we override in most cases?

davidrapson:
:+1: LGTM aside from a couple of comments.

TesterSpike:
The error messages are not removed when the user replaces the overly long text with appropriately length text.  
I'll add my vote to improving the text of the error messages so that the user knows what the limits are.

davidrapson:
As @TesterSpike says, on second look this doesn't actually validate fields correctly. The values are not live updated. Going to close this one and put up a new PR to resolve this.

PR #152, opened 2015-07-29
Remove config that does not vary by environment

rtyley:
Because we can deploy as easily as we can change config, there's no point in putting values that don't vary by environment into config files - best to just hard-code an inline constant.

cc @afiore


rtyley:
Successfully created a subscription in UAT with this updated configuration: https://apisandbox.zuora.com/apps/Subscription.do?method=view&id=2c92c0f94ed8d0d7014ed991a5721cd2

PR #153, opened 2015-07-29
Switch PROD to use the `UAT` configuration stanzas

rtyley:
This is because PROD currently /is/ pointing to Touchpoint UAT, so it's better to be honest - and it allows me to update the PROD config with actual PROD values without breaking the active site.

cc @ostapneko

### DIFF
--- a/conf/PROD.conf
+++ b/conf/PROD.conf
@@ -1,4 +1,5 @@
 include "touchpoint.PROD.conf"
+include "touchpoint.UAT.conf"
 include "application"
 
 stage="PROD"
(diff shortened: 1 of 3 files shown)


ostapneko:
:+1:

conf/touchpoint.PROD.conf
```diff
@@ -1,26 +1,26 @@
...
+            }
+        }
+        zuora {
+            paymentDelayInDays = 16
+            api {
+                url = "https://api.zuora.com/apps/services/a/58.0"
+                username = ""
+                password = ""
+            }
+            digital = "???"
```
rtyley: This product hasn't been set up yet...

PR #156, opened 2015-07-30
Update terms in review details section

davidrapson:
screen shot 2015-07-30 at 13 59 00

These are the changes requested but I'm going to try and push to reduce this or display it in a different way, as in my opinion this is too much information to show.

@rtyley


TesterSpike:
:+1:

PR #157, opened 2015-07-30
Allow analytics to be disabled

davidrapson:
Models all analytics in JS, rather than html includes/script tags.
Allows analytics to be disabled, will only run if:

- `guardian.analyticsEnabled` is `true`
- `Navigator.doNotTrack` is not set
- We don't have an `ANALYTICS_OFF` cookie for test users

If the CSP test is successful on Membership we'll also be implementing this here, which will help to track the impact of these analytics.

@rtyley


assets/javascripts/modules/analytics/affectv.js
```diff
@@ -0,0 +1,28 @@
+/*global Raven */
```
rtyley: 'Raven'?

assets/javascripts/modules/analytics/krux.js
```diff
@@ -0,0 +1,16 @@
+/*global Raven */
```
rtyley: I'm guessing this C&P?

assets/javascripts/modules/analytics/affectv.js
davidrapson: It's the Sentry client; It's exposed as a global.

assets/javascripts/modules/analytics/krux.js
davidrapson: As above

assets/javascripts/modules/analytics/affectv.js
rtyley: Oh, fair enough.

rtyley:
:+1:

PR #158, opened 2015-07-30
SVG Icons

davidrapson:
This PR adds the ability to use SVGs for icons, using the same technique as Membership.  Only updates a couple of icons for now:

screen shot 2015-07-30 at 15 18 55
screen shot 2015-07-30 at 15 31 49


TesterSpike:
:+1:

PR #160, opened 2015-07-31
IE cross-browser fixes

davidrapson:
Some cross-browser fixes to better support IE on checkout pages.

- Uses an alternate `Promise` polyfill; and actually loads it…
I'm not a huge fan of using an internal guardian fork that isn't even used on frontend anymore. Also we actually never loaded the polyfill. The fact we missed this for so long shows how good native `Promise` support is now…maybe.

- ~~Fixes some dodgy form markup. Double `<form>`!!
I can only assume this was a merge conflict gone awry. IE 10 has handy markup validation reporting in the console which is how I spotted it. Otherwise it seems most browsers just don't care. Resilience!~~ Moved to https://github.com/guardian/subscriptions-frontend/pull/161

None of this is even slightly embarrassing; nope.

@TesterSpike @rtyley

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -146,7 +146,7 @@ module.exports = function(grunt) {
                     baseUrl: '<%= dirs.assets.javascripts %>',
                     paths: {
                         '$': 'utils/$',
-                        'Promise': 'bower_components/native-promise-only/lib/npo.src',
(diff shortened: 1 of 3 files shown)


TesterSpike:
:+1:

PR #162, opened 2015-08-04

davidrapson:
Re-opened version https://github.com/guardian/subscriptions-frontend/pull/151 with temporary fix for length validations not being live updated.

As a general comment, I regret not pushing harder for a simpler approach to form validation. I've added a card to the backlog to use an approach more similar to Membership. There are solid HTML5 validation attributes we can use an enhance on top of, rather than re-implementing things. Not to diminish @ostapneko's work which was completely appropriate in the context.


davidrapson:
@TesterSpike Look OK to you?

PR #163, opened 2015-08-04

davidrapson:
This is a version of https://github.com/guardian/subscriptions-frontend/pull/162 with the JS validation removed, only including the basic server validation from @ostapneko .

@TesterSpike


TesterSpike:
:+1:   If the fields are too long the user is returned to the Your Details page.  No system error is displayed.

PR #167, opened 2015-08-07
Adapt checkout acceptance test to new markup structure

afiore:
The Accept GNM marketing button has been recently moved from the personal details to the confirm section of the form, thus breaking the checkout acceptance test. This updates the test so that it works with the new markup structure.

@tudorraul


TesterSpike:
:+1:

PR #170, opened 2015-08-07
Include marketing preferences to salesforce API call

afiore:
Fixes an [issue](https://trello.com/c/RANmZpQ5/108-email-preferences-are-not-sending-through-to-salesforce) reported on UAT whereby end-user email preferences are not being sent to Salesforce.

**TODO:**

- [x] Get access to Salesforce
- [x] Verify on Salesforce that the fix is effective 

@rtyley @tudorraul

### DIFF
--- a/app/services/SalesforceService.scala
+++ b/app/services/SalesforceService.scala
@@ -24,9 +24,7 @@ trait SalesforceService extends LazyLogging {
       Keys.MAILING_CITY -> personalData.address.town,
       Keys.MAILING_POSTCODE -> personalData.address.postcode,
       Keys.MAILING_COUNTRY -> "United Kingdom",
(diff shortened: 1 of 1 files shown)


afiore:
Verified on Salesforce DEV and this is working as expected. In the example below, I have tried to set the marketing options to false (which is not the default on Salesforce) and the value is propagating correctly:

screen shot 2015-08-10 at 14 58 13

davidrapson:
Retroactive :+1: …

afiore:
Ops... sorry, I have accidentally merged this without being reviewed. I have not blocked the deploy and reverted it as it is fairly small and safe change

davidrapson:
@afiore As discussed, in this case this change is pretty safe so we can let it go through but if otherwise just revert the change and merge the revert. You can always cancel a deploy in RiffRaff too as well if needed.

PR #171, opened 2015-08-10
fix log file CW read path

tudorraul:
Cloudwatch agent is not reading from the correct path.


afiore:
:+1: This should fix it. I guess the problem before was that the log file name didn't match the one defined in `conf/logback.xml`

tudorraul:
possibly so

PR #172, opened 2015-08-10
Terms copy changes & remove mandate PDF

davidrapson:
This PR

- Updates terms copy on payment page
- Removes mandate generation as it is no longer required (sorry @ostapneko)

screen shot 2015-08-10 at 12 35 00

@afiore @rtyley


afiore:
:+1: One feature less!

TesterSpike:
:+1:

PR #174, opened 2015-08-10
Update global error handler to catch Payment errors

afiore:
Catch Zuora errors with code [TRANSACTION_FAILED](https://knowledgecenter.zuora.com/BC_Developers/SOAP_API/L_Error_Handling/Errors) and display a payment error message to the user (arguably a bit more informative than the standard 500 error page).

screen shot 2015-08-10 at 14 09 54

**Note on implementation**

It would be arguably better handling this error explicitly as a value by modelling it into the return type of `CheckoutService.processSubscription`. This would involve changing the definition of `CheckoutResult` into something like:

```scala
trait CheckoutResult

case class CheckoutOk(
    salesforceMember: MemberId,
    userIdData: UserIdData,
    zuoraResult: SubscribeResult) extends CheckoutResult

case object PaymentError extends CheckoutResult
...
```

Happy to change the implementation and doing this, but maybe a bit out of scope given that this ticket is intended as a quick fix.

@tudorraul

### DIFF
--- a/app/monitoring/ErrorHandler.scala
+++ b/app/monitoring/ErrorHandler.scala
@@ -30,17 +30,15 @@ class ErrorHandler @Inject() (env: Environment,
   }
 
   override protected def onDevServerError(request: RequestHeader, exception: UsefulException): Future[Result] = {
-    val default: Handler = { case e =>
+    specialHandler.applyOrElse(exception.cause, { case e =>
       super.onDevServerError(request, exception)
-    }
-    specialHandler.applyOrElse(exception.cause, default)
+    }: Handler)
   }
 
   override protected def onProdServerError(request: RequestHeader, exception: UsefulException): Future[Result] = {
-    val default: Handler = { case _ =>
+    specialHandler.applyOrElse(exception.cause, { case _ =>
       Future.successful(NoCache(InternalServerError(views.html.error500(exception))))
-    }
-    specialHandler.applyOrElse(exception.cause, default)
+    }: Handler)
   }
 
   private val specialHandler: Handler = {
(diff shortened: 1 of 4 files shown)


app/views/zuoraTransactionFailed.scala.html
```diff
@@ -0,0 +1,9 @@
+@main("Payment failed") {
+    <main class="page-container gs-container">
+        @fragments.page.header("Payment error", None)
+        <p>
+            Sorry, we were unable to process your payment. Please check your details and try again.
+            If this error persists, please contact support on <a href="mailto:subscriptions.dev@@theguardian.com">subscriptions.dev@@theguardian.com</a>.
```
davidrapson: "support on" => "support at" maybe?

afiore: Well spotted, need to change that also on the main 500 page

app/monitoring/ErrorHandler.scala
```diff
@@ -27,6 +29,22 @@ class ErrorHandler @Inject() (
...
-    Future.successful(NoCache(InternalServerError(views.html.error500(exception))))
+  override protected def onDevServerError(request: RequestHeader, exception: UsefulException): Future[Result] = {
+    val default: Handler = { case e =>
+      super.onDevServerError(request, exception)
+    }
+    specialHandler.applyOrElse(exception.cause, default)
+  }
+
+  override protected def onProdServerError(request: RequestHeader, exception: UsefulException): Future[Result] = {
+    val default: Handler = { case _ =>
```
tudorraul: can this be inlined?

tudorraul:
:+1: bar the comment

app/monitoring/ErrorHandler.scala
afiore: Is not going to look much prettier, but why not...

PR #175, opened 2015-08-10
Improve sort-code UX

davidrapson:
- Merge three sort-code fields into one
- Add input masking component
- Add placeholders & HTML5 validation attributes

sort-code mov

@afiore @tudorraul

### DIFF
--- a/app/model/SubscriptionData.scala
+++ b/app/model/SubscriptionData.scala
@@ -3,9 +3,7 @@ package model
 import com.gu.identity.play.IdUser
 
 case class PaymentData(account: String, sortCodeValue: String, holder: String) {
-  val sortCode = sortCodeValue
-    .replaceAll(" ", "")
-    .trim.replaceAll("-", "")
+  val sortCode = sortCodeValue.filter(_.isDigit)
 }
 
 case class AddressData(address1: String, address2: String, town: String, postcode: String) {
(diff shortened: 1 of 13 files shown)


TesterSpike:
Tested in Firefox/Chrome/IE11/Safari and the field works as expected except for Firefox which shows a red border even if everything is correct

davidrapson:
@TesterSpike Latest Firefox i'm guessing?

TesterSpike:
Yes - I haven't tried any of the older ones (except for IE where I tested IE9 as well)

davidrapson:
@TesterSpike Thanks, I'll take a look.

davidrapson:
@TesterSpike Found the issue with Firefox, mind checking again?

app/model/SubscriptionData.scala
```diff
@@ -2,8 +2,10 @@ package model
 
 import com.gu.identity.play.IdUser
 
-case class PaymentData(account: String, sortCode1: String, sortCode2: String, sortCode3: String, holder: String) {
-  val sortCode = s"$sortCode1$sortCode2$sortCode3"
+case class PaymentData(account: String, sortCodeValue: String, holder: String) {
+  val sortCode = sortCodeValue
```
afiore: Assuming what you want to do is to strip everything but the sort code digits, a slightly better way of writing this would be:

```scala
sortCode.filter(_.isDigit)
```

afiore:
:+1: Looking good!

TesterSpike:
:+1:

PR #176, opened 2015-08-10
Prevent accidental checkout form submits

tudorraul:


tudorraul:
cc @davidrapson

app/views/checkout/payment.scala.html
```diff
@@ -34,6 +34,7 @@
             <form class="form js-checkout-form" method="POST">
 
                 @* ===== Your details ===== *@
+                <a name="yourDetails"></a>
```
davidrapson: Could just give the `<fieldset>` and `id`? Same effect, but one-less element. Whilst I think about it, we should also do this for the **edit** links--same process but in reverse, and `<a`> pointing to the *previous* container.

davidrapson:
Noted that we should do the same for **edit** links, but :+1:

PR #178, opened 2015-08-11
README

vivekkr:
Changes to README.md file for setting up a new install for the subscription

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,20 @@
 # Subscriptions frontend
 
+### NGinx
+
+   To run standalone you can use the default nginx installation as follows:
+
+   Install nginx:
+
+   Mac OSX: `brew install nginx`
+
+   Make sure you have a sites-enabled folder under your nginx home. This should be
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,7 +38,7 @@
 1. Download our private keys from the `subscriptions-private` S3 bucket. If you have the AWS CLI set up you can run:
 
     ```
-    aws s3 cp s3://subscriptions-private/DEV/subscriptions-frontend.conf /etc/gu
+    aws s3 cp s3://subscriptions-private/DEV/subscriptions-frontend.conf /etc/gu  --profile membership
```
tudorraul: This command is dependent on the configuration of `~/.aws/credentials`, so setting the profile is optional. Please indicate this in the readme.

tudorraul:
:+1: bar the comment

README.md
vivekkr: Valid comment. If we had a default profile then we don't need to supply the profile explicitly as 'membership'. Currently the credentials file has [membership] so using that

PR #183, opened 2015-08-13
Max lengths

davidrapson:
This PR resolves the length constraints issue on post-codes by generally viewing the form attributes:

- Add HTML5 validation attributes to all fields
- Add accessible labels to edit links
- Pass signed in state *into* validation function

@TesterSpike @afiore


afiore:
:+1: Looks good

TesterSpike:
:+1:  :-1:  the validation is working as expected but I cannot submit the final details successfully.  I believe it's my environment configuration so that is the reason for the ambivalence

PR #184, opened 2015-08-14
Adding better logging to exact target and zuora calls

dominickendrick:
This PR adds logging to the futures that look up information from Zuora for sending emails via exact target. 
It only logs in failure situations, so should help for debugging.

### DIFF
--- a/app/services/ExactTargetService.scala
+++ b/app/services/ExactTargetService.scala
@@ -18,18 +18,27 @@ import scala.concurrent.duration._
 trait ExactTargetService extends LazyLogging {
   def etClient: ETClient
 
+  implicit class FutureWithLogging[T](f: Future[T]) {
+    def withLogging(label: String): Future[T] = {
+      f.onFailure {
+        case e: Throwable => logger.error(s"Future ${label} failed with error : ${e}", e)
+      }
+      f
+    }
+  }
+
   def sendETDataExtensionRow(subscribeResult: SubscribeResult, subscriptionData: SubscriptionData, zs: ZuoraService): Future[Unit] = {
-    val subscription = zs.subscriptionByName(subscribeResult.name)
+    val subscription = zs.subscriptionByName(subscribeResult.name).withLogging("subscriptionByName")
 
     val accAndPaymentMethod = for {
       subs <- subscription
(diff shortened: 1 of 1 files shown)


dominickendrick:
@tudorraul

app/services/ExactTargetService.scala
```diff
@@ -18,18 +18,27 @@ import scala.concurrent.duration._
 trait ExactTargetService extends LazyLogging {
   def etClient: ETClient
 
+  implicit class FutureWithLogging[T](f: Future[T]) {
```
tudorraul: I would move this into `services` package

app/services/ExactTargetService.scala
```diff
@@ -18,18 +18,27 @@ import scala.concurrent.duration._
...
+      f.onFailure {
+        case e: Throwable => logger.error(s"Future ${label} failed with error : ${e}", e)
+      }
+      f
+    }
+  }
+
   def sendETDataExtensionRow(subscribeResult: SubscribeResult, subscriptionData: SubscriptionData, zs: ZuoraService): Future[Unit] = {
-    val subscription = zs.subscriptionByName(subscribeResult.name)
+    val subscription = zs.subscriptionByName(subscribeResult.name).withLogging("subscriptionByName")
```
tudorraul: I would delegate the error reporting to corresponding services

ostapneko:
I tend to agree with @tudorraul: I don't think it's the role of a Future to give business-specific error context. Rather, the underlying service should be more telling if needed. That provides more flexibility when we use the service: we should not lose the error information if we want to use the service in a different way.

davidrapson:
@ostapneko @tudorraul Anything happening with this?

tudorraul:
This PR emerged out of an investigation that revealed some blind spots in our logging, I think they should be tackled, but I also stand by what I commented above. We can consider reassigning this, maybe @joelochlann can look at it?

davidrapson:
In absence of Raul, perhaps someone else should pick this up? @joelochlann Maybe @dominickendrick can walk you through this?

ostapneko:
Do we really need the PR anymore? It was opened in the context of an investigation. I would be for closing it.

PR #186, opened 2015-08-14
Render payment form with errors

davidrapson:
It was bugging me that we don't handle form errors server-side. This PR updates the checkout form to render-with errors if there is an issue.

**Warning** I've just Play documentation copy-pasted and IntelliJ autocompleted my way to a solution for this. I'm sure there is a more idiomatic way to do this; but it was a useful learning experiment trying this.

I'm also aware that Play has a bunch of form helpers so I'm probably repeating some things, but as we're not using them anywhere it felt OK to follow suit and keep the error class handling manual.

screen shot 2015-08-14 at 14 43 26

@afiore

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -1,48 +1,42 @@
 package controllers
 
 import actions.CommonActions._
-import com.gu.identity.play.IdUser
+import com.gu.identity.play.{AuthenticatedIdUser, IdUser}
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config.Identity.webAppProfileUrl
 import forms.{FinishAccountForm, SubscriptionsForm}
 import model.SubscriptionData
-import model.zuora.BillingFrequency
 import play.api.data.Form
 import play.api.libs.json._
-import play.api.mvc
 import play.api.mvc._
+import services.AuthenticationService.authenticatedUserFor
 import services.CheckoutService.CheckoutResult
 import services._
-import utils.TestUsers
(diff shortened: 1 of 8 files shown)


app/controllers/Checkout.scala
```diff
@@ -1,46 +1,41 @@
...
 import utils.TestUsers.{NameEnteredInForm, PreSigninTestCookie}
 import views.html.{checkout => view}
-
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
-import AuthenticationService.authenticatedUserFor
 
 object Checkout extends Controller with LazyLogging {
 
+  def fillForm(form: Form[SubscriptionData], authUserOpt: Option[AuthenticatedIdUser]): Future[Form[SubscriptionData]] = {
```
afiore: I agree with moving this one out of `renderCheckout`, but I would also mark it as private

afiore:
:+1: LGTM

app/views/fragments/checkout/fieldsPayment.scala.html
```diff
@@ -1,11 +1,15 @@
 @(form: Form[model.SubscriptionData])
 
-<div class="form-field js-checkout-account">
+@hasError(key: String) = @{
```
afiore: If you find yourself repeating this kind of code too much, remember that you can always create helper methods under the `views` package, and they will be available in the view without the need of explicitly importing them

davidrapson:
@TesterSpike Look OK to you?

PR #189, opened 2015-08-14
Omniture tracking

tudorraul:
This adds Omniture tracking to the checkout process, as follows:
- checkout form transitioning between personal details, payment details and form review steps
- checkout form submission
- thank you page


assets/javascripts/main.js
```diff
@@ -2,7 +2,7 @@ require([
     'utils/ajax',
     'modules/analytics/setup',
     'modules/toggle',
-    'modules/optionMirror',
+    'modules/checkout/optionMirror',
```
davidrapson: How come this was moved?

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -20,6 +20,8 @@ define(function () {
         s.prop47 = 'UK';
         s_code = s.t();
 
+        checkoutOmniture.init(s);
```
davidrapson: Is there a way to write this so the checkout code is initialised elsewhere? Or does is need to be done here for Omniture?

assets/javascripts/modules/checkout/omniture.js
```diff
@@ -0,0 +1,48 @@
+define(['$'], function ($) {
+    'use strict';
+
+    var omniture = {};
+
+    function getSubscriptionInfo(){
+        var selectedFrequency = $('.js-option-mirror-group input:checked');
```
davidrapson: I'd recommend using a different `.js-` class to avoid coupling this to the `optionMirror` component.

assets/javascripts/main.js
tudorraul: since it's used only in checkout, I thought it should reside with the rest of checkout related scripts

davidrapson: Sure, but it's generic enough that it doesn't need to.

tudorraul:
cc @davidrapson @afiore

assets/javascripts/modules/checkout/reviewDetails.js
```diff
@@ -47,8 +53,17 @@ define([
         });
     }
 
+    function paymentSubmissionTracking() {
+        if(formEls.$CHECKOUT_SUBMIT.length) {
```
davidrapson: You can probably use `clickHelper` defined above to minimise boilerplate here.

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -33,6 +72,7 @@ define(function () {
     }
 
     return {
-        init: init
+        init: init,
```
davidrapson: Did you not need the caching approach here in the end?

davidrapson:
Couple of minor comments on the JS but otherwise looks good :+1: Probably want to get another +1 from @afiore on the Scala

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -20,6 +57,8 @@ define(function () {
         s.prop47 = 'UK';
         s_code = s.t();
 
+        bindDataTracking();
+
         if (s_code) {
             /*jslint evil: true */
             document.write(s_code);
```
davidrapson: I would **love** if we could get rid of the `document.write`

app/controllers/Checkout.scala
```diff
@@ -57,7 +59,16 @@ object Checkout extends Controller with LazyLogging {
 
     touchpointBackendResolution.backend.checkoutService.processSubscription(formData, idUserOpt).map { case CheckoutResult(_, userIdData, subscription) =>
       val passwordForm = userIdData.toGuestAccountForm
-      Ok(view.thankyou(subscription.name, formData.personalData, passwordForm, touchpointBackendResolution))
+
+      val ratePlan = touchpointBackendResolution.backend.zuoraService.products.filter(_.ratePlanId == formData.ratePlanId).head
+
+      val dataNumberOfMonths = ratePlan.frequency match {
```
afiore: Feels like this logic should be a method. Something like `ratePlan.frequency.numberOfMonths`

afiore:
:+1: bar the comment

app/controllers/Checkout.scala
```diff
@@ -57,7 +59,16 @@ object Checkout extends Controller with LazyLogging {
 
     touchpointBackendResolution.backend.checkoutService.processSubscription(formData, idUserOpt).map { case CheckoutResult(_, userIdData, subscription) =>
       val passwordForm = userIdData.toGuestAccountForm
-      Ok(view.thankyou(subscription.name, formData.personalData, passwordForm, touchpointBackendResolution))
+
+      val ratePlan = touchpointBackendResolution.backend.zuoraService.products.filter(_.ratePlanId == formData.ratePlanId).head
```
afiore: Nearly missed this as the line 63 is so long that it overflows on Github's diff view.

1) are we getting a product or a ratePlan here?... Looks like are getting a Product, but we name the value `ratePlan`
2) The direct call to `.head` is not great as it will produce a very uninformative exception (e.g head called on null object...). If a product without the expected rate plan is definitely a non recoverable circumstance, it still would be better to fail with a slightly more informative message:

```scala
products.filter(_.ratePlanId == formData.ratePlanId).getOrElse(sys.error("Cannot find $"))
```

assets/javascripts/modules/checkout/reviewDetails.js
tudorraul: `clickHelper ` prevents defaults, but we need to let the form to be submitted.

davidrapson: Good point, as you were…

assets/javascripts/modules/analytics/omniture.js
tudorraul: Yes, but I didn't get the chance to push the changes until now, please review

PR #194, opened 2015-08-24
Update final review copy

ostapneko:
This updates the copy of the final checkout review, and adds a links to [Why your data matters to us](https://www.theguardian.com/info/video/2014/nov/03/why-your-data-matters-to-us-video)

@vivekkr


vivekkr:
:+1: Looks Good

PR #196, opened 2015-08-24
Alleviate deployment issues caused by Zuora Authentication race-condition

tudorraul:
https://trello.com/c/hwwXGTFr/125-alleviate-deployment-issues-caused-by-zuora-authentication-race-condition-wait-on-auth-for-salesforce-zuora

This story is about Zuora taking a very long time to reach an authenticated state with product data

### DIFF
--- a/app/controllers/Healthcheck.scala
+++ b/app/controllers/Healthcheck.scala
@@ -9,8 +9,8 @@ object Healthcheck extends Controller with LazyLogging{
 
   def index = Action.async {
       (for {
-       //TODO: Touch salesforce
         products <- TouchpointBackend.Normal.zuoraService.products if products.nonEmpty
+        sfAuth <- TouchpointBackend.Normal.salesforceService.repo.salesforce.getAuthentication
       } yield Ok("OK"))
       .recover { case t: Throwable =>
         logger.error("Health check failed", t)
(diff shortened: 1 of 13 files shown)


app/controllers/Healthcheck.scala
```diff
@@ -1,15 +1,22 @@
...
-object Healthcheck extends Controller {
+object Healthcheck extends Controller with LazyLogging{
 
-  def index = Action {
-    Cached(1) {
-      if (TouchpointBackend.Normal.zuoraService.products.nonEmpty) Ok("OK")
-      else ServiceUnavailable("Service Unavailable")
-    }
+  def index = Action.async {
+      (for {
```
ostapneko: Minor detail, but should we create `isReady` methods in services, and maybe call these methods while looping over those services? It's just that health check might not be the ideal place to determine what it means for each service to be ready

app/controllers/Healthcheck.scala
```diff
@@ -1,15 +1,22 @@
...
+        products <- TouchpointBackend.Normal.zuoraService.products if products.nonEmpty
+        sfAuth <- TouchpointBackend.Normal.salesforceService.repo.salesforce.getAuthentication
+      } yield Ok("OK"))
+      .recover { case t: Throwable =>
+        logger.error("Health check failed", t)
```
ostapneko: Again minor, but we should use WARN log level at most, we don't want to raise alerts for that

ostapneko:
:+1: , minus the very minor comments. Also, it would be nice to unify this initialisation method with the ExactTargetService one.

chrisjowen:
Sucessfully created a UAT user on production 

image

PR #199, opened 2015-08-27
Add security headers as suggested by Pentura

rtyley:
These 3 headers are added by this config:

```
$ curl --silent --insecure -I https://sub.thegulocal.com/ | grep X-
X-Frame-Options: SAME-ORIGIN
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
```

This is how the headers relate to the Penture report:

X-XSS-Protection: Finding 1.1
X-Frame-Options: Finding 1.3
X-Content-Type-Options: Not mentioned by Pentura, but a reasonable addition

https://www.playframework.com/documentation/2.4.x/SecurityHeaders

I chose to just directly configure the SecurityHeadersFilter through a case class, rather than muck around with the config file.

I chose to not enable the `permittedCrossDomainPolicies` and `contentSecurityPolicy` headers because I wasn't sure how those would interact with analytics, tracking etc.

@vivekkr @davidrapson @blackyjnz

### DIFF
--- a/app/Global.scala
+++ b/app/Global.scala
@@ -3,9 +3,18 @@ import play.filters.csrf.CSRFFilter
 import monitoring.SentryLogging
 import play.api.Application
 import play.api.mvc.WithFilters
+import play.filters.headers.{SecurityHeadersConfig, SecurityHeadersFilter}
 import services.TouchpointBackend
 
-object Global extends WithFilters(CheckCacheHeadersFilter, CSRFFilter(), AddEC2InstanceHeader) {
+object Global extends WithFilters(
+  CheckCacheHeadersFilter,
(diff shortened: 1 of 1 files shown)


rtyley:
Verified that the new security headers are in place on PROD:

```
$ curl --silent -I https://subscribe.theguardian.com/ | grep X-
X-Content-Type-Options: nosniff
X-Frame-Options: SAME-ORIGIN
X-XSS-Protection: 1; mode=block
...
```

image

However, did just get a (possibly unrelated) Fastly error while signing up with a test user:

```
Error 503 backend read error

backend read error

Guru Mediation:

Details: cache-lcy1133-LCY 1440755146 361684025

Varnish cache server
```

PR #201, opened 2015-08-27
Add bank details validation

ostapneko:
See ticket:  https://trello.com/c/kIvgLKlv/128-better-user-experience-when-zuora-checkout-transaction-fails

This does an Ajax check between the payment details input and the confirmation. It uses GoCardless API to do the check.

The error message is displayed in relation to the account number field. The error for this field is generic, whether it is due to wrong formatting or invalid account.

image

### DIFF
--- a/app/services/GocardlessService.scala
+++ b/app/services/GocardlessService.scala
@@ -27,6 +27,11 @@ object GoCardlessService extends GoCardlessService with LazyLogging {
         .getUrl
     }
 
+  /**
+   *
+   * @return true if either the bank details are correct, or the rate limit for this enpoint is reached.
+   *         In the latter case an error is logged.
+   */
   override def checkBankDetails(paymentData: PaymentData): Future[Boolean] = {
     val sortCode = paymentData.sortCode.replaceAllLiterally("-", "")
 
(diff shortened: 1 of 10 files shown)


ostapneko:
@vivekkr @afiore @rtyley

app/services/GocardlessService.scala
```diff
@@ -8,9 +10,10 @@ import scala.concurrent.Future
 
 trait GoCardlessService {
   def mandatePDFUrl(paymentData: PaymentData): Future[String]
```
afiore: This method probably should go!

assets/javascripts/modules/checkout/paymentDetails.js
```diff
@@ -37,16 +38,19 @@ define([
...
+        validatePayment({
             accountNumber: formEls.$ACCOUNT.val(),
             accountHolderName: formEls.$HOLDER.val(),
             sortCode: formEls.$SORTCODE.val(),
             detailsConfirmed: formEls.$CONFIRM_PAYMENT[0].checked
+        }).then(function(validity){
+            if (validity.allValid) {
+                nextStep();
+            } else {
+                console.log(validity);
```
afiore: Remove console log

afiore:
:+1: looks fine by me

ostapneko:
image

rtyley:
There seems to be a bit of a problem with the validation: the 'Bank account number' validation always fails if any other element in that section is failing, or so it would seem?

image

*Possibly not directly related, but still nasty:* If I satisfy every requirement in the section, it will advance to the next section. However, if I then click 'edit' on the "2 Payment details" section, every field that was failing before is still red (looks like the validation status was not corrected when the advance from section 2 to 3 took place):

image


cc @ostapneko and @chrisjowen

ostapneko:
This is fixed in:

https://github.com/guardian/subscriptions-frontend/pull/204

PR #205, opened 2015-08-28
Checkout submit error

chrisjowen:
Added DomReady and only initiated the javascript components when the dom has been loaded.


rtyley:
:+1: we should merge this now, @davidrapson can maybe take a look at the fix when he gets back.

chrisjowen:
Checked on prod and now can no longer replicate the error

chrisjowen:
fixed with https://github.com/guardian/subscriptions-frontend/issues/209

PR #207, opened 2015-09-01
ELB for OSSEC --- CF template change request

rtyley:
Jonathan Westlake (@westy4096) provided this change to support INFOSec requirements.

https://trello.com/c/Xg4wF7IO/114-ossec-will-need-elb-monitoring-to-be-logged-to-an-s3-bucket-so-that-monitoring-can-be-enabled-for-infosec

### DIFF
--- a/cloud-formation/subscriptions-app.cf.json
+++ b/cloud-formation/subscriptions-app.cf.json
@@ -195,6 +195,12 @@
                     "Interval" : "10",
                     "Timeout" : "5"
                 },
+                "AccessLog": {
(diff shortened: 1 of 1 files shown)


rtyley:
8328c68 failed when we tried to apply it as cloud formation: 

```
11:07:49 UTC+0100	UPDATE_FAILED	AWS::ElasticLoadBalancing::LoadBalancer	FrontendElasticLoadBalancer	Encountered unsupported property AccessLog
```

PR #208, opened 2015-09-01
Redirect to thank you + remove Members ad

ostapneko:
Relates to tickets

- [131 - Ability to refresh the thank you page + remove user details](https://trello.com/c/ZNgolAft/131-after-a-post-request-upon-refreshing-the-page-the-form-details-are-resent-to-the-server-which-leads-to-a-error)
image

- [150 - Remove Members ad](https://trello.com/c/jjWK3mpf/150-remove-membership-ad-on-order-confirmation-page)

image


app/controllers/Checkout.scala
```diff
@@ -11,17 +11,30 @@ import play.api.libs.json._
...
+  }
+
+  def zuoraService(implicit res: TouchpointBackend.Resolution): ZuoraService =
+    res.backend.zuoraService
+
+  def checkoutService(implicit res: TouchpointBackend.Resolution): CheckoutService =
+    res.backend.checkoutService
 
   def renderCheckout = NoCacheAction.async { implicit request =>
+    implicit val resolution = TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)
```
rtyley: `touchpointBackend` is probably a better name for this field.

app/controllers/Checkout.scala
```diff
@@ -11,17 +11,30 @@ import play.api.libs.json._
...
+import scala.Function.const
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
 object Checkout extends Controller with LazyLogging {
+  object SessionKeys {
+    val subsName = "newSubs_subscriptionName"
+    val ratePlanId = "newSubs_ratePlanId"
+    val userId = "newSubs_userId"
+    val token = "newSubs_token"
```
rtyley: These field names should be capitalised:

    "Constant names should be in upper camel case. That is, if the member is final, immutable
    and it belongs to a package object or an object, it may be considered a constant (similar to
    Java’s static final members):"

http://docs.scala-lang.org/style/naming-conventions.html#values_variable_and_methods



Also, token should be more specifically named: ie 'IdentityGuestPasswordSettingToken'

ostapneko: I thought that this was implied by the `val` keyword... Changing it nonetheless

davidrapson:
:+1: Worth removing the styles and image associated with `membershipPromo` too. Doesn't need to hold up this PR though.

ostapneko:
@davidrapson Removed the membership styling, can you check?

davidrapson:
:gem:

tudorraul:
:+1: :shipit:

ostapneko:
On PROD:

image

PR #209, opened 2015-09-01
Removed domready from the main script required modules. Side effect i…

chrisjowen:
…s that domready is global but main script loadeder for cut the mustard script will only load when dom is ready for it.

Note this is due to the fact that the pull request from https://github.com/guardian/subscriptions-frontend/pull/205 will not fix the issue found in trello https://trello.com/c/jSUOXoK3/152-continue-button-on-checkout-page-does-not-open-payment-details this willl

#QA 
- [ ] Load the checkout page, hitting the continue button should cause a validation error, NOT just anchor you to the link in the page


dominickendrick:
:+1: but we should try an think of a more elegant fix if poss

chrisjowen:
Checked on prod and the error no longer exists

PR #211, opened 2015-09-01
Ab test product flow

chrisjowen:
Uses javascript to push 50% of the traffic to the new checkout page as per https://trello.com/c/zspzOOBU/113-a-b-test-new-release-of-the-product-against-the-previous-flow

For QA please note that once the user hits the country select page the first time a cookie named COUNTRY_FLOW is dropped which will ensure the user always goes to the same URL in the future. Remove this cookie to test the url switches.


TesterSpike:
:+1:

dominickendrick:
hm

PR #213, opened 2015-09-01
Ab test product flow (To fix the revert)

chrisjowen:
Reverts guardian/subscriptions-frontend#212

Pushed AB testing to frontend before it was required, so had to revert This is not the revertion of that revertion i.e. the original commit.

### DIFF
--- a/app/views/digitalpack/info.scala.html
+++ b/app/views/digitalpack/info.scala.html
@@ -10,7 +10,7 @@
         </div>
         <div class="pricing-cta__action">
             <a class="button button--large button--primary"
-               href="https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=@edition.cmp"
+               href="/digital/country"
                data-test-id="digital-pack-@edition.id-trial-header"
             >Start your free trial</a>
         </div>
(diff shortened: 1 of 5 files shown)


assets/javascripts/modules/country/elements.js
```diff
@@ -0,0 +1,7 @@
+define(['$'], function ($) {
+    'use strict';
+
+    return {
+        $CHECKOUT_LINK: $('.js-checkout-link')
```
davidrapson: I'd probably just inline this var in `country.js`

davidrapson:
:+1: aside from minor comment about elements.js. Build is failing, but doesn't look like anything to do with this, so probably just need to restart it.

app/views/digitalpack/info.scala.html
```diff
@@ -10,7 +10,7 @@ <h4 class="pricing-cta__pricing__title">Free for 14 days</h4>
         </div>
         <div class="pricing-cta__action">
             <a class="button button--large button--primary"
-               href="https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=@edition.cmp"
+               href="/digital/country"
```
rtyley: We can do this with routes!

PR #214, opened 2015-09-02
Correct X-Frame-Options value

davidrapson:
Minor correction to https://github.com/guardian/subscriptions-frontend/pull/199, the `X-Frame-Options` header value should be  `SAMEORIGIN` not `SAME-ORIGIN`

@rtyley


rtyley:
Doh! Thank you @davidrapson, the docs agree: https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options :+1:

PR #215, opened 2015-09-02
Fix scaling issues with App badges

davidrapson:
This fixes the scaling issues with app badges on IE. IE has some scaling bugs with SVGs as images, for the sake of these app badges PNGs are fine.

screen shot 2015-09-02 at 14 51 11

This PR also uses the new amazon badge (previously Available on Fire), amazon now have a single app badge which was released August 26th.

@tudorraul


tudorraul:
we don't serve different CSS to IE do we?

davidrapson:
@tudorraul Yep, we do. Anyway, in this case scaling issues applies to *all* IE versions <11

tudorraul:
so can we only make this change for IE browsers and let the other browsers render SVGs?

davidrapson:
@tudorraul I don't think there's any benefit in doing that. There's nothing wrong with these images being PNGs, don't think the complexity of multiple formats is worth it in this case.

tudorraul:
:+1:

PR #217, opened 2015-09-07
Acceptance tests exit code

afiore:
Ensure acceptance tests exit with the correct process exit code. It uses Sbt's  [TestResultLogger](http://www.scala-sbt.org/0.13.7/api/index.html#sbt.TestResultLogger) class to fetch the suite execution outcome and set exit code accordingly. This is required in order for TravisCI to report correctly

@ostapneko @rtyley

### DIFF
--- a/test/acceptance/DatesSpec.scala
+++ b/test/acceptance/DatesSpec.scala
@@ -5,7 +5,7 @@ import utils.Dates
 
 class DatesSpec extends FeatureSpec {
   feature("Date Formatters") {
-    scenario("Ordinal day formatting", Acceptance) {
+    scenario("Ordinal day formatting") {
       assert(Dates.getOrdinalDay(11) == "11th")
       assert(Dates.getOrdinalDay(12) == "12th")
       assert(Dates.getOrdinalDay(13) == "13th")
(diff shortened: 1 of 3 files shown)


test/acceptance/DatesSpec.scala
```diff
@@ -5,7 +5,7 @@ import utils.Dates
 
 class DatesSpec extends FeatureSpec {
   feature("Date Formatters") {
-    scenario("Ordinal day formatting", Acceptance) {
+    scenario("Ordinal day formatting") {
```
afiore: This feels more like a unit test, so I have removed the `Acceptance` tag

davidrapson: Feels like this is in the wrong place then? Should it be in https://github.com/guardian/subscriptions-frontend/tree/master/test/utils

rtyley:
:+1: on the `TestResultLogger` fix, awesome.

I think @davidrapson  is right about the `DatesSpec` test - it should be moved to the regular unit tests folder, but maybe for a separate PR @dominickendrick ?

davidrapson:
Yep don't let me stop this shipping :+1: :shipit:

rtyley:
The TestResultLogger was introduced with https://github.com/sbt/sbt/pull/1225 by the way.

afiore:
happy to move it as part of this PR

PR #218, opened 2015-09-08
Moves feature tests into own package for clarity

TesterSpike:


afiore:
I have already moved the unit test `DateSpec` out of the acceptance package. Honestly I am not sure we would gain much from having a nested `acceptance.features` package

PR #221, opened 2015-09-10
Disable Analytics for Test Users

vivekkr:
Disable Analytics for Test Users by dropping a Cookie ANALYTICS_OFF_KEY. It seems to work only on the Second Request to https://sub.thegulocal.com/test-users# after the PLAY_SESSION Cookie is set

The JS changes were made as part of this commit

https://github.com/guardian/subscriptions-frontend/commit/8a187ad4abb8782fcb251b207f094e4f18cf8616

@davidrapson

### DIFF
--- a/app/controllers/Testing.scala
+++ b/app/controllers/Testing.scala
@@ -2,20 +2,28 @@ package controllers
 
 import actions.CommonActions._
 import com.typesafe.scalalogging.LazyLogging
+import controllers._
 import play.api.mvc.{Controller, Cookie}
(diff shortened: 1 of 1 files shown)


davidrapson:
>  It seems to work only on the Second Request to https://sub.thegulocal.com/test-users# after the PLAY_SESSION Cookie is set

This sounds a bit odd?

davidrapson:
:confused:

vivekkr:
We tried it on @davidrapson Machine and seems to work correctly the first time as well

PR #222, opened 2015-09-10
Template refactoring to remove misplaced DOM elements

tudorraul:
Payment options should not be displayed on the select country page

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -38,12 +38,12 @@
 
             <div class="checkout-container__sidebar">
                 @fragments.checkout.basketPreview(products)
-                <div class="basket-preview__toggle">
+                <div class="basket-options-toggle">
                     <button class="text-button u-button-reset js-toggle" data-toggle="change-payment-frequency">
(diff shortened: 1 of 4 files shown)


app/views/checkout/payment.scala.html
```diff
@@ -20,6 +38,22 @@
 
             <div class="checkout-container__sidebar">
                 @fragments.checkout.basketPreview(products)
+                <div class="basket-preview__toggle">
```
davidrapson: Don't love having BEM fragments `basket-preview__toggle` outside the parent component…

Would rather pass a `showPaymentOptions` boolean to `basketPreview`, or else refactor the markup to make this two separate components.

davidrapson:
:+1: This leaves us with some BEM/markup-fragment weirdness but happy to ship it for now to avoid the issue.

davidrapson:
Interestingly, been through this before it seems… https://github.com/guardian/subscriptions-frontend/commit/98e7926a929c953d161f624e0c298a4f82219556

davidrapson:
:100: thanks for fixing

PR #225, opened 2015-09-15
Simplify JavaScript loading

davidrapson:
This is a slightly simpler approach for the temporary fix in https://github.com/guardian/subscriptions-frontend/pull/209

The reason the original issue happened was because we loaded are main JS in the `<head>`, meaning that the DOM might not be ready in time for our App to have everything it needs.

The solution at the time was to use `domReady`, which does the job but has a couple of side-effects a) required another blocking script call and b) is side-stepping the issue slightly.

Slightly simpler approach is to move the call to `main.js` to the footer (what we do on Membership), by doing this we avoid the issues with DOM readiness.

@chrisjowen @dominickendrick

### DIFF
--- a/app/views/fragments/footerJavaScript.scala.html
+++ b/app/views/fragments/footerJavaScript.scala.html
@@ -1,0 +1,12 @@
+@import controllers.CachedAssets.hashedPathFor
+
+<script id="script-main">
+    /* Only load JS if the browser cuts the mustard */
+    (function(isModern) {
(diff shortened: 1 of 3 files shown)


davidrapson:
Ping @chrisjowen

ostapneko:
:+1:

tudorraul:
:ship:

davidrapson:
200_s

PR #226, opened 2015-09-16
Upgrade zxcvbn to 3.5

davidrapson:
See: https://github.com/guardian/membership-frontend/pull/766

- Updates to explicit version number now that `zxcvbn` has stabilised a bit with version 3.5. No longer using git hash
- Allows us to use zxcvbn as an AMD module rather than having to load as a global.

@tudorraul @afiore

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -55,10 +55,10 @@ module.exports = function(grunt) {
                 expand: true
             },
             zxcvbn: {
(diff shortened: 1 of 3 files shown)


davidrapson:
@dominickendrick @ostapneko Anyone…?

ostapneko:
:+1:

PR #227, opened 2015-09-16
Sbt exitcode redux

afiore:
Looks like the test-suite run [Output](https://github.com/sbt/sbt/blob/1.0.x/main/actions/src/main/scala/sbt/Tests.scala#L31) value parsed by sbt and passed on to `TestResultLogger` is not always correct, and might not match the string-based report produced by the underlying test framework. If Sbt fails to detect an error, we manually parse the string returned by the test framework (scalatest) and use that information to decide whether we should set set an non-zero exit code and mark the build as failed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,7 @@
 import play.sbt.PlayImport
 
+import scala.util.Try
+
 name := "subscriptions-frontend"
 
 version := "1.0-SNAPSHOT"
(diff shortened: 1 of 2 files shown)


afiore:
@rtyley @chrisjowen

afiore:
This is also a long-lived PR. Can someone pick it up? It is intended to make the TravisCI build a bit more reliable

rtyley:
:+1:

afiore:
Verified on travis-ci and it seems to be setting the exit code correctly (failed builds are now red!)
screen shot 2015-10-02 at 17 39 12
screen shot 2015-10-02 at 17 39 24

PR #228, opened 2015-09-17
Add loader to checkout form

davidrapson:
screen shot 2015-09-17 at 11 39 06

Adds processing loader to checkout form. I just want to note that I consider this a ux enhancement only and *not* a way of actually preventing double submission (although it at least masks the problem). Forms can and should work without any javascript so if we have an issue with double submission we should look a preventing this on the backend too. Does at least soften the issue though so we can play the root-cause fix at a later time.

@rtyley @tudorraul


rtyley:
:+1:

PR #232, opened 2015-09-29
Use eslint:recommended set of rules

davidrapson:
See https://github.com/guardian/membership-frontend/pull/785

### DIFF
--- a/assets/javascripts/modules/checkout/validatePayment.js
+++ b/assets/javascripts/modules/checkout/validatePayment.js
@@ -1,3 +1,4 @@
+/* global Raven */
 define([
     'utils/ajax',
     'modules/forms/regex',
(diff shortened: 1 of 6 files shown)


assets/javascripts/modules/checkout/validatePayment.js
```diff
@@ -1,3 +1,4 @@
+/* global Raven */
```
joelochlann: Any reason why this has to be a global rather than passing it as an explicit dependency via require.js?

davidrapson: Raven is an error reporter (it's the client for [Sentry](https://app.getsentry.com/the-guardian/), so is a bit unique in that it *needs* to be present globally as it hooks into `window.onerror`  and various other things and runs outside the application code.

joelochlann:
:+1:

PR #233, opened 2015-09-29
Inline svg sprite in html

davidrapson:
Our svg sprite is small enough that we can just inline it into the HTML rather than loading via JS and then ultimately doing the same thing in a round about way. Warning: rookie scala alert, but based on approach on frontend.

Bonus: Icons render without JS
screen shot 2015-09-29 at 12 56 53

@rtyley @chrisjowen


app/utils/Assets.scala
```diff
@@ -0,0 +1,30 @@
...
+
+object AssetFinder {
+  def apply(assetPath: String): URL = {
+    Option(Play.classloader(Play.current).getResource(assetPath)).getOrElse {
+      throw AssetNotFoundException(assetPath)
+    }
+  }
+}
+
+case class AssetNotFoundException(assetPath: String) extends Exception {
```
afiore: TIP: if you don't want to bother creating an exception class ex-novo, and you cannot find an already existing one which fits the bill, consider using Scala's `sys.error` method, which will throw a generic runtime exception

app/utils/Assets.scala
```diff
@@ -0,0 +1,30 @@
...
+import java.net.URL
+import controllers.CachedAssets.hashedPaths
+import play.api.Play
+import org.apache.commons.io.IOUtils
+
+object Assets {
+
+  def inlineResource(path: String) = {
+    val url = AssetFinder("public/" + hashedPaths.getOrElse(path, path))
+    IOUtils.toString(url)
```
afiore: Since when do we use apache commons? I cannot see it in the build.sbt dependencies...

app/utils/Assets.scala
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  val svgSprite = inlineResource("images/svg-sprite.svg")
+}
+
+object AssetFinder {
+  def apply(assetPath: String): URL = {
```
afiore: I would move this logic into your main Singleton object `Assets, and wrap it into a method called something like `Assets.find`

app/utils/Assets.scala
davidrapson: Is there some magic going on which I'm unaware of? It auto-completed and compiled, so *seemed* to be available? But there were a couple of possibilities  shown for IOUtils

afiore: You should be able to simplify the entire module by using  following snippet 

```scala

import play.api.Play.current
import scala.io.Source

object Asset {
  def inlineResource(path: String): String = 
     current.resource.map(Source.fromURL(_).mkString).getOrElse(
         sys.error(s"Cannot find asset $path. You probably need to run 'grunt compile'."))
}
``` 

References:
[scala.io.Source](http://www.scala-lang.org/api/current/index.html#scala.io.Source$):

davidrapson:
@afiore Mind taking another look?

dominickendrick:
@davidrapson can we clean up the conflicts ?

davidrapson:
@dominickendrick Done

dominickendrick:
:+1:

PR #236, opened 2015-10-01
Login after registration

ostapneko:
Whenever a guest user is converted to a registered user, drop GU cookies so that the newly created user is automatically logged in.

image

@rtyley @dominickendrick


rtyley:
:+1:

ostapneko:
This should be fixed by https://github.com/guardian/subscriptions-frontend/pull/247

@rtyley

PR #237, opened 2015-10-01
Link tracking between new subs site and QSS

chrisjowen:
After investigation there were several problems with out omniture implementation. From the emit point of view we are sending "s.t()" tracking calls to omniture on every page load. This I have determined is a representation (as least in our case) of a page load event. 

Previously when clicking the checkout link we were calling this "s.t()" method twice, once on the click and once again on the page load. What was required was a call to "s.tl()" which means to track a link and to ensure that our page load tracking can contain additional tracking properties if required. 

Update
----------

Unfortunately in order to get accurate link tracking between QSS and the new subs page there were several errors that needed addressing with the current tracking in our checkout page. Namely:

- Country select page needed to broadcast products on load
- Checkout page needed to broadcast products on load (and not after submitting personal details)
- Checkout - Personal details submitted -  We should send the payment details page info, because this represents that the payment details section is open
- Checkout - Payment details submitted - We should sent the confirmation page details
- Checkout - Confirmation details submitted - We should not broadcast anything because this info will be sent on load of the ThankYou page. 
- Thank you - include product information

### DIFF
--- a/assets/javascripts/modules/analytics/omniture.js
+++ b/assets/javascripts/modules/analytics/omniture.js
@@ -85,6 +85,6 @@ define([
 
     return {
         init: init,
-        sendEvent: trackEvent
+        trackEvent: trackEvent
     };
 });
--- a/assets/javascripts/modules/analytics/omniture.js
+++ b/assets/javascripts/modules/analytics/omniture.js
@@ -54,10 +54,10 @@ define([
         s.prop42 = 'Subscriber';
         s.prop47 = 'UK';
 
-        var main = $('main')[0];
+        var main = document.querySelector('main');
         if(main) {
             var attributes = [].slice.call(main.attributes);
-            attributes.map(function(attr){
+            attributes.forEach(function(attr){
                 if(attr.name.indexOf('data-tracking-')>-1){
                     var prop = attr.name.replace('data-tracking-','');
                     s[prop] = attr.nodeValue;
--- a/assets/javascripts/modules/analytics/omniture.js
+++ b/assets/javascripts/modules/analytics/omniture.js
@@ -5,10 +5,25 @@ define([
 ], function ($, bean) {
     'use strict';
 
-    var dataTrackingClickableElements = ['a', 'button', 'input'],
-        omniture, s;
+    var omniture, s;
 
-    function sendEvent(prop17, pageName, products) {
+
+    function bindLinkTracking() {
+        $('a[data-link-tracking]').each(function (domElem) {
+            bean.on(domElem, 'click', function () {
+                trackLink($(this).attr('href'));
+            });
+        });
+    }
+
+    function trackLink(link) {
+        omniture = omniture || init();
+        omniture.then(function(){
+            s.tl(true, 'o', link);
+        });
+    }
+
+    function trackEvent(prop17, pageName, products) {
         omniture = omniture || init();
 
         omniture.then(function(){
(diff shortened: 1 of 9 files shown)


assets/javascripts/modules/analytics/omniture.js
```diff
@@ -51,6 +53,18 @@ define([
         s.eVar19 = 'D=c19';
         s.prop42 = 'Subscriber';
         s.prop47 = 'UK';
+
+        var main = $('main')[0];
```
davidrapson: Can we do this in a different way? It's completely by chance that we are using the `<main>` element for this, probably at least add a `.js-` class.

Couple of minor points: If you are just plucking the element out with `[0]` there's no need to use `$` and we could just use `querySelector` directly. The `map` could be a `forEach` as we're not actually returning anything (i.e., performing side-effects) nor using the result.

This code feels a bit magic generally though.

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -5,10 +5,25 @@ define([
...
     'use strict';
 
-    var dataTrackingClickableElements = ['a', 'button', 'input'],
-        omniture, s;
+    var omniture, s;
 
-    function sendEvent(prop17, pageName, products) {
+
+    function bindLinkTracking() {
+        $('a[data-link-tracking]').each(function (domElem) {
```
davidrapson: Are we OK to scope this to an `a` element?

davidrapson:
A few comments but broadly :+1: . Omniture code is pretty gross in general so it would be great to document this a bit more possibly (another PR). i.e, how do we know what `prop17 ` means.

chrisjowen:
Because of the time requirements of this, I have additionally fixed other inconsistencies with omniture which were needed to meet the initial requirements of this story. Now the PR has changed, please can you review, not including the comments which I will address.

assets/javascripts/modules/analytics/omniture.js
chrisjowen: Its OK for now because now a elements are the only one firing these events. If in the future other elements are used to fire these events then I think we should support at that time, but not before.

assets/javascripts/modules/analytics/omniture.js
chrisjowen: 1) Agreed, actually this is additional data for the page load so I should really add these to the body tag. 

2) Just used map for better browser support, but thats just an old mindset of mine as forEach it pretty well supported. 

3) Agree, I'm just adding to the magic here, I suggest another card to do a mapping from a structured descriptive JSON object to the magic propXX numbers.

davidrapson: Hmm, [map and forEach have the same browser support afaik](http://kangax.github.io/compat-table/es5/). Just a reminder we only load the JS app in ES5 (~IE9 environments) so for application JS you can mostly use e5 methods.

Rest sounds good though.

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -51,6 +53,18 @@ define([
         s.eVar19 = 'D=c19';
         s.prop42 = 'Subscriber';
         s.prop47 = 'UK';
+
```
joelochlann: Comment to explain that you're just extracting any attribute containing data-tracking-? Took me a bit of squinting to figure it out

joelochlann:
:+1: apart from comment

chrisjowen:
image correct values coming through in prod

PR #242, opened 2015-10-06
Fix 404 for international landing page

davidrapson:
Up until last week the international edition on theguardian.com was a pretend edition and reported as UK which meant that the Subscribe link pointed to the UK subscriptions landing page.

Now that the international edition is a real edition the Subscribe link points to https://subscribe.theguardian.com/int which results in a 404.

This PR redirects requests to /int to /us as we want to display the US subscriptions page to international visitors.

@joelochlann


davidrapson:
Ingore this the `/int` link wasn't intentional so has been fixed directly in frontend https://github.com/guardian/frontend/pull/10773

PR #243, opened 2015-10-06
Add Google re-marketing code

davidrapson:
Add Google re-marketing code and split setup between core and secondary. For https://trello.com/c/KR9L326X/202-digital-pack-google-remarketing-pixel

@joelochlann

### DIFF
--- a/assets/javascripts/modules/analytics/setup.js
+++ b/assets/javascripts/modules/analytics/setup.js
@@ -1,23 +1,23 @@
-/*global guardian:true */
 define([
     'utils/cookie',
     'modules/analytics/ga',
     'modules/analytics/omniture',
+    'modules/analytics/remarketing',
     'modules/analytics/krux',
     'modules/analytics/affectv'
 ], function (
     cookie,
     ga,
     omniture,
+    remarketing,
     krux,
     affectv
 ) {
     'use strict';
 
     function init() {
-
         var analyticsEnabled = (
-            guardian.analyticsEnabled &&
+            window.guardian.analyticsEnabled &&
             !navigator.doNotTrack &&
             !cookie.getCookie('ANALYTICS_OFF_KEY')
         );
@@ -25,10 +25,13 @@ define([
         if (analyticsEnabled) {
             ga.init();
             omniture.init();
-            krux.init();
-            affectv.init();
-        }
 
+            if (!window.guardian.isDev) {
+                remarketing.init();
+                krux.init();
+                affectv.init();
+            }
+        }
     }
 
     return {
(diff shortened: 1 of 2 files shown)


assets/javascripts/modules/analytics/setup.js
```diff
@@ -1,34 +1,37 @@
...
+        !cookie.getCookie('ANALYTICS_OFF_KEY')
+    );
 
+    function init() {
         if (analyticsEnabled) {
             ga.init();
             omniture.init();
+        }
+
+        if (analyticsEnabled && !window.guardian.isDev) {
```
joelochlann: I would probably put the `if (!window.guardian.isDev)` condition inside the `if (analyticsEnabled)` block

assets/javascripts/modules/analytics/setup.js
```diff
@@ -1,34 +1,37 @@
...
 
-    function init() {
-
-        var analyticsEnabled = (
-            guardian.analyticsEnabled &&
-            !navigator.doNotTrack &&
-            !cookie.getCookie('ANALYTICS_OFF_KEY')
-        );
+    var analyticsEnabled = (
+        window.guardian.analyticsEnabled &&
```
joelochlann: Are all these values are set correctly when this module is loaded? i.e. was there definitely no reason for waiting until `init()` is called before checking them?

assets/javascripts/modules/analytics/setup.js
```diff
@@ -1,34 +1,37 @@
-/*global guardian:true */
```
joelochlann: do we need `/*global navigator:true */` here?

assets/javascripts/modules/analytics/setup.js
joelochlann: Why `window.guardian` but not `window.navigator`?

davidrapson: More of a linting thing, guardian is defined by us so using `window` tells linter it's a global w/out needing to define it.

assets/javascripts/modules/analytics/setup.js
davidrapson: Nope, we tell the linter we're in a browser context and so it knows about those globals. The global definitions are only for non-host globals (i.e., ones we define)

assets/javascripts/modules/analytics/setup.js
davidrapson: Agreed

assets/javascripts/modules/analytics/setup.js
davidrapson: > Are all these values set correctly when this module is loaded? i.e. was there definitely no reason for waiting until init() is called before checking them?

They get set correctly, but yeah this can just go inside init.

joelochlann: Ha, you're quite right, I'd forgotten `navigator` is actually a native object!

joelochlann:
:+1: Looks good

PR #244, opened 2015-10-06
Omniture tracking only happens in one place and is mapped to a decent…

chrisjowen:
In order to proceed with dashboard tracking we needed to clean up some of the ways that omniture tracking ocured. Now instead of magik propx we update a js object on guardian.pageInfo and the mapping occurs once in omniture.js

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -32,9 +32,7 @@
     bodyClasses = List("is-wide"),
     touchpointBackendResolutionOpt = Some(touchpointBackendResolution)
 ) {
-<main class="page-container gs-container"
-      data-tracking-prop17="GuardianDigiPack:Name and address"
-      data-tracking-products="Subscriptions and Membership;GUARDIAN_DIGIPACK;@{selectedProduct.frequency.numberOfMonths};@{selectedProduct.price};scOpen">
+<main class="page-container gs-container" data-no-page-load-tracking >
 
     <section class="checkout-container">
         <form class="form js-checkout-form" action="@routes.Checkout.renderCheckout().toString()" method="POST">
--- a/app/views/checkout/thankyou.scala.html
+++ b/app/views/checkout/thankyou.scala.html
@@ -13,11 +13,18 @@
 
 @main("Confirmation | Digital | Subscriptions | The Guardian", bodyClasses=List("is-wide"), touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
 
-<main class="page-container gs-container gs-container--slim"
-data-tracking-prop17="GuardianDigiPack:Order Complete"
-data-tracking-products="Subscriptions and Membership;GUARDIAN_DIGIPACK;@{subscriptionProduct.frequency.numberOfMonths};@{subscriptionProduct.price};purchase">
->
+<script type="text/javascript">
+    guardian.slug="GuardianDigiPack:Order Complete";
+    guardian.product = {
+        source: 'Subscriptions and Membership',
+        type: 'GUARDIAN_DIGIPACK',
+        eventName: 'purchase',
+        amount: @{subscriptionProduct.price},
+        frequency: @{subscriptionProduct.frequency.numberOfMonths}
+    };
+</script>
 
+<main class="page-container gs-container gs-container--slim">
     @fragments.page.header("Thank you for your order", None, List("l-padded"))
 
     <section class="section-slice section-slice--bleed section-slice--limited">
--- a/app/views/checkout/thankyou.scala.html
+++ b/app/views/checkout/thankyou.scala.html
@@ -14,8 +14,8 @@
 @main("Confirmation | Digital | Subscriptions | The Guardian", bodyClasses=List("is-wide"), touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
 
 <script type="text/javascript">
-    guardian.slug="GuardianDigiPack:Order Complete";
-    guardian.product = {
+    guardian.pageInfo.slug="GuardianDigiPack:Order Complete";
+    guardian.pageInfo.productData = {
         source: 'Subscriptions and Membership',
         type: 'GUARDIAN_DIGIPACK',
         eventName: 'purchase',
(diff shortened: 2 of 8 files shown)


app/views/checkout/payment.scala.html
```diff
@@ -32,9 +32,7 @@
     bodyClasses = List("is-wide"),
     touchpointBackendResolutionOpt = Some(touchpointBackendResolution)
 ) {
-<main class="page-container gs-container"
-      data-tracking-prop17="GuardianDigiPack:Name and address"
-      data-tracking-products="Subscriptions and Membership;GUARDIAN_DIGIPACK;@{selectedProduct.frequency.numberOfMonths};@{selectedProduct.price};scOpen">
+<main class="page-container gs-container" data-no-page-load-tracking >
```
davidrapson: Could we define whether we enable this in the global config too? Maybe defined in the controller?

app/views/checkout/thankyou.scala.html
```diff
@@ -13,11 +13,18 @@
 
 @main("Confirmation | Digital | Subscriptions | The Guardian", bodyClasses=List("is-wide"), touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
 
-<main class="page-container gs-container gs-container--slim"
-data-tracking-prop17="GuardianDigiPack:Order Complete"
-data-tracking-products="Subscriptions and Membership;GUARDIAN_DIGIPACK;@{subscriptionProduct.frequency.numberOfMonths};@{subscriptionProduct.price};purchase">
->
-
+<script type="text/javascript">
+    guardian.slug="GuardianDigiPack:Order Complete";
```
davidrapson: Would prefer to avoid script blocks in the middle of the markup, could we define these extra properties in a controller (or in scala anyway) and then conditionally expose them in the config in `javaScriptFirstSteps`

app/views/fragments/javascriptFirstSteps.scala.html
```diff
@@ -96,3 +96,18 @@
         }
     })(guardian.isModernBrowser);
 </script>
+
+
+<script id="gu_page_info">
+    guardian.pageInfo = {
```
davidrapson: Minor niggle, but personally I'd find it clearer to define this next to the other `guardian.` definitions in `id="gu"`. We could always move the `isModernBrowser` scoped code into another script block if that aids clarity. If that made any sense.

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -18,73 +17,72 @@ define([
...
-    }
-
-    function onSuccess() {
-        var s_code;
-
-        window.s_account = 'guardiangu-subscribe,guardiangu-network';
-        s = s_gi('guardiangu-network');
+    function setTracker(s) {
+        var pageInfo = guardian.pageInfo,
+            productData = guardian.productData;
```
davidrapson: I'd prefer to scope this to a single key, so `slug, pageInfo , productData` under one key.

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -18,73 +17,72 @@ define([
...
 
-        s_code = s.t();
+    function triggerPageLoadEvent() {
+        omniture = omniture || init();
+        omniture.then(function () {
+            setTracker(s);
+            var s_code = s.t();
+            setTracker(s);
+            if (s_code) {
+                /*jslint evil: true */
```
davidrapson: Not your code but `/*jslint evil: true */` can go as we don't use `jslint` anymore so this line does nothing.

*However* we should have a quick look (as another PR) into whether we can avoid using `document.write` completely.

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -18,73 +17,72 @@ define([
...
+        });
+    }
 
-        if (s_code) {
-            /*jslint evil: true */
-            document.write(s_code);
+    function onSuccess() {
+        window.s_account = 'guardiangu-subscribe,guardiangu-network';
+        s = s_gi('guardiangu-network');
+        if(!document.querySelector('main[data-no-page-load-tracking]')) {
```
davidrapson: As commented above, I'd rather expose this in the `guardian.*` global object.

assets/javascripts/modules/analytics/omniture.js
chrisjowen: Yeah I was wondering about this, that appears to be a hack for older browsers which we arent supporting.

assets/javascripts/modules/analytics/omniture.js
chrisjowen: these SHOULD all be under the pageInfo key, its my mistake.

app/views/fragments/javascriptFirstSteps.scala.html
chrisjowen: Ok, done.

davidrapson:
:+1: Great clean up. My comments are mostly around *how* we structure and expose the data, the clean up is great.

app/views/checkout/thankyou.scala.html
chrisjowen: Im not sure I agree, I think this is view specific stuff and would rather not have it in the controller but I see there is similar with the JSVars so ill stick with it for consistency.

assets/javascripts/modules/analytics/omniture.js
davidrapson: Bingo!

https://marketing.adobe.com/developer/forum/general-topic-forum/why-document-write
> (Netscape 3+, IE4+)

I *think* we might be alright…

app/views/checkout/payment.scala.html
chrisjowen: Make sense

app/views/checkout/thankyou.scala.html
davidrapson: Yeah maybe, I think it's worth the slight abuse of MV* in exchange for the clarity of a single place where we expose global properties. Plus it encourages scripts in the middle of the page, which can cause blocking if we're not careful.

PR #245, opened 2015-10-07
Move DigitalEdition objects to model, add redirect function

tomverran:
This PR will result in the interstitial UK / RoW country select page being skipped for users entering the signup process from the AU / US editions. UK users will still get the page for tracking / A-B testing

I also wrote a test, which necessitated moving the DigitalEdition objects into model/, which seems like a good move regardless of whether the test is considered worthwhile or not.

### DIFF
--- a/app/model/DigitalEdition.scala
+++ b/app/model/DigitalEdition.scala
@@ -12,7 +12,7 @@ object DigitalEdition {
   def getRedirect(edition: DigitalEdition): String = {
     edition match {
       case UK => "/digital/country"
-      case DigitalEdition(id, name, price, cmp) => "https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=" + cmp
+      case DigitalEdition(_, _, _, cmp) => "https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=" + cmp
     }
   }
 }
--- a/app/model/DigitalEdition.scala
+++ b/app/model/DigitalEdition.scala
@@ -9,8 +9,8 @@ object DigitalEdition {
   object US extends DigitalEdition("us", "US", "$19.99", "dis_2378")
   object AU extends DigitalEdition("au", "Australia", "$21.50", "dis_2379")
 
-  def getRedirect(e: DigitalEdition): String = {
-    e match {
+  def getRedirect(edition: DigitalEdition): String = {
+    edition match {
       case UK => "/digital/country"
       case DigitalEdition(id, name, price, cmp) => "https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=" + cmp
     }
--- a/app/model/DigitalEdition.scala
+++ b/app/model/DigitalEdition.scala
@@ -1,18 +1,25 @@
 package model
 
+import configuration.Config
+
 case class DigitalEdition(id: String, name: String, price: String, cmp: String)
 
 
 object DigitalEdition {
 
+  trait UrlConfig {
+    val externalSubscriptionUrl: String
+  }
+
   object UK extends DigitalEdition("uk", "UK", "£11.99", "dis_2408")
   object US extends DigitalEdition("us", "US", "$19.99", "dis_2378")
   object AU extends DigitalEdition("au", "Australia", "$21.50", "dis_2379")
+  implicit lazy val defaultUrlConfig: UrlConfig = Config
 
-  def getRedirect(edition: DigitalEdition): String = {
+  def getRedirect(edition: DigitalEdition)(implicit urlConfig: UrlConfig): String = {
     edition match {
       case UK => "/digital/country"
-      case DigitalEdition(_, _, _, cmp) => "https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=" + cmp
+      case DigitalEdition(_, _, _, cmp) => urlConfig.externalSubscriptionUrl + "?prom=DGA38&CMP=" + cmp
     }
   }
 }
--- a/app/model/DigitalEdition.scala
+++ b/app/model/DigitalEdition.scala
@@ -2,7 +2,7 @@ package model
 
 import configuration.Config
 
-case class DigitalEdition(id: String, name: String, price: String, cmp: String)
+case class DigitalEdition(id: String, name: String, price: String, campaign: String)
 
 
 object DigitalEdition {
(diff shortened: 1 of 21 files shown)


test/model/DigitalEditionTest.scala
```diff
@@ -0,0 +1,16 @@
+package model
+
+import model.DigitalEdition.{US, UK}
+import play.api.test.PlaySpecification
+
+class DigitalEditionTest extends PlaySpecification {
+    "getRedirect" should {
```
joelochlann: I think the general convention for indentation in Scala is two spaces

app/model/DigitalEdition.scala
```diff
@@ -0,0 +1,18 @@
...
+object DigitalEdition {
+
+  object UK extends DigitalEdition("uk", "UK", "£11.99", "dis_2408")
+  object US extends DigitalEdition("us", "US", "$19.99", "dis_2378")
+  object AU extends DigitalEdition("au", "Australia", "$21.50", "dis_2379")
+
+  def getRedirect(edition: DigitalEdition): String = {
+    edition match {
+      case UK => "/digital/country"
+      case DigitalEdition(_, _, _, cmp) => "https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=" + cmp
```
joelochlann: could we move this URL out to a config somewhere?

joelochlann:
:+1: apart from some minor comments. You may want to get someone who knows the codebase better to give it a quick look over too though. Maybe @rtyley?

tomverran:
@joelochlann made some changes to put the URL into a config file, probably worth rechecking

joelochlann:
Looks good :+1: :+1:

app/model/DigitalEdition.scala
```diff
@@ -0,0 +1,25 @@
+package model
+
+import configuration.Config
+
+case class DigitalEdition(id: String, name: String, price: String, cmp: String)
```
joelochlann: Actually, what is `cmp`? I know you just moved this code rather than writing it but am wondering if we could improve the name of this field so that it's more self-explanatory

tomverran: ha yeah I have no idea, I held off renaming it until I could find out
I shall do so though!

rtyley: Yeah, I believe it's really 'campaign code' - so an identifier for some given advertising campaign.

app/model/DigitalEdition.scala
```diff
@@ -0,0 +1,25 @@
...
+  trait UrlConfig {
+    val externalSubscriptionUrl: String
+  }
+
+  object UK extends DigitalEdition("uk", "UK", "£11.99", "dis_2408")
+  object US extends DigitalEdition("us", "US", "$19.99", "dis_2378")
+  object AU extends DigitalEdition("au", "Australia", "$21.50", "dis_2379")
+  implicit lazy val defaultUrlConfig: UrlConfig = Config
+
+  def getRedirect(edition: DigitalEdition)(implicit urlConfig: UrlConfig): String = {
```
afiore: I think you are reaching out to heavy artillery (i.e. `implicit urlConfig: UrlConfig`) to do something fairly simple here. Why not just provide a base url as the first function argument?

```scala
def getRedirect(baseUrl: URL, edition: DigitalEdition): URL = ???
```
Also, Perhaps it is a good idea to use `java.net.URL` instead of String, at least for the baseUrl value (Typesafe config provides a convenient [parseUrl](https://typesafehub.github.io/config/latest/api/com/typesafe/config/ConfigFactory.html#parseURL-java.net.URL-) method to read these from config files?

tomverran: Ah that ```java.net.URL``` tip is a good one!

With the use of the ```UrlConfig``` trait - if the urlConfig argument becomes a baseUrl string then the controller calling getRedirect will have to become dependant on the global Config super object to retrieve the correct value to use.

By using a trait like this if we can move the dependency on Config from the caller to

```scala
implicit lazy val defaultUrlConfig: UrlConfig = Config
```

Which keeps the controller free of global state and allows us to swap out the source of Config in one place if the function ever gets repeatedly called from different scopes

The use of an ```implicit val``` to inject the chosen implementation of ```UrlConfig``` is pretty dubious on my part, you'd achieve the same thing with just a normal val somewhere.

I know this is a lot of complexity for a very simple feature, I did it like this to find out what people thought of dependency injection patterns such as this in general.

Given that the controllers are unlikely ever to be unit tested I'll swap to using a ``java.net.URL``` URL argument for simplicity's sake :smile:

davidrapson: You can always have a chat in person for this kind of thing :smile: https://github.com/guardian/recommendations/blob/master/pull-requests.md#talk-is-better

That said: I might be exposing my naivety here, but if we just use the base URL in one method and it doesn't change based on environment does it even need to be in the config?

joelochlann: @davidrapson That was Tom's original solution but I suggested possibly moving it to a config. So you can blame me for that one

tomverran: ah thats interesting I don't have access to that repo!
And yes, I think you might be right about the config on second thoughts, I think I'll revert to 58afade, Joe and I just had a brief chat about it

rtyley: Just to follow up on `java.net.URL` - unfortunately it's quite a **bad** class, see  http://www.eishay.com/2008/04/javas-url-little-secret.html or
http://michaelscharf.blogspot.co.uk/2006/11/javaneturlequals-and-hashcode-make.html

Because of this, `java.net.URI` is preferable - or the representation in https://github.com/NET-A-PORTER/scala-uri , which we already use elsewhere.

joelochlann:
:+1:

app/views/digitalpack/country.scala.html
```diff
@@ -18,7 +19,7 @@ <h2 class="country-select__title">The Digital Pack is available throughout the w
                     <h3 class="u-pad-bottom">Choose your location</h3>
                     <div class="button-group">
                         <a class="button button--primary button--large js-checkout-link" data-link-tracking
-                           data-previous-href="https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=@edition.cmp"
+                           data-previous-href="https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=DGA38&CMP=@edition.campaign"
```
craigjbass: Excellent use of full names

test/model/DigitalEditionTest.scala
```diff
@@ -0,0 +1,16 @@
+package model
+
+import model.DigitalEdition.{US, UK}
+import play.api.test.PlaySpecification
+
+class DigitalEditionTest extends PlaySpecification {
```
craigjbass: Nice use of testing

PR #246, opened 2015-10-07
Event tracking

chrisjowen:
Snowplow tracking added for all page loads and checkout start and registration events.


app/views/digitalpack/country.scala.html
```diff
@@ -3,36 +3,35 @@
 @main("Select your country | Digital | Subscriptions | The Guardian", bodyClasses = List("is-wide")) {
     <main class="page-container gs-container">
 
-    <section class="checkout-container">
+        <section class="checkout-container">
```
davidrapson: I can't tell what's changed in this file?

app/views/fragments/footerJavaScript.scala.html
```diff
@@ -10,3 +11,12 @@
         }
     })(guardian.isModernBrowser);
 </script>
+
+
+
+<script type="text/javascript">
+    var trackerUrl = '@{Config.trackerUrl}';
+    snowplow_name_here("newTracker", "subscriptions", "@{Config.trackerUrl}", {
```
davidrapson: `snowplow_name_here`?

app/views/digitalpack/country.scala.html
tomverran: if you suffix the PR files/ URL with ?w=1 whitespace only line changes get ignored, which is handy for situations like this

app/views/fragments/javascriptFirstSteps.scala.html
```diff
@@ -108,3 +106,9 @@
 </script>
 
 
+<script type="text/javascript" async=1>
+;(function(p,l,o,w,i,n,g){if(!p[i]){p.GlobalSnowplowNamespace=p.GlobalSnowplowNamespace||[];
+p.GlobalSnowplowNamespace.push(i);p[i]=function(){(p[i].q=p[i].q||[]).push(arguments)
+};p[i].q=p[i].q||[];n=l.createElement(o);g=l.getElementsByTagName(o)[0];n.async=1;
+n.src=w;g.parentNode.insertBefore(n,g)}}(window,document,"script","//d1fc8wv8zag5ca.cloudfront.net/2.5.1/sp.js","snowplow_name_here"));
```
davidrapson: Hmm, couple of things here:

- I'd prefer to either add this to `vendor` or maybe add to bower if it's versioned.
- Also, can we move this into the main app js and only load if the app has been booted?

assets/javascripts/modules/analytics/omniture.js
```diff
@@ -66,7 +60,7 @@ define([
     function onSuccess() {
         window.s_account = 'guardiangu-subscribe,guardiangu-network';
         s = s_gi('guardiangu-network');
-        if(!document.querySelector('main[data-no-page-load-tracking]')) {
+        if (!document.querySelector('main[data-no-page-load-tracking]')) {
```
davidrapson: Mentioned in the omniture PR but could we define whether we enable this in the global config, rather than via an attr on main?

assets/javascripts/modules/analytics/setup.js
```diff
@@ -4,25 +4,24 @@ define([
...
+             affectv,
+             snowplow) {
     'use strict';
 
     function init() {
         var analyticsEnabled = (
-            window.guardian.analyticsEnabled &&
-            !navigator.doNotTrack &&
-            !cookie.getCookie('ANALYTICS_OFF_KEY')
+            window.guardian.analyticsEnabled && !navigator.doNotTrack && !cookie.getCookie('ANALYTICS_OFF_KEY')
```
davidrapson: Why the change to a one-liner?

assets/javascripts/modules/analytics/snowplow.js
```diff
@@ -0,0 +1,82 @@
...
+        if (productData) {
+            data.productBillingFrequency = productData.frequency.toString();
+            data.productBillingAmount = productData.amount;
+            data.productType = productData.type;
+        }
+
+        trackActivity('pageLoaded', data);
+    }
+
+    function init() {
```
davidrapson: Mentioned above, but could we handle the actual loading of the snowplow js and setup in here too, rather than as inline script blocks?

assets/javascripts/modules/analytics/snowplow.js
```diff
@@ -0,0 +1,82 @@
+/*global snowplow_name_here, guardian */
+define(['lodash/object'], function (_) {
```
davidrapson: It's not immediately obvious, but it's preferable to only load the methods you need as `lodash/object` will include *all* object methods in our build. You only seem to need `merge` so you *should* be able to include `lodash/object/merge` and then name it as `merge` instead of `_`

assets/javascripts/modules/checkout/tracking.js
```diff
@@ -21,20 +21,23 @@ define(['$', 'modules/analytics/omniture'], function ($, omniture) {
...
         omniture.triggerPageLoadEvent();
+	snowplow.trackPageLoad();
     }
 
-    function paymentReviewTracking(){
+    function paymentReviewTracking() {
         guardian.pageInfo.slug = 'GuardianDigiPack:Review and confirm';
-        guardian.pageName ='Payment submission/signup | Digital | Subscriptions | The Guardian';
-        guardian.pageInfo.productData =  subscriptionProducts('scCheckout');
+	guardian.pageName = 'Payment submission/signup | Digital | Subscriptions | The Guardian';
```
davidrapson: Is this a page title? Does `document.title` give us what we want?

davidrapson:
A few comments here so probably worth having a chat to get some more context on these changes.

assets/javascripts/modules/analytics/setup.js
chrisjowen: This is my annoying intellij auto reformat settings, I agree this was more readable before

app/views/fragments/javascriptFirstSteps.scala.html
chrisjowen: Yeah I agree the cloudfront url is odd but this is there documented way of loading the script. I can move this out of javascriptFirstSteps to vendor though as suggested

app/tracking/activities/SubscriptionRegistrationActivity.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def truncatePostcode(postcode: String) = {
+    postcode.splitAt(postcode.length - 3)._1.trim
+  }
+
+}
+
+case class SubscriptionRegistrationActivity(memberData: MemberData) extends TrackerData {
+  override def toMap: JMap[String, Object] = {
+    println(memberData.toMap)
```
dominickendrick: remove me!

app/views/fragments/footerJavaScript.scala.html
chrisjowen: Tracker URL was the only reason, so yes I could just expose that and move the rest into a require. The name "snowplow_name_here" is what that horrible script loader with the cloudformation name you commented on brings into the global namespace. Yes I agree its all very funky, just followed the docs here but I think you're right that it needs cleaning up.

app/views/fragments/javascriptFirstSteps.scala.html
davidrapson: Yeah, move it into vendor and then probably give it a name in `javascriptLaterSteps` like we do for omniture. You can then use something like `require(['js!snowplow', function() {})` to conditionally load it in. Probably worth initialising it in `analytics/setup` and guarding it with the `analyticsEnabled` check too.

assets/javascripts/modules/analytics/setup.js
davidrapson: :ok_hand:

app/tracking/activities/SubscriptionRegistrationActivity.scala
chrisjowen: Why?

chrisjowen:
@davidrapson  all changes requested have been made

app/views/fragments/javascriptFirstSteps.scala.html
```diff
@@ -29,13 +29,14 @@
...
         name: document.title,
-        publication: 'GU.co.uk',
-        commissioningDesk: 'Subscriber',
         product: 'Subscriber',
-        edition: 'UK',
         productData: null
     };
 
+    guardian.trackerUrl = '@{Config.trackerUrl}';
+    guardian.ignorePageLoadTracking = @{jsVars.ignorePageLoadTracking};
```
davidrapson: I *think* this might already get merged here: https://github.com/guardian/subscriptions-frontend/blob/event-tracking/app/views/fragments/javascriptFirstSteps.scala.html#L16 so you might be overriding it here?

(I don't love having this JsVars stringify stuff at the root `guardian`, if it's dynamic we should move it down a level to avoid conflicts. But that's a problem for another day)

davidrapson:
Needs merge and you probably wan't someone else to review the Scala by :+1: for the JS.

tomverran:
scala looks ok to me :+1:

PR #247, opened 2015-10-07
Use a separate domain for Identity cookies

ostapneko:
This should fix the problem highlighted in https://github.com/guardian/subscriptions-frontend/pull/236

@rtyley

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -32,8 +32,6 @@ object Config {
   val stage = config.getString("stage")
   val stageProd: Boolean = stage == "PROD"
 
-  lazy val sessionDomain = config.getString("session.domain")
-
   object Identity {
     private val idConfig = config.getConfig("identity")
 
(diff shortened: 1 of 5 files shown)


tomverran:
Looks good to me, in that GU_U cookies and their ilk are on . while PLAY_SESSION is on sub.
Only concern is how come sessionDomain is camelCase while every other setting is dotted?

If this is to avoid a conflict with some built in play config settings could it be prefixed or something? Isn't it already prefixed in that it is identity.session.domain as opposed to session.domain?

rtyley:
> If this is to avoid a conflict with some built in play config settings could it be prefixed or something? Isn't it already prefixed in that it is identity.session.domain as opposed to session.domain?
> 
> ccing @rtyley - I know you want this live, are my quibbles misplaced?

I think you're right that we can+should use `identity.session.domain` for consistency, but let's just get this out for now...

rtyley:
Confirmed on PROD

Play cookies are cookies are correctly dropped on `subscribe.theguardian.com`:

image

On completing the guest user with a password, cookies are correctly dropped on `.theguardian.com`:

image

PR #241, opened 2015-10-05
Add Signed In Navigation

davidrapson:
Adds signed in status and account navigation to the header.

screen shot 2015-10-05 at 14 24 41

**Header before**
screen shot 2015-10-05 at 12 58 49

**Header after**
screen shot 2015-10-05 at 12 58 54

PR #248, opened 2015-10-08
Apply raven improvement from Membership

davidrapson:
Applies some of the Raven improvements @tomverran and I worked on for Subscriptions.

- Adds logging in development mode, to give visibility into what Raven is reporting
- Moves raven config into separate module and avoids, implicit require
- Ignore some noisy errors

Once https://github.com/guardian/subscriptions-frontend/pull/241 has gone out we will also be able to record the identity ID for logged in users.

@tomverran

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,6 +10,7 @@
     },
     "rules": {
         "curly": [2, "all"],
+        "no-console": 0,
         "no-trailing-spaces": 2,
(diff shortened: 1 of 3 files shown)


tomverran:
:+1 LGTM, would be nice to resolve the duplicate code between the two repos one day in the far future

PR #249, opened 2015-10-08
Show verbose compiler warnings when running sbt

tomverran:
Tiny PR just to enable -feature when compiling with SBT so warnings get displayed in full


dominickendrick:
:+1:

PR #250, opened 2015-10-08
Reduce noise with script-loading errors

davidrapson:
We currently use the requirejs script loader to load our analytics code. However it's a bit of a sledgehammer to crack a nut. As it's intended for loading AMD/application code it throws and Error when a script failed to load.

Sentry catches these and reports them, this is usually a good thing but for some third-party code we don't really care if the script fails (or certainly don't want it throwing an Error).

For non-critical third-party code using a simpler script loading method allows us to fail more gracefully if the script is unavailable.

@rtyley

### DIFF
--- a/assets/javascripts/modules/analytics/affectv.js
+++ b/assets/javascripts/modules/analytics/affectv.js
@@ -1,5 +1,4 @@
-/*global Raven */
-define(function() {
+define(['utils/loadJs'], function(loadJs) {
     'use strict';
 
     var digitalPackId = '546dd61001896028ffd29273';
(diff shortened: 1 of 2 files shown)


rtyley:
:+1: to reducing Sentry noise due to client side events. Shame that require.js doesn't support us choosing how severe we view a resource failing to load.

davidrapson:
So conceptually I actually prefer not using the `js!` loader for this kind of thing as it's really the wrong use-case. We're using an alternative CurlJS version too which comes with this warning…

https://github.com/cujojs/curl/wiki/js#do-not-use-this-plugin-

> DO NOT USE THIS PLUGIN :)
> I wanted to name this plugin legacy! in stead of js!, but in respect of the several 2005-era javascript libraries out there, I refrained. If you're serious about AMD and/or javascript modules, then you should be using all modules, all of the time.

PR #255, opened 2015-10-13
Be more defensive with country metrics code

davidrapson:
Avoids errors when `COUNTRY_FLOW` is set but not on country select page as code gets called on any page. This PR updates the JS to be a bit more defensive as to when it runs.

screen shot 2015-10-13 at 17 33 46

@chrisjowen

### DIFF
--- a/assets/javascripts/modules/country.js
+++ b/assets/javascripts/modules/country.js
@@ -22,6 +22,15 @@ define([
         elements.$CHECKOUT_LINK.attr('href', elements.$CHECKOUT_LINK.attr('data-previous-href'));
     }
 
+    function recordRedirect() {
+        var link = elements.$CHECKOUT_LINK[0];
+        if (link) {
+            bean.on(link, 'click', function () {
+                snowplow.trackActivity('redirectedToQss');
+            });
+        }
+    }
+
     function init() {
         guardian.pageInfo.productData = {
             source: 'Subscriptions and Membership',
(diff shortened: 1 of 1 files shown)


joelochlann:
:+1: But I can't resist pointing out that this is one of the cases where jQuery (or [zepto](http://zeptojs.com/) if you want a lighter version) would make things much nicer. If you did something like
```
$('.checkout-link').on('click', function() { ... }
```
and couldn't find any `.checkout-link` element, you would get an empty wrapped set, which would just call the method in question on zero items. So you get an elegant guard against these kind of errors.

joelochlann:
Or maybe what's needed is for someone to extend [bonzo](https://github.com/ded/bonzo) with event handling? Then we could do away with `bean` which I'm not too impressed by. You may have arguments to the contrary though :smile:

davidrapson:
This is the same case with using `$` directly in our codebase. There would be no reason to add event-handlers to bonzo as they are deliberately designed to handle separate tasks. And intended to be wrapped together, as in [$.js](https://github.com/guardian/subscriptions-frontend/blob/master/assets/javascripts/utils/$.js)

joelochlann:
Yeah but you don't get the `$('.blah').on('click', ...)` syntax, right? Unless I've missed something. You'd have to do
```
$('.blah').each(function(elem) {
  bean.on(elem, 'click', function(e) {
   ...
  }
});
```
Seems unnecessarily verbose to me. Maybe we could modify [$.js](https://github.com/guardian/subscriptions-frontend/blob/master/assets/javascripts/utils/$.js) to achieve the simpler syntax

davidrapson:
Yep, agree with that for sure. Or perhaps https://github.com/filamentgroup/shoestring would be a good fit.

joelochlann:
Yeah, looks nice. You're right, the micro-modular approach makes more sense than including a monothilic library with loads of stuff you don't need. There's just some a few bits of syntactic sugar that I'm really missing with our current crop of libraries. One for [the to-do list](https://trello.com/c/cPhMzlAm/200-allow-jquery-like-syntax-for-event-binding)

joelochlann:
But feel free to merge this :+1: :+1:

PR #256, opened 2015-10-15
Handle credit cards from the backend

ostapneko:
This provides backend support for different payment methods.
Upon form submission, in addition to the existing `payment.*` fields, it requires a `payment.type` field with two possible values: `credit-card` and `direct-debit`. In the former case, another field, `payment.token`, containing the Stripe token, is expected.

Note that the backend code is *untested*; we will need to wait for the front-end to be ready before writing end-to-end tests.

~~Also, note that the existing integration tests are broken right now on master, so we will need to wait for them to be fixed before being able to perform regression tests~~

@joelochlann @afiore @rtyley


afiore:
Overall :+1: and also agreed on the need of adding coverage for this in the end-to-end tests.
While discussing the changes in this PR, we also discussed the possibility of moving the `PaymentData` trait and maybe some of its related form machinery into membership-commons.

app/controllers/Checkout.scala
```diff
@@ -93,7 +93,7 @@ object Checkout extends Controller with LazyLogging with ActivityTracking {
         products <- zuoraService.products
         product = products.find(p => p.ratePlanId == formData.ratePlanId)
       } yield {
```
joelochlann: Not your change but out of curiosity, does this need to be a `yield` if we're not using the resulting expression?

joelochlann:
A belated :+1: from me. Can we get this merged please?

PR #258, opened 2015-10-16
Analytics should not fire if using test users 

chrisjowen:
Currently the analytics setup component was guarding on initialisation of the various analytics components we used. However, there were other entry points exposed to call analytics which made the data we were getting a little off. This PR adds the additional guards around those exposed calls. 

See https://trello.com/c/lshDPxMa/213-ensure-analytics-always-respect-analytics-off-key-cookie for more informaiton

### DIFF
--- a/assets/javascripts/modules/analytics/analyticsEnabled.js
+++ b/assets/javascripts/modules/analytics/analyticsEnabled.js
@@ -1,0 +1,19 @@
+define([
+    'utils/cookie'
+], function (cookie) {
+    'use strict';
+
+    var analyticsEnabled = (
+        window.guardian.analyticsEnabled &&
+        !navigator.doNotTrack &&
+        !cookie.getCookie('ANALYTICS_OFF_KEY')
+    );
+
+    return function(cb){
+        return function() {
+            if (analyticsEnabled) {
+                return cb()
+            }
+        }
+    }
+});
--- a/assets/javascripts/modules/country.js
+++ b/assets/javascripts/modules/country.js
@@ -23,9 +23,8 @@ define([
     }
 
     function recordRedirect() {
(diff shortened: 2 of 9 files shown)


assets/javascripts/modules/country.js
```diff
@@ -23,9 +23,8 @@ define([
     }
 
     function recordRedirect() {
-        var link = elements.$CHECKOUT_LINK[0];
-        if (link) {
-            bean.on(link, 'click', function () {
+        if (elements.$CHECKOUT_LINK && elements.$CHECKOUT_LINK.length==1) {
```
davidrapson: Not sure you need the `==1` but if, so maybe use strict equality `===`?

assets/javascripts/modules/analytics/analyticsEnabled.js
```diff
@@ -0,0 +1,19 @@
...
+
+    var analyticsEnabled = (
+        window.guardian.analyticsEnabled &&
+        !navigator.doNotTrack &&
+        !cookie.getCookie('ANALYTICS_OFF_KEY')
+    );
+
+    return function(cb){
+        return function() {
+            if (analyticsEnabled) {
```
davidrapson: Not quite sure what this gives us over just returning `analyticsEnabled`.

davidrapson:
:+1:

assets/javascripts/modules/analytics/analyticsEnabled.js
chrisjowen: Either way works for me, the only benefit I can see this way is that if we wanted to toggle tracking for events that load after page loaded then we could. I dont think there a strong argument either way.

PR #264, opened 2015-10-28
Removed the 'so domething' Complete supplements on digi pack page

chrisjowen:
Simplet test change, removing "Do Something" from digipack Complete supplements
image


ostapneko:
:+1:

PR #265, opened 2015-10-29
Add manifest endpoint so status app knows our build version

rtyley:
So that we can have build version numbers here:

http://status.membership.guardianapis.com/PROD

image

The Status App reads those version numbers:

https://github.com/guardian/status-app/blob/0a77637/app/model/Instance.scala#L163

Implementation mostly stolen from:

https://github.com/guardian/status-app/blob/0a7763/app/controllers/Management.scala

cc @philwills

PR #269, opened 2015-11-06
Allow StatusApp to hit EC2 instances so it can get their version

rtyley:
I moaned that version information wasn't showing in Status App, and Phil said:

https://github.com/guardian/subscriptions-frontend/pull/265#issuecomment-152990000

So I copied the tiny bit of CF from Ophan:

https://github.com/guardian/ophan/blob/601868/cloudformation/dashboard.template.yaml#L121

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-syntax

Finally, we can see the machine build version:

image

cc @tomverran

### DIFF
--- a/cloud-formation/subscriptions-app.cf.json
+++ b/cloud-formation/subscriptions-app.cf.json
@@ -36,6 +36,10 @@
             "Type" : "String",
             "Default" : "0.0.0.0/0"
         },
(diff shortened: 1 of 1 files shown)


ostapneko:
:+1:

PR #270, opened 2015-11-06
Compatibility with membership-common 0.99

tomverran:
Soap configs needed updating, also the mandate ID is now an option

@joelochlann do you think we should throw an exception from SubscriptionDataExtensionRow if the mandateId is None at that point? The behaviour in this PR is to default it to "" which is probably wrong


tomverran:
Parked until credit card work gets merged in

PR #272, opened 2015-11-16
Card payment form

joelochlann:
**Important:** `subscriptions-frontend.conf` has been updated with the Stripe private key. For this to work you will need to re-download with the following command:
```aws s3 cp s3://subscriptions-private/DEV/subscriptions-frontend.conf /etc/gu  --profile membership```

This adds the ability to pay for your UK subscription by credit/debit card. We don't want to expose this functionality to real users until we have released the Manage My Account functionality, since that would allow them to set up a subscription with a credit card and then never be able to alter the payment method - even if, for instance, the card expired. So the card payment option is only exposed to test users. This will be changed once Manage My Account is released.

There's some related backend work which has already been merged in [this PR](https://github.com/guardian/subscriptions-frontend/pull/256).

Here's what the payment details section now looks like for test users:

picture 40
picture 45

Note there is still some placeholder copy on the right-hand side ("Payments will be taken on or shortly after [DATE]"). We're still waiting for confirmation on exactly what this should say. It will be updated before we expose this to real users.

### CODE REVIEW

**Comment by joelochlann** [comment]
@chrisjowen @tomverran @afiore

**Comment by chrisjowen** [comment]
For some reason I can't comment on line numbers here anymore?!?

There is quite a lot to consume here, and lot of this is FE work, so acceptance tests would be really useful here. 

Nitpicking side would say if we are swallowing failed ExactTarget email sends then we need to think about anon users who never retrieve there credentials. Also, do we really need to pass the touchpointBackendResolution factory to the view (i.e. javascriptFirstSteps) we should know the correct stripe key to use server side and thats all the view needs. 

Other than that everything *looks* right so :+1: but a testing :+1: should be added before deploy IMO.

**Comment by joelochlann** [comment]
Thanks @chrisjowen. You're right about tests. We do actually have an acceptance test for Direct Debit signup so we can verify that's still working correctly. @mario-galic and I are going to put together an acceptance test for the credit card flow today.

File: app/controllers/Checkout.scala [hunk 1]
```diff
@@ -96,7 +98,10 @@ object Checkout extends Controller with LazyLogging with ActivityTracking {
 	      product.foreach(prod => trackAnon(SubscriptionRegistrationActivity(MemberData(result, formData, prod))))
       }
 
-      Redirect(routes.Checkout.thankYou()).withSession(session)
+      Ok(Json.obj("redirect" -> routes.Checkout.thankYou().url)).withSession(session);
+    }.recover {
+      case err: soap.Error if err.code == "TRANSACTION_FAILED" => Forbidden
```
ostapneko: `Forbidden` looks a bit inaccurate; is it possible to get a more qualified exception? The checkout process goes through a lot of steps, and it's not really evident from the HTTP response that it's the payment step that fails

app/model/JsVars.scala
```diff
@@ -3,16 +3,33 @@ package model
...
+    ignorePageLoadTracking = false,
+    stripePublicKey = None
+  )
 
   implicit val jsVarsWrites = new Writes[JsVars] {
-    def writes(jsVars: JsVars) = Json.obj(
-      "user" -> Json.obj(
-        "isSignedIn" -> jsVars.userIsSignedIn,
-        "ignorePageLoadTracking" -> jsVars.ignorePageLoadTracking
+    def writes(jsVars: JsVars) = {
```
ostapneko: If we use the `Json.writes[JsVars]` macro, the optional fields don't appear when null. It might be more concise and less confusing to use the standard serialization for `JsVars` and wrap it inside a `user` field at call point.

app/model/exactTarget/SubscriptionDataExtensionRow.scala
```diff
@@ -42,21 +52,18 @@ object SubscriptionDataExtensionRow {
...
         "Address 2" -> address.lineTwo,
         "City" -> address.town,
         "Post Code" -> address.postCode,
-        "Default payment method" -> formatPaymentMethod(paymentMethod.`type`),
         //TODO hardcoded!
         "Country" -> "UK",
         "Date of first payment" -> formatDate(subscription.contractAcceptanceDate),
         "Currency" -> formatCurrency(account.currency),
         //TODO to remove, hardcoded in the template
         "Trial period" -> "14",
```
ostapneko: The trial period shouldn't be hardcoded anymore. Or will this be part of another PR instead?

app/views/checkout/payment.scala.html
```diff
@@ -60,7 +60,22 @@
...
                 <div class="u-display-from-tablet" data-set="checkout-notices">
                     <div class="js-append">
-                        @fragments.checkout.notices()
+                        <div class="js-payment-type-direct-debit">
+                            @fragments.checkout.notices()
+                        </div>
+                        <div class="js-payment-type-card" hidden="true">
+                            <div class="notice">
+                                <h3 class="notice__header">First payment</h3>
+                                <div class="notice__content">You first payment will be taken in 14 days under the name Guardian News & Media.</div>
```
ostapneko: Price hardcoding, see above

app/views/checkout/payment.scala.html
```diff
@@ -60,7 +60,22 @@
...
+                            </div>
+                            <div class="notice">
+                                <h3 class="notice__header">Ongoing payments</h3>
+                                <div class="notice__content">Your card will be charged
+                                    @products.find(_.frequency == BillingFrequency.Month).map {product => <span class="js-option-mirror-value">@product.price.pretty every month</span>}.
```
ostapneko: Can we move that out of the template? What happens if we have a `None`?

app/views/checkout/payment.scala.html
```diff
@@ -60,7 +60,22 @@
...
+                                    Payments will taken on or shortly after the [DATE].
```
ostapneko: Is it a placeholder?

ostapneko:
Is it also possible to squash the less informative commits, such as whitespace deletion?

app/views/checkout/payment.scala.html
joelochlann: Yes, the exact text here still needs to be decided. I'm going to do the dynamic payment date stuff as a separate PR.

app/views/checkout/payment.scala.html
joelochlann: Yes, this isn't great. I was copying this, which is used above to create the billing frequency options:

```scala
<fieldset class="basket-billing-options is-hidden js-payment-frequency js-option-mirror-group" id="change-payment-frequency">
    @products.find(_.frequency == BillingFrequency.Month).map { product =>
        @priceOption(product, "every month", product.frequency.numberOfMonths, isChecked=true)
    }
    @products.find(_.frequency == BillingFrequency.Quarter).map { product =>
        @priceOption(product, "every 3 months", product.frequency.numberOfMonths)
    }
    @products.find(_.frequency == BillingFrequency.Annual).map { product =>
        @priceOption(product, "every 12 months", product.frequency.numberOfMonths)
    }
</fieldset>
```
This would suffer from the same issue - if there is no product with a frequency of ```BillingFreqency.month```, the default checked option will not be rendered in the template.

I guess checking outside of the template that we have all expected billing frequencies would be a better option? And throw an error if not?

app/model/exactTarget/SubscriptionDataExtensionRow.scala
joelochlann: Separate PR

app/views/checkout/payment.scala.html
joelochlann: Separate PR

joelochlann:
@ostapneko I've squashed some related commits together so the commit history should be tidier now

ostapneko:
:+1: for the backend code, minus the comments

app/views/checkout/payment.scala.html
joelochlann: As per our discussion, I will look at improving this as part of a PR to dynamically handle payment dates

app/controllers/Checkout.scala
joelochlann: Yes, that's a good point. I will be doing separate PRs for improving this error handling in membership-common, membership-frontend and subscriptions-frontend.

joelochlann:
The signed-in user flow is broken on prod - email validation fails. Fixing now

joelochlann:
Here's the [PR for the fix](https://github.com/guardian/subscriptions-frontend/pull/274). Not yet sure if it was actually these changes which broke things, will investigate

joelochlann:
Wait, now I'm really confused. Having restarted `sbt` the variable is now named `guardian.user.isSignedIn` which is what the frontend expects. No idea what's going on here

joelochlann:
Ah, that confusion was just branch-changing synchronisation issues.... the breaking change was [one I made right at the end in response to @ostapneko's comment](https://github.com/guardian/subscriptions-frontend/commit/9b8b4aa07b7e4490b3bd54666c7fc7fddca2960d)

joelochlann:
Now working on PROD, fixed [by this PR](https://github.com/guardian/subscriptions-frontend/pull/275)

PR #274, opened 2015-11-16
Fix incorrect variable name so we don't validate email when signed in

joelochlann:
This is broken on PROD because it's checking a variable named `guardian.user.isSignedIn` but the variable is called `guardian.user.userIsSignedIn` for some reason. This actually breaks the checkout page because the user cannot get past the Personal Details section - it tries to validate the email but can't find it (because a necessary class is omitted from the template for signed-in users), so passes `null` which always fails validation.

No idea what broke this, will investigate after.

@chrisjowen


chrisjowen:
:+1:

joelochlann:
I've found the root cause of this - it was [this commit](https://github.com/guardian/subscriptions-frontend/commit/9b8b4aa07b7e4490b3bd54666c7fc7fddca2960d), so I'm going to revert that instead.

PR #275, opened 2015-11-16
Revert "Simplify JSON serialisation"

joelochlann:
This tidy-up actually broke the checkout flow for signed-in users. The frontend expects the variable to be named `guardian.user.isSignedIn` but because we do a straight serialisation it's now called `guardian.user.userIsSignedIn`. This means user cannot get past the Personal Details section - it thinks the user is not signed in, so it tries to validate the email but can't find it (because a necessary class is omitted from the template for signed-in users), so passes `null` which always fails validation.

Reverting so we can get the checkout flow on PROD working again!

@ostapneko

### DIFF
--- a/app/model/JsVars.scala
+++ b/app/model/JsVars.scala
@@ -3,18 +3,33 @@ package model
 import play.api.libs.json._
 
 object JsVars {
-  def default = JsVars()
+  def default = JsVars(
+    userIsSignedIn = false,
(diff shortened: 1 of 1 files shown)


ostapneko:
:+1:

joelochlann:
Verified on PROD, this has fixed the issue. In another PR I'll re-instate the simplification but ensure variables match up with what the frontend expects

PR #276, opened 2015-11-20
URI-encode the email address when we check if it's in use

joelochlann:
Currently when we make an AJAX request to check whether the supplied email address is in use, we don't URI encode the email. This means that, for example, a plus sign in an email address would be seen as a space by the backend, meaning that in-use emails would be treated as available. This breaks the final checkout step, which assumes the email is not taken when it tries to create a guest identity account.


tomverran:
:+1:

PR #277, opened 2015-11-20
Log actual ExactTarget response

joelochlann:
We're currently logging an object by mistake - we should be logging the actual response body


tomverran:
looks good assuming CI is happy :+1:

PR #278, opened 2015-11-23
Push stripe info to salesforce when subscribing by cc

tomverran:
@joelochlann and @ostapneko noticed this when testing subs MMA

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -72,15 +72,15 @@ class CheckoutService(
       }
       subscribe <- subscribeAction(payment)
       result <- zuoraService.subscribe(subscribe)
+      postSubscribe = subscribe.paymentMethodOpt match {
+        case Some(CreditCardReferenceTransaction(customer)) => salesforceService.setCardInformation(userData.id, customer.id, customer.card.id)
+        case _ => Future(memberId)
+      }
+      postSubscribeResult <- postSubscribe
     } yield {
       updateAuthenticatedUserDetails()
       sendETDataExtensionRow(result)
-      subscribe.paymentMethodOpt match {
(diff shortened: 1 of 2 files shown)


app/services/CheckoutService.scala
```diff
@@ -75,7 +75,12 @@ class CheckoutService(
     } yield {
       updateAuthenticatedUserDetails()
       sendETDataExtensionRow(result)
+      subscribe.paymentMethodOpt match {
```
joelochlann: Maybe this should go inside the for comprehension, so that the subscribe action fails if this request fails? In membership-frontend that's how it works: https://github.com/guardian/membership-frontend/blob/master/frontend/app/services/MemberService.scala#L118-L119

app/services/CheckoutService.scala
```diff
@@ -75,7 +75,12 @@ class CheckoutService(
...
+        case Some(CreditCardReferenceTransaction(customer)) =>
```
joelochlann: I think it's also worth asking whether we should be pushing the Direct Debit details into Salesforce as well

joelochlann:
:+1: Though would like a second opinion about whether Direct Debit details should go into Salesforce

joelochlann:
Actually @tomverran hold off on merging this for the time being - we've got some weirdness with the Salesforce integration when signing up to Membership and Subs with the same identity account. Salesforce seems to push a second lot of Stripe details into both Zuora billing accounts when the second sign-up happens. Not sure what effect (if any) this change might have on that behaviour (let me know if you do!) but it might complicate our attempts to figure out what's currently going on.

tomverran:
So it is now actually quite important that subs _doesn't_ push to salesforce
hence lets close this

PR #280, opened 2015-11-23
Only display payment notices once you get to the payment details section

joelochlann:
The payment notices on the right-hand side should not be visible when the user is on the Personal Details section.

picture 53

then:

picture 54

then:

picture 55


tomverran:
:+1: looks ok to me

PR #281, opened 2015-11-23
Refresh ExactTarget access token more frequently

joelochlann:
Occasionally we get 401 Unauthorized errors back from ExactTarget which is most likely due to an expired token: https://code.exacttarget.com/apis-sdks/rest-api/using-the-api-key-to-authenticate-api-calls.html

Reducing the delay between token refreshes to see if this helps.

@ostapneko


ostapneko:
:+1:

PR #282, opened 2015-11-24
Don't doubly hyphenate a sort code

joelochlann:
We end up passing an already-hyphenated sortcode (e.g. `20-00-00`) to this function before sending the sortcode to ExactTarget. This results in `20--0-0--0` appearing in the email. So I opted to filter out the non-digits before hyphenating.


chrisjowen:
:+1:

PR #286, opened 2015-11-27
Update to latest membership-common

rtyley:
I just want to make a small change to membership-common and use it, but given the gap in version numbers, I'm updating to latest before starting...

cc @tomverran @ostapneko


app/touchpoint/TouchpointBackendConfig.scala
```diff
@@ -11,7 +11,8 @@ import org.joda.time.Period
 case class TouchpointBackendConfig(
```
rtyley: A lot of this class is becoming a duplicate of what's in the TouchpointBackendConfig in common, maybe a good chunk of it can be removed soon.

tomverran:
looks good to me :+1:

PR #287, opened 2015-11-27
Attempt to take payment straightaway on UAT

joelochlann:
There are certain Stripe payment errors that we never see (and therefore can't test) in the frontend if Zuora doesn't attempt to take payment straightaway. Given that we apparently have the scheduled payments turned off in UAT Zuora anyway, I don't think the payment delay in UAT has much value so I propose we remove it.

@rtyley Do you agree?

### DIFF
--- a/conf/touchpoint.UAT.conf
+++ b/conf/touchpoint.UAT.conf
@@ -14,7 +14,7 @@ touchpoint.backend.environments {
             }
         }
         zuora {
-            paymentDelayInDays = 1
+            paymentDelayInDays = 0
             api {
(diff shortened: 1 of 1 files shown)


joelochlann:
cc @mario-galic

joelochlann:
@rtyley has rightly pointed out that this causes PROD and UAT to diverge unnecessarily so maybe not such a good idea. It's actually unlikely that the trial period will be removed in PROD and therefore unlikely that we will ever need to handle these card decline cases in the UI. Closing this

PR #289, opened 2015-12-02
Add timing to Zuora ProductsCache code

rtyley:
This seems like maybe it's very slow sometimes, so going to monitor it's time.

Also note the ProductsCache currently uses the SOAP API - using the REST API like Membership might be quicker.

cc @ostapneko

### DIFF
--- a/app/services/ProductsCache.scala
+++ b/app/services/ProductsCache.scala
@@ -1,17 +1,19 @@
 package services
 
 import akka.actor.ActorSystem
-import com.gu.membership.util.FutureSupplier
+import com.gu.membership.util.{FutureSupplier, Timing}
 import com.gu.membership.zuora.soap
+import com.gu.membership.zuora.soap.Readers._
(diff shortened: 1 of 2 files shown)


ostapneko:
:+1:

ostapneko:
Also, I think that's one of the reasons the value in an agent should always be immutable and readily available.

["The state of an Agent should be immutable"](http://doc.akka.io/docs/akka/snapshot/scala/agents.html)

rtyley:
> "The state of an Agent should be immutable"

Ha! Fair play. I guess one thing I'm trying to avoid is the difficult-to-read ClassLoader errors that can occur when a static initialisation error occurs.

rtyley:
Waiting for the metric to show up here: https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metrics:metricFilter=Pattern%253Dsubscriptions-frontend

PR #290, opened 2015-12-03
Show two payment options to all users

joelochlann:
**DO NOT MERGE UNTIL WE'RE READY TO GO LIVE WITH CARD PAYMENT**

This exposes the two payment options (direct debit vs card) to all users. Previously only test users had this choice. Normal users just saw direct debit as before.

@tomverran


tomverran:
looks good to me when snap CI is happy and we're ready to go live :+1:

PR #292, opened 2015-12-04
Fix bug in acceptance test

mario-galic:
@joelochlann 

Made sure a single test-user is used throughout a single test.


joelochlann:
:+1:

PR #293, opened 2015-12-07
Replace Chrome webdriver with Firefox

mario-galic:
@joelochlann

### DIFF
--- a/test/acceptance/pages/Register.scala
+++ b/test/acceptance/pages/Register.scala
@@ -6,7 +6,7 @@ import Config.profileUrl
 import org.scalatest.selenium.{Page, WebBrowser}
 
 class Register(testUser: TestUser) extends Page with WebBrowser with Util {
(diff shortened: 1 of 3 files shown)


test/acceptance/pages/Register.scala
```diff
@@ -6,7 +6,7 @@ import Config.profileUrl
 import org.scalatest.selenium.{Page, WebBrowser}
 
 class Register(testUser: TestUser) extends Page with WebBrowser with Util {
-  val url = s"${profileUrl}/register?skipConfirmation=true&returnUrl=${baseUrl}/checkout"
+  val url = s"""${profileUrl}/register?returnUrl=${baseUrl}/checkout&skipConfirmation=true"""
```
joelochlann: what does `s"""` do versus `s"`?

joelochlann: I imagine the problem here is to do with whether the `skipConfirmation` param gets associated with the link URL, or the `returnUrl` parameter. Not sure what the exact URI encoding rules are here but worth digging into

PR #295, opened 2015-12-08
Send US subscribers to the US supporter page

tomverran:


app/model/DigitalEdition.scala
```diff
@@ -9,10 +9,13 @@ object DigitalEdition {
...
-  def getRedirect(edition: DigitalEdition): String = {
-    edition match {
+  def getRedirect(edition: DigitalEdition): String = edition match {
       case UK => "/digital/country"
       case _ => "https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=dga38&CMP=FAB_3062"
-    }
+  }
+
+  def getMembershipLandingPage(edition: DigitalEdition): String =  edition match {
+    case US => "https://membership.theguardian.com/us/supporter"
```
afiore: Shouldn't we use a `java.net.URL`?

tomverran:
So I decided to use the netaporter library as working with plain Java URIs isn't too much fun

afiore:
:+1:

PR #307, opened 2016-01-06
Compatibility with membership-common 0.123

ostapneko:
This leaves the website in a quirky state where it should compile and work, but without really using the elements that were adding to mem-common 0.123.

This will happen while working on the multi-currency feature

### DIFF
--- a/app/forms/SubscriptionsForm.scala
+++ b/app/forms/SubscriptionsForm.scala
@@ -2,6 +2,7 @@ package forms
 
 import com.gu.i18n.Country
 import com.gu.memsub.Address
+import com.gu.memsub.Subscription.ProductRatePlanId
 import model._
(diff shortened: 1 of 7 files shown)


app/forms/SubscriptionsForm.scala
```diff
@@ -54,6 +55,8 @@ object SubscriptionsForm {
     "address" -> addressDataMapping
   )(PersonalData.apply)(PersonalData.unapply)
 
+  val productRatePlanIdMapping = mapping("ratePlanId" -> text)(ProductRatePlanId.apply)(ProductRatePlanId.unapply)
```
tomverran: so I hope this was the right thing to do here, I felt converting the string to a ProductRatePlanId should happen as early as possible

tomverran:
:+1:

PR #308, opened 2016-01-07
Stop Identity cookie expiration being reported to Sentry

rtyley:
See https://github.com/guardian/identity-play-auth/pull/6


afiore:
:+1:

PR #312, opened 2016-01-08
Log 'X-Forwarded-For' headers to help with diagnosing recorded IPs

rtyley:
[Ed Fox](http://spike/content/3433) has noticed that we're actually recording _Fastly's_ IP address for subscribers, rather than the IP of the actual user.

https://trello.com/c/I94f2NAl/284-custom-data-fields-for-2-0-subscriber-ip-address

Fastly is already set up to add the relevant forwarded headers...

https://docs.fastly.com/guides/basic-configuration/manipulating-the-x-forwarded-for-header
https://app.fastly.com/#configure/service/6CcQKlAJNA1HitqjWn4c10/version/23/settings
image

...so I'm not sure why this isn't working, adding this debug should help.

cc @tomverran

### DIFF
--- a/app/controllers/Testing.scala
+++ b/app/controllers/Testing.scala
@@ -3,7 +3,7 @@ package controllers
 import actions.CommonActions._
 import com.typesafe.scalalogging.LazyLogging
 import controllers._
-import play.api.mvc.{Controller, Cookie}
+import play.api.mvc.{Headers, Controller, Cookie}
 import utils.TestUsers.testUsers
 
 object Testing extends Controller with LazyLogging {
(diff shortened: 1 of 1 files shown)


rtyley:
The logged data looks like this (for me, hitting https://subscribe.theguardian.com/test-users from within the Guardian):

```
2016-01-08 17:26:04,328 [application-akka.actor.default-dispatcher-345] INFO  controllers.Testing$ - remoteAddress=185.31.18.24 X-Forwarded-For=77.91.250.233, 77.91.250.233, 185.31.18.24 Forwarded=
```

So, no `Forwarded` headers at all, but with `X-Forwarded-For=77.91.250.233, 77.91.250.233, 185.31.18.24` :

*  [`77.91.250.233`](https://myip.ms/info/whois/77.91.250.233) is the Guardian
*  [`185.31.18.24`](https://myip.ms/info/whois/185.31.18.24) is Fastly

...so Guardian is first, then Fastly - but Play's `remoteAddress` is giving the Fastly address. Why...?

The answer must be in https://github.com/playframework/playframework/blob/2.4.6/framework/src/play-server/src/main/scala/play/core/server/common/ForwardedHeaderHandler.scala

rtyley:
Oh dear. Play (fairly reasonably) returns the first ip address that isn't on a whitelist... and we don't have Fastly's IP addresses trusted, at least yet...

https://docs.fastly.com/guides/securing-communications/accessing-fastlys-ip-ranges
https://api.fastly.com/public-ip-list

PR #313, opened 2016-01-13
Remove recover combinator from ET service

mario-galic:
@joelochlann @afiore 

The recover combinator does not seem to add more value, as the error is already sufficiently logged at line [50](https://github.com/guardian/subscriptions-frontend/compare/fix-superfluous-recover-combinator?expand=1#diff-4b43727a795be0d3aaf5e0d734bb7f44R50).

[Sentry events](https://app.getsentry.com/the-guardian/subscriptions/issues/106183481/events/)

### DIFF
--- a/app/services/ExactTargetService.scala
+++ b/app/services/ExactTargetService.scala
@@ -29,7 +29,7 @@ trait ExactTargetService extends LazyLogging {
       pm <- zuoraService.getDefaultPaymentMethod(acc)
     } yield (acc, pm)
 
-    (for {
+    for {
       subs <- subscription
       rpc <- zuoraService.recurringRatePlanCharge(subs)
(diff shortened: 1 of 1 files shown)


tomverran:
:+1:

app/services/ExactTargetService.scala
```diff
@@ -50,8 +50,6 @@ trait ExactTargetService extends LazyLogging {
           logger.error(errorMsg)
           throw new ExactTargetException(errorMsg)
```
joelochlann: Should you not remove the thrown exception if you've removed the `recover` block? Otherwise you'll change the behaviour - this error will bubble up and possibly cause a 500 error

mario-galic: Fixed.

joelochlann:
:+1:

PR #315, opened 2016-01-13
Multi currency

ostapneko:
~~This is waiting for https://github.com/guardian/membership-common/pull/151~~

This PR implements multi-currency:

  - It show different fields and address validation based on the particular country group:
image
image
image
image
image

  - It switches currency based on the same rules as membership
image
image

  - It prevents international users from using Direct Debit (both front end and backend)
image
image

  - It relies on the identity user's country or else the `countryGroup` query string parameter to determine the default country, similar to membership

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -60,7 +60,7 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
     subscriptionData map { subsData =>
       val form = subsData.fold(SubscriptionsForm()) { data => SubscriptionsForm().fill(data) }
       val countryGroupWithDefault =
-        subsData.flatMap { data => CountryGroup.byCountryCode(data.personalData.address.countryCode) }
+        subsData.flatMap { data => CountryGroup.byCountryNameOrCode(data.personalData.address.countryName) }
                 .getOrElse(countryGroup)
       val desiredCurrency = countryGroupWithDefault.currency
       val supportedCurrencies = defaultPlan.pricing.underlying.keySet
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -63,7 +63,7 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
         subsData.flatMap { data => CountryGroup.byCountryNameOrCode(data.personalData.address.countryName) }
                 .getOrElse(countryGroup)
       val desiredCurrency = countryGroupWithDefault.currency
-      val supportedCurrencies = defaultPlan.pricing.underlying.keySet
+      val supportedCurrencies = defaultPlan.currencies
       val currency = if (supportedCurrencies.contains(desiredCurrency)) desiredCurrency else GBP
 
       Ok(views.html.checkout.payment(form = form,
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -2,6 +2,7 @@ package controllers
 
 import actions.CommonActions._
 import com.gu.identity.play.ProxiedIP
+import com.gu.i18n.{GBP, CountryGroup}
 import com.gu.memsub.Subscription.ProductRatePlanId
 import com.gu.stripe.Stripe
 import com.gu.zuora.soap
@@ -18,6 +19,7 @@ import tracking.ActivityTracking
 import tracking.activities.{CheckoutReachedActivity, MemberData, SubscriptionRegistrationActivity}
 import utils.TestUsers.{NameEnteredInForm, PreSigninTestCookie}
 import views.html.{checkout => view}
+import views.support.CountryWithCurrency
 
 import scala.Function.const
 import scala.concurrent.ExecutionContext.Implicits.global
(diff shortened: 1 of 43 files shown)


app/model/SubscriptionData.scala
```diff
@@ -54,8 +74,8 @@ object SubscriptionData {
       lineTwo = u.privateFields.getOrBlank(_.billingAddress2),
       town = u.privateFields.getOrBlank(_.billingAddress3),
       postCode = u.privateFields.getOrBlank(_.billingPostcode),
-      countyOrState = u.privateFields.getOrBlank(_.country),
```
ostapneko: Note that we were sending **country** to **county/state**. I don't know if there might be implication, but a couple of people might now see that they live in the county of "United Kingdom"

app/views/checkout/thankyou.scala.html
```diff
@@ -41,7 +44,7 @@
                 </p>
                 <p>Open times: Lines are open Monday&ndash;Friday 8.00am&ndash;5.30pm. Saturday &amp; Sunday 8.30am&ndash;12.30pm</p>
```
ostapneko: This should be changed for international customers

ostapneko:
TODO: The paper subscription should not be accessible to the international users.

app/controllers/Checkout.scala
```diff
@@ -32,33 +36,43 @@ object Checkout extends Controller with LazyLogging with ActivityTracking {
...
-          SubscriptionsForm()
-        }
-      }
-    for {
-      filledForm <- fillForm()
-      products <- zuoraService.products
-      selectedProduct = products.find(p => p.frequency == BillingFrequency.Month).get
-    } yield {
-      Ok(views.html.checkout.payment(filledForm, userIsSignedIn = authUserOpt.isDefined, products, selectedProduct, touchpointBackend))
+    trackAnon(CheckoutReachedActivity("United Kingdom"))
```
tomverran: is this hard coding correct?

ostapneko: Good catch!

app/forms/SubscriptionsForm.scala
```diff
@@ -29,14 +29,24 @@ object SubscriptionsForm {
   private val addressMaxLength = 255
   private val emailMaxLength = 240
 
-  def unapplyUkAddress(a: Address) = Some((a.lineOne, a.lineTwo, a.town, a.postCode))
-  def applyUkAddress(lineOne: String, lineTwo: String, town: String, postCode: String) = Address(lineOne, lineTwo, town, "", postCode, Country.UK)
+  // Unfortunately, the UI wants a country code, but Identity has a country name,
```
tomverran: this comment is a trifle misleading isn't it? Since identity could have either

ostapneko:
The county/state should be above postcode

app/views/checkout/thankyou.scala.html
afiore: Maybe we can just avoid switching and have two paragraphs here, one for UK the other for the rest of the word

ostapneko:
The `countryGroup` QS needs to appear on the us subscribe link

app/controllers/Checkout.scala
```diff
@@ -32,33 +36,43 @@ object Checkout extends Controller with LazyLogging with ActivityTracking {
...
+    val plans = catalog.planMap.values.toSeq
+    val defaultPlan = catalog.digipackMonthly
+
+    subscriptionData map { subsData =>
+      val form = subsData.fold(SubscriptionsForm()) { data => SubscriptionsForm().fill(data) }
+      val countryGroupWithDefault =
+        subsData.flatMap { data => CountryGroup.byCountryNameOrCode(data.personalData.address.countryName) }
+                .getOrElse(countryGroup)
+      val desiredCurrency = countryGroupWithDefault.currency
+      val supportedCurrencies = defaultPlan.pricing.underlying.keySet
```
afiore: you could replace that with just `defaultPlan.currencies`

app/controllers/Checkout.scala
```diff
@@ -32,33 +36,43 @@ object Checkout extends Controller with LazyLogging with ActivityTracking {
...
+      val currency = if (supportedCurrencies.contains(desiredCurrency)) desiredCurrency else GBP
+
```
afiore: Just noticing that in both membership and here this whitelisting logic is repeated twice between form rendering and form processing. We should push it into commons in a later PR

app/views/checkout/thankyou.scala.html
ostapneko: Instead, I just changed the paragraph content slightly

app/controllers/Checkout.scala
ostapneko: Agreed

app/views/support/CountryWithCurrency.scala
```diff
@@ -0,0 +1,24 @@
+package views.support
+
```
afiore: This should go in membership-common

app/model/AddressValidation.scala
```diff
@@ -0,0 +1,18 @@
+package model
+
+import com.gu.memsub.Address
+
+object AddressValidation {
+  def validateForCountry(address: Address) = {
```
afiore: Could do with a test1

afiore:
:+1:

app/model/SubscriptionData.scala
```diff
@@ -41,28 +61,29 @@ case class SubscriptionData(personalData: PersonalData, paymentData: PaymentData
...
     }
 
     val addressData = Address(
-      lineOne = u.privateFields.getOrBlank(_.billingAddress1),
-      lineTwo = u.privateFields.getOrBlank(_.billingAddress2),
-      town = u.privateFields.getOrBlank(_.billingAddress3),
-      postCode = u.privateFields.getOrBlank(_.billingPostcode),
-      countyOrState = u.privateFields.getOrBlank(_.country),
-      country = Country.UK
+      lineOne = u.privateFields.getOrBlank(_.billingAddress1, _.address1),
```
afiore: Turns out this logic is wrong as it would potentially cause to interleaf fields from billing and delivery address, while we should build  `AddressData` from either one or the other.

ostapneko:
seen on prod:

image

PR #317, opened 2016-01-15
Use country rather than currency to hide/show direct debit

ostapneko:
The Direct Debit form should only be shown to UK resident. The deciding element to show or hide it is currently the currency. However the countries which currencies are not defined in Zuora default to GBP.
This PR switches on the country instead.

cc @tomverran @afiore


afiore:
:+1:

PR #318, opened 2016-01-15
Removal of A/B checkout flow test and country restrictions on new checkout

AWare:
Country.js wasn't really used for anything. Although it was initialising on every page. And sometimes changing the pages name.


AWare:
There were some issues with the international footer being shown/not being shown. Have refactored this so that digitalEdition is now aware of CountryGroup- and then using this to find phone support details (or lack of) 

@tomverran @afiore

app/model/DigitalEdition.scala
```diff
@@ -1,20 +1,24 @@
...
+  def id = countryGroup.id
+  def name = countryGroup.name
+}
 
 object DigitalEdition {
 
-  object UK extends DigitalEdition("uk", "UK", "£11.99", "dis_2408")
-  object US extends DigitalEdition("us", "US", "$19.99", "dis_2378")
-  object AU extends DigitalEdition("au", "Australia", "$21.50", "dis_2379")
+  object UK extends DigitalEdition("£11.99", "dis_2408", CountryGroup.UK)
```
afiore: Here we are still hardcoding the prices while they have been recently configured in Zuora prod. This is outside of the scope of this PR, but we should do it right after we merge this

app/model/SupportPhoneDetails.scala
```diff
@@ -0,0 +1,11 @@
+package model
```
afiore: I would start moving this kind of code in the `view` package (perhaps in `view.model`). This helps conveying the fact that it is only a presentation concern.

app/views/checkout/thankyou.scala.html
```diff
@@ -14,7 +15,14 @@
 @import services.IdentityToken.{paramName => identityTokenParam}
 @import services.UserId.{paramName => userIdParam}
 
-@main("Confirmation | Digital | Subscriptions | The Guardian", bodyClasses=List("is-wide"), touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
+@main(
+    "Confirmation | Digital | Subscriptions | The Guardian",
+    bodyClasses=List("is-wide"),
+    touchpointBackendResolutionOpt = Some(touchpointBackendResolution),
+    countryGroup = CountryGroup.UK
+    //TODO: FIXME
```
afiore: Shouldn't we wire this up?

app/views/support/CountryGroupOps.scala
```diff
@@ -0,0 +1,16 @@
+package views.support
+
+import com.gu.i18n.CountryGroup
+import model.SupportPhoneDetails
+
+/**
```
afiore: Annoyingly IntelliJ adds this by default (It is easy to disable..). Can we please lose the comment ;)?

app/model/DigitalEdition.scala
AWare: Yeah, @tomverran and I talked about this, but decided that it was probably outside scope.

app/views/checkout/thankyou.scala.html
AWare: 😱 my bad.

AWare:
Superseded by (pull request to follow)

PR #322, opened 2016-01-19
Make postcode always mandatory

afiore:
The creation of a guest user in Identity requires us to supply a postcode; so we are making postcode mandatory, even if in some countries this is not ideal

### DIFF
--- a/assets/javascripts/modules/checkout/fieldSwitcher.js
+++ b/assets/javascripts/modules/checkout/fieldSwitcher.js
@@ -33,9 +33,7 @@ define([
     };
 
     var redrawAddressFields = function(model) {
-        var newPostcode = addressFields.postcode(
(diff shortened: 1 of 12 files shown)


afiore:
@ostapneko

assets/javascripts/modules/checkout/fieldSwitcher.js
```diff
@@ -33,8 +33,7 @@ define([
     };
 
     var redrawAddressFields = function(model) {
-        var newPostcode = addressFields.postcode(
-            model.postcodeRules.required,
+        var newPostcode = addressFields.postcode(true,
```
ostapneko: Why do we keep the first argument?

ostapneko:
:+1:

PR #324, opened 2016-01-20
Postcode optional again

afiore:
- Revert mandatory postcode
- Send a bogus value to Identity when user-supplied postcode is blank

### DIFF
--- a/app/services/IdentityService.scala
+++ b/app/services/IdentityService.scala
@@ -183,9 +183,18 @@ object IdentityApiClient extends IdentityApiClient with LazyLogging {
   override val createGuest: PersonalData => Future[WSResponse] = {
     val endpoint = authoriseCall(WS.url(s"$identityEndpoint/guest"))
 
-    guestData => endpoint.post(guestData: JsObject)
-      .withWSFailureLogging(endpoint)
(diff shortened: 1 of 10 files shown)


afiore:
@ostapneko

app/services/IdentityService.scala
```diff
@@ -183,9 +183,18 @@ object IdentityApiClient extends IdentityApiClient with LazyLogging {
...
     val endpoint = authoriseCall(WS.url(s"$identityEndpoint/guest"))
 
-    guestData => endpoint.post(guestData: JsObject)
-      .withWSFailureLogging(endpoint)
-      .withCloudwatchMonitoringOfPost
+    guestData => {
+      // When postcode is blank (e.g. Ireland), force-feed identity with a bogus postcode
+      val idSafeGuestData =
+        if (guestData.address.postCode.isEmpty)
+          guestData.copy(address = guestData.address.copy(postCode = "---"))
```
ostapneko: We also need to change the way this information is read from identity

ostapneko:
:+1:

app/services/IdentityService.scala
afiore: The user never returns to the checkout page, so I haven't done that

PR #327, opened 2016-01-21
Styling tweaks to "Review and confirm"

tomverran:
- Switch input for button so the ::after arrow shows up
- Add u-note to the direct debit information
 - Add prose class so links look okay

see trello 286

@joelochlann 

### mobile

image

### tablet

image

### desktop

image

### DIFF
--- a/assets/stylesheets/modules/_checkout.scss
+++ b/assets/stylesheets/modules/_checkout.scss
@@ -40,6 +40,9 @@
         padding-left: gs-span(1.5);
     }
 }
+.checkout-container__frequency {
+    cursor: pointer;
+}
 
 /* Country Select
    ========================================================================== */
@@ -209,6 +212,7 @@ $sprite-card-offset-visa:               -355px;
         margin-left: 35%;
     }
 }
+
 
 /* ==========================================================================
    Review Panel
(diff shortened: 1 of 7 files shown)


assets/stylesheets/modules/_checkout.scss
```diff
@@ -209,6 +209,11 @@ $sprite-card-offset-visa:               -355px;
         margin-left: 35%;
     }
 }
+.review-details__notes {
+    @include fs-textSans(1);
+    line-height: 1.125rem;
```
joelochlann: Use pixels since we have a px-to-rem postprocessor in our Grunt compile task which will turn these into rem

assets/stylesheets/modules/_checkout.scss
```diff
@@ -209,6 +209,11 @@ $sprite-card-offset-visa:               -355px;
         margin-left: 35%;
     }
 }
+.review-details__notes {
```
joelochlann: Looking at Ben's designs, he wants the same styling on the text on the right-hand side, so scoping this to a child of review-details might not be the best decision. You may want to deal with that in the PR for the right-hand side text though

joelochlann:
:+1: Looks good

PR #329, opened 2016-01-21
Fix Firefox analytics

paulbrown1982:
Fixing bug where analytics did not work on Firefox, because navigator.doNotTrack === "unspecified" so prefixing ! returned true rather than false. Chrome returns 'null' for that value. Also fixed that IE and Safari use window.doNotTrack rather than navigator.

### DIFF
--- a/assets/javascripts/modules/analytics/analyticsEnabled.js
+++ b/assets/javascripts/modules/analytics/analyticsEnabled.js
@@ -3,9 +3,19 @@ define([
 ], function (cookie) {
     'use strict';
 
+    /*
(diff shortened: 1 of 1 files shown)


rtyley:
:+1: we simplify the DNT-checking logic here a bit, compared to http://stackoverflow.com/a/16947583/438886 - we're basically saying that if you want DoNotTrack to work for you, your browser needs to exhibit behaviour respecting what passes for the latest version of the spec : https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack

PR #330, opened 2016-01-21
Add a page to query CAS subscriptions

ostapneko:
This related to [this Trello card](https://trello.com/c/qCoJN5Ri/287-provide-userhelp-with-a-subscription-checker-tool-that-works-with-zuora-as-well-as-legacy-cas)

image

@rtyley @afiore


assets/javascripts/modules/cas/form.js
```diff
@@ -0,0 +1,65 @@
...
+
+    function submit() {
+        var data = serializer([].slice.call($FORM[0].elements));
+        console.log('submitting data', data);
+        ajax({
+            url: $FORM[0].action,
+            method: 'post',
+            data: data,
+            success: function(successData) {
+                console.log(successData);
```
tomverran: console.log!

app/forms/CASForm.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import model.CASLookup
+import play.api.data.Forms._
+import play.api.data._
+
+object CASForm {
+  def apply(): Form[CASLookup] = Form(
+    "cas" -> mapping(
+      "number" -> nonEmptyText,
+      "lastName" -> nonEmptyText,
```
rtyley: I'm not sure that this should be required - usually we only require that _either_ the `postCode`, or failing that, the `lastName` is supplied.

assets/javascripts/modules/cas/form.js
ostapneko: I want to keep it, it's only for staff and it makes things easier to debug if they have a problem

app/forms/CASForm.scala
ostapneko: I know but really, I wanted to stick the the already existing CASService interface, which is a bit unfortunate:

https://github.com/guardian/membership-common/blob/6169b7460568104eb1940c63a67f0ea30cb56b61/src/main/scala/com/gu/subscriptions/CASApi.scala#L43

I could try to rewrite that, but is it worth it?

app/controllers/CAS.scala
```diff
@@ -0,0 +1,24 @@
...
+import play.api.mvc.Controller
+import views.support.CASResultOps._
+import play.api.libs.concurrent.Execution.Implicits._
+
+object CAS extends Controller {
+  def index = GoogleAuthenticatedStaffAction { implicit request =>
+    Ok(views.html.staff.cas())
+  }
+
+  def search = GoogleAuthenticatedStaffAction.async(parse.form(CASForm())) { request =>
```
rtyley: `GoogleAuthenticatedStaffAction` is good - tho' I'm surprised to notice that, [unlike Membership](https://github.com/guardian/membership-frontend/blob/1fe86747a6/frontend/app/controllers/Testing.scala#L18-L28), we don't have any code in place to restrict access based on Google-Group membership - ideally, we should really be restricting use of this tool to `userhelp@` and ourselves.

PR #331, opened 2016-01-21
Add an international landing page

tomverran:
The international landing page is the same as the US / AU ones but has no savings mentioned.

@joelochlann @ostapneko 312 on trello

image

### DIFF
--- a/app/views/support/DigitalEdition.scala
+++ b/app/views/support/DigitalEdition.scala
@@ -17,7 +17,7 @@ object DigitalEdition {
       val params = "INTCMP" -> s"GU_SUBSCRIPTIONS_${edition.id.toUpperCase}_PROMO"
 
       edition match {
-        case INTL => "https://membership.theguardian.com/int/supporter" ? params
+        case INT => "https://membership.theguardian.com/int/supporter" ? params
         case US => "https://membership.theguardian.com/us/supporter" ? params
         case _ => "https://membership.theguardian.com/join" ? params
       }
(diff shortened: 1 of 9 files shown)


app/views/support/DigitalEdition.scala
```diff
@@ -0,0 +1,38 @@
...
+    def redirect: Uri = edition match {
+      case UK => "/digital/country"
+      case _ => "https://www.guardiansubscriptions.co.uk/digitalsubscriptions/?prom=dga38&CMP=FAB_3062"
+    }
+
+    def membershipLandingPage: Uri = {
+      val params = "INTCMP" -> s"GU_SUBSCRIPTIONS_${edition.id.toUpperCase}_PROMO"
+
+      edition match {
+        case INTL => "https://membership.theguardian.com/int/supporter" ? params
```
ostapneko: Could we change that to INT so it's consistent with the URLs

conf/routes
```diff
@@ -22,6 +23,7 @@ GET         /au/digital/country              controllers.DigitalPack.selectAu
 GET         /digital                         controllers.DigitalPack.uk
 GET         /us/digital                      controllers.DigitalPack.us
 GET         /au/digital                      controllers.DigitalPack.au
+GET         /int/digital                     controllers.Default.redirect(to = "/us/digital")
```
ostapneko: We shouldn't need that, International is a CG with a currency

conf/routes
```diff
@@ -9,6 +9,7 @@ GET         /favicon.ico                     controllers.CacheBustedAssets.at(pa
 GET         /                                controllers.Homepage.index
 GET         /au                              controllers.Homepage.indexAu
 GET         /us                              controllers.Homepage.indexUs
+GET         /int                             controllers.Homepage.indexIntl
```
ostapneko: s/intl/int, see above

PR #335, opened 2016-01-25
Add endpoint for promo-code validation

afiore:


app/model/bla.sc
```diff
@@ -0,0 +1,3 @@
+import model.SubscriptionData
```
tomverran: hmm

tomverran:
:+1:

PR #336, opened 2016-01-25
Thank you page & footer copy amends

blackyjnz:
Updating
- the thank you page with international country code so we can
introduce intl card payments
- the footer with the same


ostapneko:
image


image

ostapneko:
:+1: once it builds

PR #337, opened 2016-01-25
Fix the acceptance tests

tomverran:
The submit button is now a `<button>` rather than an `<input type="button">`


mario-galic:
:+1:

PR #339, opened 2016-01-25
Use ScheduledTask from membership-common

mario-galic:
@tomverran @afiore


tomverran:
Looks good, I had some idea there was an import you could use to get an implicit ActorSystem but I seem to be mistaken :+1:

PR #342, opened 2016-01-26
Upgrading to latest membership-common 1.4.5

paulbrown1982:
This is to be sure subscriptions does not contain the 5-minute healthcheck membership-common bug in 1.4.3+.


ostapneko:
:+1:

PR #343, opened 2016-01-26
Fix payment options in Firefox/IE

ostapneko:
Firefox (and IE?) doesn't accept a change event on a disabled input...

@paulbrown1982 @tomverran


tomverran:
:+1:

ostapneko:
Seen on PROD

image

PR #347, opened 2016-01-29
Log ExactTarget token renewal failures

ostapneko:


app/services/ExactTargetService.scala
```diff
@@ -16,6 +16,9 @@ import play.api.libs.json._
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import com.gu.memsub.Subscription
+
+import scala.util.Try
```
tomverran: we don't need this import

tomverran:
:+1:

PR #350, opened 2016-02-03
Promocodes landing page - stub

paulbrown1982:
Added a stub landing page which shows when the promotion is put in the URL.
Also tweaked error message when a promo isn't found, advertising the demo promo.
Also fixed some JS errors.
All of this is behind the /test-users cookie toggle.

/cc @joelochlann 
/cc @tomverran
/cc @rtyley


rtyley:
LGTM :+1:

PR #352, opened 2016-02-05
Turned on SessionCam

paulbrown1982:
Turning on SessionCam for 100% of traffic.
Removed endpoints which used to set or unset the feature toggling 'sessioncam-on' cookie.

/cc @tomverran 
/cc @rtyley


rtyley:
:+1: 

membership version is at https://github.com/guardian/membership-frontend/pull/977

PR #354, opened 2016-02-16
Promotion landing page

paulbrown1982:
Incentive-type promotion landing page, pending code review and cross-browser testing.

/cc @joelochlann @rtyley


assets/stylesheets/modules/_roundel.scss
```diff
@@ -0,0 +1,68 @@
...
+    margin: 10px 10px;
+
+    @include mq(desktop) {
+        width: 140px;
+        height: 140px;
+        padding: 20px 20px;
+        margin: 20px 20px;
+    }
+
+    align-content: center;
```
joelochlann: These should go before the media query - media queries should always override more general rules above them

assets/stylesheets/modules/_roundel.scss
```diff
@@ -0,0 +1,68 @@
...
+    display: flex;
+    justify-content: center;
+    flex-direction: column;
+}
+
+.roundel__title {
+    @include fs-headline(1);
+    font-weight: guss-font-weight(ultrablack);
+    line-height: 1.3rem;
```
joelochlann: this should be in px, we have a px-to-rem pre-processor as part of our Grunt compile process

assets/stylesheets/modules/_roundel.scss
```diff
@@ -0,0 +1,68 @@
...
+
+    @include mq(desktop) {
+        font-size: 13px;
+        @include fs-headline(1);
+    }
+}
+
+.roundel__strong {
+    display: block;
+    font-size: 2rem;
```
joelochlann: no rem

assets/stylesheets/modules/_promotion.scss
```diff
@@ -0,0 +1,97 @@
+.promotion__landing-page {
+    section:nth-child(even) {
+        background-color: rgb(231, 231, 231);
+
+        .page-slice {
+            background-color: rgb(241, 241, 241);
+        }
+    }
+    .page-slice {
+        padding-bottom: 0.75rem;
```
joelochlann: should be px

assets/stylesheets/objects/_grid.scss
```diff
@@ -80,6 +95,19 @@ $gutter-width-fluid: 2%;
     }
 }
 
+.grid--2up-step {
```
joelochlann: Can you add a comment describing what this does and why it's needed?

assets/stylesheets/modules/_promotion.scss
```diff
@@ -0,0 +1,97 @@
+.promotion__landing-page {
```
joelochlann: According to BEM convention, this is a child selector which describes the `landing-page` child of a `promotion` component. But where is the promotion component? Similarly for all other rules here: we have lots of children but no parent.

assets/stylesheets/modules/_roundel.scss
```diff
@@ -0,0 +1,68 @@
...
+        font-weight: guss-font-weight(ultrablack);
+    }
+}
+
+.roundel__description {
+    font-size: 12px;
+
+    @include mq(desktop) {
+        font-size: 13px;
+        @include fs-headline(1);
```
joelochlann: won't this include override the `font-size: 13px` above it?

assets/stylesheets/objects/_grid.scss
```diff
@@ -32,6 +38,15 @@ $gutter-width-fluid: 2%;
     }
 }
 
+.grid__item--flex-2-columns {
+    border: 0 !important;
```
joelochlann: Not very happy with this use of `!important`. The lack of a border has nothing to do with the fact that this is 2-column flex - it's entirely specific to your particular use case. I would add a utility class like `.u-no-border` to override border styles and add this class to the element in question

assets/stylesheets/modules/_promotion.scss
```diff
@@ -0,0 +1,97 @@
...
+.promotion__landing-page__section-head {
+    margin-bottom: $gs-baseline * 2;
+}
+
+.promotion__digipack {
+    float: right;
+    width: auto;
+}
+
+.promotion__digipack-shots {
```
joelochlann: Class names are supposed to reflect the HTML structure. The hyphenated `digipack-shots` is where you should be using BEM child syntax instead.
`.digipack` on parent
`.digipack__shots` on child

assets/stylesheets/modules/_promotion.scss
```diff
@@ -0,0 +1,97 @@
...
+    }
+    @include mq(leftCol) {
+        background-position: top right;
+        margin-right: 30px;
+        min-height: 382px;
+        min-width: 700px;
+    }
+}
+
+.promotion__digipack-benefits {
```
joelochlann: again, these class names should encode the information that `.promotion__digipack-benefit` is a child of `.promotion__digipack-benefits`
`.promo-benefits`
`.promo-benefits__benefit`

joelochlann:
:+1: Looks good! Main things to look at in a future iteration
* Considering using Grid images in `` tags (using `ResponsiveImageGroup` class) instead of background images
* Consider whether we could make the CSS more any more BEM-compliant (perhaps not, given the landing page nature of this)
* Remove cruft from SVGs. Sorry, forgot to mention this one earlier. Here was a comment David Rapson made on one of my first PRs:

> Event though these are minified buy the build step it's probably still worth running these through https://jakearchibald.github.io/svgomg/ just to clean out the cruft and maybe remove fill="#BDBDBD" just to be on the safe side.

paulbrown1982:
Nice, I did not know about svgomg. I'll make sure to make that a task in the "release this" story which we'll play early next sprint.

paulbrown1982:
picture 8

PR #356, opened 2016-02-18
Allow codes that last a full year in the CAS code generator tool

blackyjnz:
Paul Browne (User Help) advised that the old CAS lookup tool allowed us to generate a 52-week token, but the new version only goes up to 13 weeks. This is to extend this to 52 weeks (useful for anyone having trouble getting access to the 6-month Galaxy Gifts offer for the Guardian app or for testing).

@rtyley


rtyley:
LGTM :+1:

blackyjnz:
[selfie-0](https://github.com/thieman/github-selfies/)

PR #357, opened 2016-02-18
Take 2: allow codes that last a full year in the CAS code generator tool

blackyjnz:
Paul Browne (User Help) advised that the old CAS lookup tool allowed us to generate a 52-week token, but the new version only goes up to 13 weeks. This is to extend this to 52 weeks (useful for anyone having trouble getting access to the 6-month Galaxy Gifts offer for the Guardian app or for testing).
@rtyley

### DIFF
--- a/app/views/staff/cas.scala.html
+++ b/app/views/staff/cas.scala.html
@@ -46,7 +46,7 @@
                     <div class="form-field">
                         <label class="label" for="cas.period">Validity period (weeks)</label>
                         <select required name="cas.period" >
-                        @for(w <- Range(1, 14)) {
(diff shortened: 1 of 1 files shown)


rtyley:
This is picking up after https://github.com/guardian/subscriptions-frontend/pull/356 - where we only changed the server-side-validation number, not the value that's used in the form - sad to see it's double-configured :crying_cat_face:

rtyley:
:+1:

blackyjnz:
[selfie-0](https://github.com/thieman/github-selfies/)

blackyjnz:
cas lookup 52 wks

PR #358, opened 2016-02-18
Use play-googleauth library's new Google-Group filtering

rtyley:
Restrict use of the CAS tool to users in certain Google Groups - as discussed with Suzanne, the tool shouldn't be open to all staff at the Guardian.

This fixes https://trello.com/c/boSbvE9P/356-user-authentication-google-groups-for-new-subscription-checker-tool-that-works-with-zuora-as-well-as-legacy-cas

See also https://github.com/guardian/play-googleauth/pull/36

cc @AWare


AWare:
:+1: 

Requisite @rtyley selfie:
[selfie-2](https://github.com/thieman/github-selfies/)

PR #359, opened 2016-02-18
Design and copy tweaks.

paulbrown1982:
Fixed bug with dodgy whitespace in T&Cs. 
Set the expiry date to 31st March.
Updated colours and copy. 
Tweaked CSS to make pictures bigger and more responsive. 
Made the main promotion image an img rather than background.

/cc @joelochlann @rtyley

### DIFF
--- a/app/views/promotion/landingPage_incentive.scala.html
+++ b/app/views/promotion/landingPage_incentive.scala.html
@@ -40,8 +40,8 @@
                             @fragments.common.roundel(promotion.roundelHtml, Seq("roundel--raised"))
                         </div>
                         <div class="grid__item digipack__benefits">
-                            <div class="digipack__benefits--height">
-                                <div class="digipack__benefits--aligned">
+                            <div class="u-vertically-center">
+                                <div class="u-vertically-centered-item">
                                     <p class="digipack__benefit">Experience your newspaper in full, on your tablet with the Guardian & Observer daily edition</p>
                                     <p class="digipack__benefit">Enjoy an advert-free experience on The Guardian's app with access to our daily crosswords and exclusive content</p>
                                     @subscribeNow(edition, catalog.digipackMonthly.priceGBP, promoCode)
--- a/app/views/promotion/landingPage_incentive.scala.html
+++ b/app/views/promotion/landingPage_incentive.scala.html
@@ -1,8 +1,10 @@
+@import com.github.nscala_time.time.Imports.{DateTime, DateTimeZone}
 @import com.gu.memsub.Price
 @import com.gu.memsub.promo.{PromoCode, Promotion}
 @import com.gu.subscriptions.DigipackCatalog
 @import model.DigitalEdition
 @import views.support.DigitalEdition._
+@import views.support.Dates._
 @(edition: DigitalEdition, catalog: DigipackCatalog, promoCode: PromoCode, promotion: Promotion)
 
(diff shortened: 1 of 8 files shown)


AWare:
Happy, except for the copy/paste of the Membership date pretty printer. The time has probably come to move this common, play dependant code into a new library.

app/views/promotion/landingPage_incentive.scala.html
```diff
@@ -30,15 +36,17 @@
...
-                                @fragments.common.roundel(promotion.roundelHtml, Seq("roundel--raised"))
-                            </div>
+                        <div class="grid__item grid__item--flex-2-columns digipack__shots">
+                            @fragments.common.roundel(promotion.roundelHtml, Seq("roundel--raised"))
                         </div>
                         <div class="grid__item digipack__benefits">
-                            <p class="digipack__benefit">Experience your newspaper in full, on your tablet with the Guardian & Observer daily edition</p>
-                            <p class="digipack__benefit">Enjoy an advert-free experience on The Guardian's app with access to our daily crosswords and exclusive content</p>
-                            @subscribeNow(edition, catalog.digipackMonthly.priceGBP, promoCode)
+                            <div class="digipack__benefits--height">
```
joelochlann: in BEM the idea of modifier classes is that they are applied to the same element as the base class
```html
<div class="digipack__benefits digipack__benefits--height">
...
</div>

File: app/views/support/Dates.scala [hunk 2]
```diff
@@ -0,0 +1,130 @@
...
+    case 11 | 12 | 13 => "th"
+    case _ => day % 10 match {
+      case 1 => "st"
+      case 2 => "nd"
+      case 3 => "rd"
+      case _ => "th"
+    }
+  }
+
+  case
```
joelochlann: what's up with this use of whitespace?

app/views/support/Dates.scala
```diff
@@ -0,0 +1,130 @@
...
+    lazy val prettyNoYearWithTime = prettyDateNoYearTime(new DateTime(dt))
+
+    def pretty(includeTime: Boolean): String = if (includeTime) prettyWithTime else pretty
+
+    def prettyWithoutYear(includeTime: Boolean): String = if (includeTime) prettyNoYearWithTime else prettyWithoutYear
+
+    def isContemporary(threshold: Duration = 24.hours) = new Interval(dt - threshold, dt + threshold).contains(DateTime.now)
+  }
+
+  implicit
```
joelochlann: why all the newlines?

joelochlann:
Frontend stuff looks OK although the it's still not quite in keeping with the Block-Element-Modifier style. It's worth re-reading the [Client-side Principle](https://github.com/guardian/membership-frontend/blob/master/docs/client-side-principles.md) and associated links on that page.

Also there's some weird whitespace in `Dates.scala`, possibly a copy-paste mistake

Other than that :+1:

app/views/support/Dates.scala
paulbrown1982: I copied this file direct from membership-frontend. We're planning to put it into a shared library, so will tidy up the styling at that point.

app/views/promotion/landingPage_incentive.scala.html
paulbrown1982: Good spot - I'll change these - I'll create them as utility classes instead.

PR #360, opened 2016-02-18
Update a bunch of dependencies

rtyley:
cc @paulbrown1982

PR #363, opened 2016-02-18
Revert "Update a bunch of dependencies"

rtyley:
Reverts guardian/subscriptions-frontend#360 - I think this may have caused the production failures on the https://subscribe.theguardian.com/checkout/check-account endpoint (due to go-cardless update?!)

See also https://github.com/guardian/subscriptions-frontend/pull/361#issuecomment-185839152

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ lazy val root = (project in file(".")).enablePlugins(
       "com.gu.memsub.Subscription.ProductRatePlanId"
   ))
 
-scalaVersion := "2.11.7"
+scalaVersion := "2.11.6"
 scalacOptions ++= Seq("-feature")
 
 val scalatestVersion = "2.2.4"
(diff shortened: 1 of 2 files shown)


rtyley:
On my own run thru at least, this does solve the problem:

image

image

Looks very likely that the go-cardless library was the cause of this problem, will look into this tomorrow.

rtyley:
This was the error, when running locally:

```
java.lang.NoSuchMethodError: com.google.common.base.Stopwatch.createStarted()Lcom/google/common/base/Stopwatch;
	at com.gocardless.http.LoggingInterceptor.intercept(LoggingInterceptor.java:20)
	at com.squareup.okhttp.Call$ApplicationInterceptorChain.proceed(Call.java:232)
	at com.squareup.okhttp.Call.getResponseWithInterceptorChain(Call.java:205)
	at com.squareup.okhttp.Call.execute(Call.java:80)
	at com.gocardless.http.HttpClient.execute(HttpClient.java:148)
	at com.gocardless.http.HttpClient.execute(HttpClient.java:81)
	at com.gocardless.http.HttpClient.execute(HttpClient.java:76)
	at com.gocardless.http.PostRequest.execute(PostRequest.java:23)
	at services.GoCardlessService$$anonfun$checkBankDetails$1.apply$mcZ$sp(GocardlessService.scala:43)
	at services.GoCardlessService$$anonfun$checkBankDetails$1.apply(GocardlessService.scala:38)
	at services.GoCardlessService$$anonfun$checkBankDetails$1.apply(GocardlessService.scala:38)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

Turns out that the Google-Group Directory service client [has a transitive dependency](http://stackoverflow.com/questions/25792398/gradle-transitive-dependency-exclusion-is-not-working-as-expected-how-do-i-get) on `com.google.guava:guava-jdk5` - a nasty dependency that clashes with the regular `com.google.guava:guava` artifact, providing two versions of the same classes on your class path! Ah, in the old days sbt-assembly would have caught this, but we don't use it anymore.

Impressive coincidence that PR https://github.com/guardian/subscriptions-frontend/pull/358 (Google-Groups, plunging us to Guava 13.0) & PR https://github.com/guardian/subscriptions-frontend/pull/360 (Go-Cardless update, that suddenly requires the [Stopwatch.createStarted()](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Stopwatch.html#createStarted()) method introduced in Guava 15.0) would both have been ok on their own - but together, they clash.

PR #364, opened 2016-02-19
Soft release and more style tweaks to promo page

paulbrown1982:
* Tweaks for better responsivity and vertical centring.
* Removed Google auth gate to the page. This is essentially a soft go-live.
* Fixed a layout issue in desktop Safari.
* Fixed responsive image issue in IE and Opera.
* Tweaked grid--2up-step to create the 2 columns at phablet rather than tablet (it's only used on promo page)

/cc @joelochlann @rtyley


app/controllers/PromoLandingPage.scala
```diff
@@ -18,7 +18,7 @@ object PromoLandingPage extends Controller {
       case _ => None
     }
 
-  def render(promoCodeStr: String) = GoogleAuthenticatedStaffAction { implicit request =>
+  def render(promoCodeStr: String) = NoCacheAction { implicit request =>
```
joelochlann: I think we should be caching this (if it's cached for no more than a day then that's within acceptable bounds for disabling the promo once it's no longer valid)

joelochlann:
:+1: except for caching

PR #365, opened 2016-02-22
Turn on Promo Code box for all customers.

paulbrown1982:
Turn on Promo Code box for all customers.

Also includes fix for layout issue in mobile Safari when clearfixed & flex-based grids occur after inline elements - the display: table needs to not be there in that instance, but it definitely does when the grid is the first element inside a block. Non-Safari browsers are fine either way!

Switched to use a different promotion graphic from the grid.

/cc @joelochlann @rtyley


AWare:
:+1:

PR #366, opened 2016-02-24
Fix guava-jdk5 dependency introduced by Google Directory API client

rtyley:
https://github.com/guardian/play-googleauth/pull/37
https://github.com/guardian/subscriptions-frontend/pull/363#issuecomment-186190081

cc @paulbrown1982


paulbrown1982:
:+1:

rtyley:
Checked on PROD, the `/checkout/check-account` is working perfectly.

If you check `target/resolution-cache/reports/subscriptions-frontend-subscriptions-frontend_2.11-runtime-internal.xml` after doing a build (use Firefox to open the file for styling!) you'll also see that the `guava-jdk5` artifact is not included anywhere.

PR #370, opened 2016-02-26
Use membership-common 162

tomverran:
Headline change is we're using a Discounter object which handles as much of the promotion process as possible - it lives in membership-common and has unit tests.


paulbrown1982:
:+1:

tomverran:
image

PR #377, opened 2016-03-03
Use membership-common 167

tomverran:
This delays discounts until contract acceptance, to avoid invoices being generated too early


AWare:
:+1:

PR #379, opened 2016-03-07
Apply a promotion's discount to the rate plan descriptions

paulbrown1982:
- Applied a promotion's discount to the rate plan descriptions, and to the payment plan on the Thank You page.
- Also fixed bug with analytics amounts not always being GBP when a user changes their currency, and that the promotion was stored int he session if a user made a repeat purchase.
- Also ensured the promotion was available to all currencies and rate plans.

cc @tomverran @joelochlann 

picture 29

Now becomes:

picture 28

when a 16.9% promotion is applied.

picture 27

The adjusted price also appears on the Thank You page.

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -5,6 +5,7 @@ import com.gocardless.GoCardlessClient
 import com.gocardless.GoCardlessClient.Environment
 import com.gu.cas.PrefixedTokens
 import com.gu.config.{DigitalPackRatePlanIds, DiscountRatePlanIds, MembershipRatePlanIds, ProductFamilyRatePlanIds}
+import com.gu.i18n.{CountryGroup, Country}
 import com.gu.identity.cookie.{PreProductionKeys, ProductionKeys}
 import com.gu.memsub.auth.common.MemSub.Google._
 import com.gu.memsub.promo.Promotion._
@@ -125,11 +126,11 @@ object Config {
   def discountPromo(env: String): Option[AnyPromotion] = {
     val prpIds = digipackRatePlanIds(env)
     Some(Promotion(
-      appliesTo = AppliesTo.ukOnly(Set(
-        prpIds.digitalPackMonthly,
-        prpIds.digitalPackQuaterly,
-        prpIds.digitalPackYearly
-      )),
+      // TODO - create an AppliesTo.All
+      appliesTo = AppliesTo(
+        prpIds.productRatePlanIds,
+        CountryGroup.countries.toSet
+      ),
       campaignName = s"DigiPack for just £9.99 a month (~17% discount)",
       codes = PromoCodeSet(PromoCode("DPA30")),
       description = "For a limited time you can enjoy the digital pack for just £9.99 a month (usually £11.99). Get every paper delivered to your tablet for less than 35p an edition, plus an ad-free experience on your live news app.",
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -5,7 +5,6 @@ import com.gocardless.GoCardlessClient
 import com.gocardless.GoCardlessClient.Environment
 import com.gu.cas.PrefixedTokens
 import com.gu.config.{DigitalPackRatePlanIds, DiscountRatePlanIds, MembershipRatePlanIds, ProductFamilyRatePlanIds}
-import com.gu.i18n.{CountryGroup, Country}
 import com.gu.identity.cookie.{PreProductionKeys, ProductionKeys}
 import com.gu.memsub.auth.common.MemSub.Google._
 import com.gu.memsub.promo.Promotion._
@@ -126,11 +125,7 @@ object Config {
   def discountPromo(env: String): Option[AnyPromotion] = {
     val prpIds = digipackRatePlanIds(env)
     Some(Promotion(
-      // TODO - create an AppliesTo.All
-      appliesTo = AppliesTo(
-        prpIds.productRatePlanIds,
-        CountryGroup.countries.toSet
-      ),
+      appliesTo = AppliesTo.all(prpIds.productRatePlanIds),
       campaignName = s"DigiPack for just £9.99 a month (~17% discount)",
       codes = PromoCodeSet(PromoCode("DPA30")),
       description = "For a limited time you can enjoy the digital pack for just £9.99 a month (usually £11.99). Get every paper delivered to your tablet for less than 35p an edition, plus an ad-free experience on your live news app.",
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -1,17 +1,16 @@
 package controllers
 
 import actions.CommonActions._
-import com.gu.i18n.{Country, CountryGroup, GBP}
+import com.gu.i18n.{Country, CountryGroup, Currency, GBP}
 import com.gu.identity.play.ProxiedIP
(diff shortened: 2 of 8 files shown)


app/controllers/Checkout.scala
```diff
@@ -98,56 +98,55 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
...
 
-          val session = (productData ++ userSessionFields ++ appliedPromoCode).foldLeft(request.session) {
-            _ + _
-          }
+        val appliedPromoCode = validPromoCode.fold(Seq.empty[(String, String)])(validPromoCode => Seq(AppliedPromoCode -> validPromoCode.get))
 
-          catalog.find(formData.productRatePlanId).foreach { plan =>
-            trackAnon(SubscriptionRegistrationActivity(MemberData(e, formData, plan.billingPeriod)))
-          }
+        val session = (productData ++ userSessionFields ++ appliedPromoCode).foldLeft(request.session.-(AppliedPromoCode)) {
```
paulbrown1982: the -(AppliedPromoCode) bit is key so that a promotion isn't kept between 2 sequential payment flows.

assets/javascripts/modules/checkout/promoCode.js
```diff
@@ -22,24 +23,79 @@ define([
...
+                    amount = Number($el.data('amount-in-currency')),
+                    currency = $el.data('currency'),
+                    currencyIdentifier = $el.data('currency-identifier'),
+                    discountPercent = promotion.promotionType.amount,
+                    frequency = $el.data('frequency'),
+                    $label = $('#label-for-' + $el.val() + '-' + currency),
+                    newDisplayPrice = formatMoney(currencyIdentifier, amount * (1 - (discountPercent / 100))) + ' ' + frequency;
+
+                $label.html(newDisplayPrice);
+                $el.attr('data-option-mirror-label', newDisplayPrice);
```
paulbrown1982: using $el.data('option-mirror-label') wouldn't work as $.data() does not set back in to the DOM, and so wouldn't get picked up by the mirrorLabel.js module which reads the attribute.

app/views/checkout/payment.scala.html
```diff
@@ -21,15 +21,19 @@
         <span class="option__input">
             <input
             type="radio" name="ratePlanId"
+            data-option-mirror-label-default="@plan.prettyPricing(currency)"
             data-option-mirror-label="@plan.prettyPricing(currency)"
-            data-amount="@plan.pricing.getPrice(currency).get.amount"
```
paulbrown1982: This was the analytics bug which I fixed, where the price was getting set to the currency-applied amount but the analytics library is not sent the currency. So I've forced it to be gbpPrice, as is what's used in the other pages of the digipack flow.

app/configuration/Config.scala
```diff
@@ -125,11 +125,7 @@ object Config {
   def discountPromo(env: String): Option[AnyPromotion] = {
     val prpIds = digipackRatePlanIds(env)
     Some(Promotion(
-      appliesTo = AppliesTo.ukOnly(Set(
-        prpIds.digitalPackMonthly,
-        prpIds.digitalPackQuaterly,
-        prpIds.digitalPackYearly
-      )),
+      appliesTo = AppliesTo.all(prpIds.productRatePlanIds),
       campaignName = s"DigiPack for just £9.99 a month (~17% discount)",
```
tomverran: hmm is this right, the rest of the promotion marketing is very UK specific

paulbrown1982: Yup, I have had that confirmed by the Commercial team.

assets/javascripts/modules/checkout/promoCode.js
```diff
@@ -22,24 +23,79 @@ define([
         bindExtraKeyListener.alreadyBound = true;
     }
 
-    function displayError(message) {
-        message = message || 'Invalid promo code, please try again.';
+    function unapplyPromotionToRatePlans() {
```
tomverran: removePromotionFromRatePlans maybe?

app/controllers/Checkout.scala
```diff
@@ -171,18 +170,20 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
...
+      ratePlanId <- session.get(RatePlanId)
+      currencyStr <- session.get(SessionKeys.Currency)
+      currency <- Currency.fromString(currencyStr)
+    } yield (subsName, ratePlanId, currency)
 
     // TODO If some pieces of information are missing, redirect to an empty form. Is it the expected behaviour?
     def redirectToEmptyForm = Redirect(routes.Checkout.renderCheckout(CountryGroup.UK, None))
 
-    sessionInfo.fold(redirectToEmptyForm) { case (subsName, ratePlanId) =>
+    sessionInfo.fold(redirectToEmptyForm) { case (subsName, ratePlanId, currency) =>
```
tomverran: if we're not using currency here you could just make it a _

paulbrown1982: It is being used a few more lines down.

assets/javascripts/modules/checkout/promoCode.js
```diff
@@ -22,24 +23,79 @@ define([
...
+    function formatMoney(currencyIdentifier, amount) {
+        // rounds any remainder up tp 1p so we always quote a higher price than the user might eventually get charged.
+        var price2dp = Math.ceil(amount * 100) / 100,
+            priceNoDot00 = price2dp.toFixed(2).replace(/\.00$/, '');
+
+        return currencyIdentifier + priceNoDot00;
+    }
+
+    function applyPromotionToRatePlans(promotion) {
+        if (promotion.promotionType.amount > 0) {
```
tomverran: I appreciate there's probably a deadline on this so this doesn't block merging but I think it'd be nice to implement this price panel thing as a client side template with some kind of model that gets modified and redisplayed when a promotion is applied

So we have say something like

```javascript
plans: [
    { id: 1, price: 11.99, name: 'Monthly' },
    { id: 2, price: 30.99, name: 'Quarterly' },
    { id: 3, price: 120.99, name: 'Yearly' }
]
```
which get fed into a template rendered with the lodash template function (or react if you're feeling ambitious)
Then instead of all of this HTML stuff we can do something nice like 
```javascript
function applyPromotionToRatePlans(promotion) {
    var newPlans = map(plans, function(plan) {
        return { 
            name: plan.name, 
            price: plan.price * 1 + (discountPercent / 100),  //etc
            id: plan.id 
        }
    });
    $sidebar.innerHTML = template('plans.html', newPlans)
}
```

I find JS like this where the logic is mixed in with DOM operations quite difficult to read!
Alex and I got lodash templates working on membership which was a bit of a battle but I think its worth it

paulbrown1982: Agreed, and it should be linked in with the mirror-label thing too. Lodash templates is a good idea as it won't be a new full framework. I'll review what you guys have done in membership and we can bundle that change in to the limited-period-promotion work. I've captured this in comments for https://trello.com/c/3BiFWPkJ/385-data-requirements-for-offer-for-a-discounted-period-renewal-comms

tomverran:
:+1:

rtyley:
[Subs Outage incident retro 07 Mar 16](https://docs.google.com/document/d/1kdrWLBN63wivYV-w5lPWatHenBpU0h3y3oHJW6Jrau8/edit)

PR #382, opened 2016-03-10
remove test user flag from phone number field +make optional

AWare:
image


AWare:
@blackyjnz trello says between email and postcode, is this correct placement?

PR #391, opened 2016-03-22
Replace ZuoraService dependency with SubscriptionService / PaymentService

tomverran:
Basically this stuff was done before SubscriptionService existed.

Now it does exist we can make use of some of the higher level services to abstract us away from Zuora calls a bit

### DIFF
--- a/app/services/ExactTargetService.scala
+++ b/app/services/ExactTargetService.scala
@@ -30,17 +30,12 @@ trait ExactTargetService extends LazyLogging {
         _.getOrElse(throw new Exception(s"Subscription with no payment method found, ${subscribeResult.subscriptionId}"))
       )
 
-    val accAndPaymentMethod = for {
-      subs <- subscription
-      pm <- paymentMethod
-    } yield (subs, pm)
-
     for {
-      subs <- subscription
-      (acc, pm) <- accAndPaymentMethod
+      sub <- subscription
+      pm <- paymentMethod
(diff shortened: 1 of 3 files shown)


app/services/ExactTargetService.scala
```diff
@@ -19,27 +20,28 @@ import scala.concurrent.duration._
...
+  def paymentService: CommonPaymentService
 
   def sendETDataExtensionRow(subscribeResult: SubscribeResult, subscriptionData: SubscriptionData): Future[Unit] = {
-    val subscription = zuoraService.getSubscription(Subscription.Name(subscribeResult.subscriptionName))
+      val subscription = subscriptionService.unsafeGetPaid(Subscription.Name(subscribeResult.subscriptionName))(Digipack)
+      val paymentMethod = paymentService.getPaymentMethod(Subscription.AccountId(subscribeResult.accountId)).map(
+        _.getOrElse(throw new Exception(s"Subscription with no payment method found, ${subscribeResult.subscriptionId}"))
+      )
 
     val accAndPaymentMethod = for {
```
rtyley: `accAndPaymentMethod` is probably misnamed now, and looks like this stuff should be merged together:

https://github.com/guardian/subscriptions-frontend/blob/3f50bef63d/app/services/ExactTargetService.scala#L33-L40

paulbrown1982:
:+1: I got an email just fine when I registered as a dev user, so I'm assuming this is OK.

PR #393, opened 2016-03-23
Tracking for Jellyfish

rtyley:
[Jellyfish](http://www.jellyfish.co.uk/seo-social/) will no longer have their own landing pages for Subscriptions, but will instead send users directly to our pages. They have the job of showing appropriate online marketing to drive people to our site.

From Jellyfish:

> the codes that have been labelled as ‘Remarketing’ allow us to re-target to visitors who have come to our various 3rd Party campaigns (Facebook, Gmail Sponsored Promotions, Google) and not gone on to convert directly. We re-target this traffic through the same channel, and we remove any subscribers from these remarketing lists (which is why we require for the Remarketing codes to be added to the confirmation page as well as all other pages in the payment funnel so we can remove anyone that sees this page from being served an ad).
>
> The codes labelled as ‘Conversion’ we require on the confirmation page only of the payment funnel so that we can track all conversions in our respective channel reporting.

The codes were delivered to us by Jellyfish in a zip file: [GOBS_-_Tracking_Codes_For_New_Funnel.zip](https://github.com/guardian/subscriptions-frontend/files/186360/GOBS_-_Tracking_Codes_For_New_Funnel.zip)

https://trello.com/c/2vIHfmAR/426-promo-codes-jellyfish-tracking-pixels-to-support-migration

See further discussion at:

https://groups.google.com/a/guardian.co.uk/d/msg/subscriptions.dev/j3bAshulM-M/PDJYtyWSCwAJ

cc @paulbrown1982

### DIFF
--- a/app/views/fragments/analytics/jellyfish/remarketing.scala.html
+++ b/app/views/fragments/analytics/jellyfish/remarketing.scala.html
@@ -1,0 +1,54 @@
+@()
+
+@import configuration.Config
+
+@if(Config.stage == "PROD") {
+    <div class="is-hidden">
+        <!-- Google Code for Remarketing Tag -->
+        <!--------------------------------------------------
+    Remarketing tags may not be associated with personally identifiable information or placed on pages related to sensitive categories. See more information and instructions on how to setup the tag on: http://google.com/ads/remarketingsetup
+    --------------------------------------------------->
+        <script type="text/javascript">
+            /* <![CDATA[ */
+            var google_conversion_id = 995657703;
+            var google_custom_params = window.google_tag_params;
+            var google_remarketing_only = true;
+            /* ]]> */
+        </script>
+        <script type="text/javascript" src="//www.googleadservices.com/pagead/conversion.js">
+        </script>
+        <noscript>
+            <div style="display:inline;">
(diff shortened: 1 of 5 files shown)


paulbrown1982:
:+1:

rtyley:
The changes are now live on https://subscribe.theguardian.com/, with no Javascript errors occurring:

image

app/views/fragments/analytics/jellyfish/remarketing.scala.html
```diff
@@ -0,0 +1,54 @@
...
+        </script>
+        <script type="text/javascript" src="//www.googleadservices.com/pagead/conversion.js">
+        </script>
+        <noscript>
+            <div style="display:inline;">
+                <img height="1" width="1" style="border-style:none;" alt="" src="//googleads.g.doubleclick.net/pagead/viewthroughconversion/1008644981/?value=0&amp;guid=ON&amp;script=0"/>
+            </div>
+        </noscript>
+
+        <!-- Twitter single-event website tag code -->
```
rtyley: Whoops, this should have been `Google Analytics - Tracking Code.txt` instead.

PR #395, opened 2016-03-29
Free trial promo

tomverran:
This mops up most of the remaining changes Paul mentioned in his first commit, with the exception of some of the 14 day trial branding which I'll sort out later.


AWare:
:+1:

PR #397, opened 2016-03-29
Quick fix for guest checkout

tomverran:
I was tripped up by some code duplication, which I will fix properly after this fix gets deployed


AWare:
:+1:

PR #403, opened 2016-04-05
Dont send invalid phone numbers to identity

AWare:
Update membership-common to 0.185
https://github.com/guardian/membership-common/pull/212

Previously, we were relying on libphonenumber to perform all its validation on creation of the phone number object- this was only true on really invalid numbers. We can call a validation function from libphonenumber- which fixes this.


paulbrown1982:
:+1:

PR #405, opened 2016-04-06
Direct promotions

paulbrown1982:
- Confirmed latest changes (from Shane) of which new promotions will be included in our direct marketing URLs.
- Increased end times where necessary.
- Removed DPA30 test promo code.
- Made the (ProNetworks) discount promo last effectively forever.

https://trello.com/c/HlCohHqm/450-review-the-urls-provided-by-shane-to-make-sure-they-are-ok-and-working

https://trello.com/c/VjZ6gZYD/449-create-landing-page-for-jellyfish-orders-split-test-coming-through-the-gnm-flow

cc @tomverran

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -105,7 +105,7 @@ object Config {
   def demoPromo(env: String) = {
     val jellyFishPromoCode = PromoCode("DGA85")
     val prpIds = digipackRatePlanIds(env)
-    val promoCodes = (Seq(jellyFishPromoCode) /: (88 to 94))((seq, i) =>  seq :+ PromoCode(s"DGA$i") :+ PromoCode(s"DGB$i"))
+    val promoCodes = (88 to 94).foldLeft(Seq(jellyFishPromoCode))((seq, i) =>  seq :+ PromoCode(s"DGA$i") :+ PromoCode(s"DGB$i"))
 
     Promotion(
       appliesTo = AppliesTo.ukOnly(Set(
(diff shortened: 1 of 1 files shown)


app/configuration/Config.scala
```diff
@@ -111,7 +114,7 @@ object Config {
         prpIds.digitalPackYearly
       )),
       campaignName = "DigiPack - £30 digital gift card",
-      codes = PromoCodeSet(PromoCode("DGA88"), PromoCode("DGB88")),
+      codes = PromoCodeSet(PromoCode("DGA88"), promoCodes:_*),
```
paulbrown1982: http://stackoverflow.com/questions/6051302/what-does-colon-underscore-star-do-in-scala

app/configuration/Config.scala
```diff
@@ -103,7 +103,10 @@ object Config {
     MembershipRatePlanIds.fromConfig(ProductFamilyRatePlanIds.config(Some(config))(env, Membership))
 
   def demoPromo(env: String) = {
+    val jellyFishPromoCode = PromoCode("DGA85")
     val prpIds = digipackRatePlanIds(env)
+    val promoCodes = (Seq(jellyFishPromoCode) /: (88 to 94))((seq, i) =>  seq :+ PromoCode(s"DGA$i") :+ PromoCode(s"DGB$i"))
```
tomverran: So while I don't have the best record when it comes to avoiding obscure operators I found this quite tricky to read, it is a bit odd to fold a list and get a list back, you'd usually use some variant of map / flatMap for that.

I'd expect to see something more like this:
```scala
Seq(PromoCode("DGA85")) ++ (88 to 94).flatMap(n => Seq(PromoCode(s"DGA$n"), PromoCode(s"DGB$n")))
```

Though maybe that doesn't look much better after all

app/configuration/Config.scala
paulbrown1982: One could rewrite this like:
val promoCodes = (88 to 94).foldLeft(Seq(jellyFishPromoCode))((seq, i) =>  seq :+ PromoCode(s"DGA$i") :+ PromoCode(s"DGB$i"))

tomverran:
:+1:

PR #407, opened 2016-04-07
Making failed /p/ codes go to digipack homepage rather than 404

paulbrown1982:
- Changed logic so that uknown/expired/invalid /p/ promo codes redirect to the digipack UK homepage rather than show a 404 response. This is to encourage any links on the web to go somewhere useful.
- Added an INTCMP campaign code on the redirect so we can track these.

cc @tomverran @AWare


AWare:
👍

PR #408, opened 2016-04-07
Support 30-day free trial promotion

paulbrown1982:
- Removed grace period when under a FreeTrial promotion.
- Ensure we pass the existing 14 to Exact Target when there's no FreeTrial promotion.
- Pretty soon we'll add rather than subtract the grace period from the payment delay when sending the "Trial period" to Exact Target, the same as what we now do for Zuora in line 170 of CheckoutService.scala

https://trello.com/c/V2sQ9n9O/432-support-30-day-free-trial

cc @tomverran


tomverran:
:+1:

PR #410, opened 2016-04-07
Update Selenium to 2.52.0

mario-galic:
@AWare


AWare:
This won't go to the next version 2.53 without a play upgrade as play tests has some dep on some old functionality. Should revisit after our next upgrade.

AWare:
:+1:

PR #411, opened 2016-04-08
Remove code duplication on checkout

tomverran:
If I'm going to switch around the order in which we create accounts I'd like to only have to do so in one place


tomverran:
@AWare

AWare:
👍

PR #417, opened 2016-04-12
Adding Google Tag Manager to the checkout and thank you subs pages.

paulbrown1982:
Adding Google Tag Manager to the checkout and thank you subs pages.
As per: https://trello.com/c/V2sQ9n9O/432-support-30-day-free-trial
Shane confirmed that the tags only need go on the /checkout and /thankyou page. The tags he'll be loading are for Pangea and soon, TTP.

Namespaced the JellyFish Google Analytics tag as per the instructions here: https://support.google.com/tagmanager/answer/6103696?hl=en

cc @rtyley @AWare

### DIFF
--- a/app/views/fragments/analytics/jellyfish/remarketing.scala.html
+++ b/app/views/fragments/analytics/jellyfish/remarketing.scala.html
@@ -48,8 +48,8 @@
                 m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
             })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
-            ga('create', 'UA-44575989-1', 'auto');
-            ga('send', 'pageview');
+            ga('create', 'UA-44575989-1', 'auto', 'jellyfishGA');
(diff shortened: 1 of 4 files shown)


app/views/fragments/analytics/jellyfish/remarketing.scala.html
```diff
@@ -48,8 +48,8 @@
                 m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
             })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
 
-            ga('create', 'UA-44575989-1', 'auto');
-            ga('send', 'pageview');
+            ga('create', 'UA-44575989-1', 'auto', 'jellyfishGA');
+            ga('jellyfishGA.send', 'pageview');
```
rtyley: Good bug fix!

rtyley:
:+1: so the Google Tag Manager stuff is now owned by Shane Hoyle? Would be cool if we devs could get read-only access so we can understand what's going on a bit better.

PR #420, opened 2016-04-15
Create guest identity account only after successful salesforce call

joelochlann:
Fixes the bug in trello ticket https://trello.com/c/KsrsYjrv/407-subs-checkout-failure-when-using-valid-card.

To fix this, we need to be able to create a Salesforce contact before we've created a guest identity account. So we need to be able to create a contact without an identity id, and update that contact once we have the id.

@tomverran @rtyley

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -1,7 +1,7 @@
 package services
 
 import com.gu.config.DiscountRatePlanIds
-import com.gu.identity.play.AuthenticatedIdUser
+import com.gu.identity.play.{IdMinimalUser, AuthenticatedIdUser}
 import com.gu.memsub.services.PromoService
 import com.gu.memsub.services.api.CatalogService
 import com.gu.salesforce.ContactId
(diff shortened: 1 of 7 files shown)


app/services/CheckoutService.scala
```diff
@@ -42,101 +42,101 @@ class CheckoutService(identityService: IdentityService,
...
-    }).run
+      result <- EitherT(createSubscription(withPromo, purchaserIds, subscriptionData))
+      _ <- EitherT(sendETDataExtensionRow(result, subscriptionData, gracePeriod(subscribe, withPromo), purchaserIds))
+      identitySuccess <- storeIdentityDetails(personalData, authenticatedUserOpt, memberId)
+    } yield CheckoutSuccess(memberId, identitySuccess.userData, result, subscribe.promoCode)).run
+  }
+
+  private def storeIdentityDetails(personalData: PersonalData, authenticatedUserOpt: Option[AuthenticatedIdUser], memberId: ContactId): EitherT[Future, NonEmptyList[SubsError], IdentitySuccess] = {
+    authenticatedUserOpt match {
+      case Some(authenticatedIdUser) => {
```
tomverran: Very minor thing but you could replace this pattern match with a fold

app/services/CheckoutService.scala
```diff
@@ -42,101 +42,101 @@ class CheckoutService(identityService: IdentityService,
...
+        EitherT(updateAuthenticatedUserDetails(authenticatedIdUser, memberId, personalData))
+      }
+      case None => {
+        logger.info(s"User does not have an Identity account. Creating a guest account")
+        for (identitySuccess <- EitherT(identityService.registerGuest(personalData))) yield {
```
tomverran: could you replace these two nested for statements with one for comprehension?
I assume we want to swallow & log both identity reg errors and salesforce update errors

app/services/CheckoutService.scala
joelochlann: interesting to know, thanks. though personally I find the pattern match more readable

app/services/CheckoutService.scala
joelochlann: we don't want to swallow identity reg errors - `CheckoutSuccess` needs the identity id so it can set it in the session and allow the user to set their password on the next page.

salesforce update errors we can ignore as long as we keep an eye on them - the immediate UX doesn't depend on salesforce having the identity id (as far as I'm aware) and as long as we keep an eye on these errors we have all the information to manually fix if necessary

tomverran: ah yes of course fair enough

tomverran:
👍

PR #421, opened 2016-04-19
Prepopulate international country from identity

paulbrown1982:
Fixed issue where an Identity user had a None country, we weren't correctly setting the country dropdown or checking the set of rate plans. Most of the rest of the code was collateral from changing Address.country to an Option[Country].

Fixes: https://trello.com/c/lOdZUGdi/486-checkout-failure-for-existing-eu-identity-user-subscribing-to-the-digital-pack

And indirectly fixes: https://trello.com/c/yt3TSe0Z/465-eu-and-international-members-cannot-subscribe

As well as: https://app.getsentry.com/the-guardian/subscriptions/issues/123751035/

cc @AWare @rtyley @tomverran @mario-galic

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -1,4 +1,4 @@
-@import com.gu.i18n.{CountryGroup, Currency}
+@import com.gu.i18n.{Country, CountryGroup, Currency}
 @import com.gu.memsub.{BillingPeriod, Month}
 @import com.gu.subscriptions.DigipackPlan
 @import configuration.Config.Identity._
@@ -10,8 +10,9 @@
     userIsSignedIn: Boolean,
     plans: Seq[DigipackPlan[BillingPeriod]],
     defaultPlan : DigipackPlan[Month],
+    country: Option[Country],
     countryGroup: CountryGroup,
-    currency: Currency,
+    defaultCurrency: Currency,
     countriesWithCurrency: List[CountryWithCurrency],
     touchpointBackendResolution: services.TouchpointBackend.Resolution)(implicit request: RequestHeader)
 
@@ -30,7 +31,7 @@
             data-number-of-months="@plan.billingPeriod.numberOfMonths"
             data-currency="@currency"
             value="@plan.productRatePlanId.get"
-            @if(plan == defaultPlan){checked}
+            @if(plan == defaultPlan && currency == defaultCurrency){checked}
             >
         </span>
         <span class="option__label" id="label-for-@plan.productRatePlanId.get-@currency">@plan.prettyPricing(currency)</span>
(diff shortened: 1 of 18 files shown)


app/views/checkout/payment.scala.html
```diff
@@ -30,7 +31,7 @@
             data-number-of-months="@plan.billingPeriod.numberOfMonths"
             data-currency="@currency"
             value="@plan.productRatePlanId.get"
-            @if(plan == defaultPlan){checked}
+            @if(plan == defaultPlan && currency == defaultCurrency){checked}
```
paulbrown1982: This is the actual fix to correctly pre-select the rate plan radio button on the checkout page.

rtyley:
If I understand correctly:

* Country is optional in Identity, or could possibly be set to something we don't recognise
* When a user submits the form at https://subscribe.theguardian.com/checkout, they will definitely have a correct Country.

This means it seems a little unfortunate to me that we need to make country an Optional field in Address, is there no easy way to get around it?

paulbrown1982:
It's because the county is derived from the Identity countryName String field, which validly could be null or empty: so the old lazy val was unsafe, hence why I made it an option. 

Defaulting the Address.country to UK (i.e. implying UK from an empty countryName) breaks the precedence ordering which controls country the drop down (e.g. ?countryGroup=US) when visiting from the US supporter page, because the country from Identity must win out if it's not None...

tomverran:
I am on board with this being merged now to fix the immediate issue 👍 

To weigh in on the design on this it looks like you basically have two different Addresses and as Roberto says it is a bit of a shame to have an Option[Country] that we _know_ must exist.

So we need:

 - An address from identity which has an optional country
 - An address from our sign up form which has a compulsory country

We can model this like so! (I skipped some fields)

```scala

import com.gu.i18n.CountryGroup
import com.gu.i18n.Country
import scalaz.Monad
import scalaz.Id._

case class Address[M[_]](lineOne: String, lineTwo: String, country: M[Country])

object Address {
  type IdentityAddress = Address[Option]
  type FormAddress = Address[Id]

  def fromIdentity(lineOne: String, lineTwo: String, country: String): IdentityAddress =
    Address[Option](lineOne, lineTwo, CountryGroup.countryByNameOrCode(country))

  def fromForm(lineOne: String, lineTwo: String, country: String): FormAddress =
    Address[Id](lineOne, lineTwo, Monad[Id].point(CountryGroup.countryByNameOrCode(country).get))
}

Address.fromIdentity("123 Fake St", "Faketown", "UK").country // Option[Country]
Address.fromForm("123 Fake St", "Faketown", "GB").country     // Country


```

paulbrown1982:
I've raised https://trello.com/c/NHQFOKeQ/493-tech-debt-identity-vs-membership-address-country-modelling in this sprint to tackle the modelling concerns raised. However I'll merge this change to fix the production issues.

PR #425, opened 2016-04-20
Revert "Revert "Create guest identity account only after successful salesforce call""

rtyley:
Reverts #422, reinstating #420 , now that the Salesforce API tweak is present in the PROD Salesforce environment:

---------- Forwarded message ----------
From: Michael Ojo
Date: 20 April 2016 at 10:50
Subject: Register Customer Now in PROD

Hi All,

Register Customer updated service is now in PROD.

Cheers

cc @paulbrown1982 @joelochlann @tomverran


paulbrown1982:
👍

PR #426, opened 2016-04-20
Add X-Gu-Identity-Id & X-Gu-Membership-Test-User headers

rtyley:
Verifying that https://github.com/guardian/subscriptions-frontend/pull/425 is succseful by looking at Fastly logs isn't possible, because although I can see users successfully hitting `/checkout/thank-you`, I can't easily see if they're logged in.

This code is borrowed from membership-frontend - but is slightly better, I think, because it uses a filter rather than an action builder.

cc @AWare


AWare:
:+1:

rtyley:
Verified headers are coming back on PROD:

Signed in on Checkout:
image

Signed out on Checkout (ie about to join as a guest user):
image

On the cached main page:
image

PR #431, opened 2016-04-26
Style changes for promo landing page

paulbrown1982:
Style changes for promo page. Only thing different from design is when the grey borders show and the vertical-centring of the items next to the right of the pack shots. I'll speak to Ben about this. Those fixes will be done once we include the new promotion: https://trello.com/c/20wV3iYm/383-promo-codes-support-pact-coffee-campaign-building-a-landing-page 
There's also a bug with the footer on IE11, which affects all subs pages, so I've raised another ticket to fix that across the whole site. https://trello.com/c/AnBGwGNL/500-ie11-big-space-before-footer-in-subs

https://trello.com/c/RDgRsrb5/400-design-tweaks-for-subs-promo-code-landing-pages

cc @joelochlann @tomverran @AWare @rtyley


assets/stylesheets/modules/_promotion.scss
```diff
@@ -1,114 +1,222 @@
...
+    border-width: 1px 0 0 0;
+    margin-left: 70px;
+    padding-top: ($gs-baseline / 3);
+
+    @include mq(phablet) {
+        border: 0;
+        margin-left: 0;
+    }
+
+    > h3 {
```
tomverran: I think the BEM convention is to avoid path based selectors like this in favour of more explicit class names

paulbrown1982: I've renamed this to feature__text

assets/stylesheets/modules/_promotion.scss
```diff
@@ -1,114 +1,218 @@
...
+        font-size: 18px;
+        font-weight: 900;
+    }
 
+    > p {
```
tomverran: I can't remember what was decided with these direct descendent selectors but I believe classes like so are more idiomatic BEM
 
```css
.feature--text__body
.feature--text__heading
```

where you're modifying the `text` element in the `feature` block with either `body` or `heading`

I think the button example in the below link is quite similar to this
http://getbem.com/introduction/

I am by no means a BEM expert though so if you chatted to someone else about this and they were happy then don't let me block you

tomverran:
👍

mario-galic:

PR #438, opened 2016-05-05
Added Pact Coffee and student promotions

paulbrown1982:
- Added the Pact Coffee promotion, plus the student promotion.
- Pulled the JellyFish Google Analytics and remarketing beacons into the site's existing JS framework for that - so it now applies on all pages, and saves a download and re-evaluation of the Google JS.
- Removed the fallbackCode stuff to make membership common upgrade easier.
- Tweaked some template signatures to pass in least-power promotion type or just Strings where possible.

cc @tomverran


tomverran:
👍 incidentally should we still have travis building this

mario-galic:

PR #441, opened 2016-05-09
Update CF with promo dynamodb read permissions

tomverran:


cloud-formation/subscriptions-app.cf.json
```diff
@@ -205,7 +232,8 @@
                 "ConnectionDrainingPolicy": {
                     "Enabled" : "true",
                     "Timeout" : "60"
-                }
+                },
+                "CrossZone": "true"
```
paulbrown1982: As we're about to do a CloudFormation change, I've drive-by'd this very small tweak (which we use in many other places) to enable cross-zone routing from the ELB, which is recommended in the Trusted Advisor.

tomverran:
This is what is live now so I am going to go ahead and merge it

mario-galic:

PR #442, opened 2016-05-09
Landing page strip colour

paulbrown1982:
Created new LandingPageOps class which handles new optional Incentive T&Cs text and ensuring the desired strip colour appears.

cc @tomverran

### DIFF
--- a/app/views/support/LandingPageOps.scala
+++ b/app/views/support/LandingPageOps.scala
@@ -2,12 +2,13 @@ package views.support
 
 import com.gu.memsub.promo.Promotion.{AnyPromotion, AnyPromotionWithLandingPage}
 import com.gu.memsub.promo._
+import org.joda.time.DateTime.now
 
 object LandingPageOps {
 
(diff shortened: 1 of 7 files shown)


app/views/support/LandingPageOps.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import com.gu.memsub.promo.Promotion.{AnyPromotion, AnyPromotionWithLandingPage}
+import com.gu.memsub.promo._
+
+object LandingPageOps {
+
+  private def getSectionColour(promotion: AnyPromotionWithLandingPage) = promotion.landingPage.sectionColour.getOrElse(Blue)
+
+  implicit class ToLandingPageClassName(promotion: AnyPromotionWithLandingPage) {
+    def landingPageSectionColour: String = {
```
tomverran: this is a tiny tiny thing but I think you can have one implicit class with multiple methods rather than an implicit class per method

tomverran:
👍

tomverran:
👍

mario-galic:

PR #447, opened 2016-05-12
Fix raven

paulbrown1982:
- Ensuring modules which use Raven pull in our wrapper module. _Except those required by our raven wrapper module!_
- Also fixed capitalisation typo in the interface of our Raven wrapper module. Makes it consistent with membership-frontend.

cc @AWare @mario-galic

### DIFF
--- a/assets/javascripts/utils/base64.js
+++ b/assets/javascripts/utils/base64.js
@@ -1,5 +1,5 @@
 /*global escape*/
-define(['modules/raven'], function (raven) {
+define(['raven'], function (raven) {
     'use strict';
 
     function AtoB(){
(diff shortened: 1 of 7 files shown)


assets/javascripts/utils/base64.js
```diff
@@ -46,7 +46,7 @@ define(['raven'], function (raven) {
         try {
             decoded = decodeURIComponent(escape(new AtoB()(str.replace(/-/g, '+').replace(/_/g, '/').replace(/,/g, '='))));
         } catch(e){
-            raven.Raven.captureException(e, {tags: { level: 'info' }});
```
AWare: This is where the circular dep was, I think it could do with a comment warning that it's required by raven.

AWare:
👍

PR #451, opened 2016-05-18
Use membership common 204

tomverran:
This is so subs frontend can read promotions created with the promo tool


AWare:
ftfy 👍

tomverran:
oh wow I'd forgotten we were on the 200s

I'm like this with dates too, still writing 2015 in May

PR #455, opened 2016-05-23
Set a correct X-Frame-Options header on the promotion preview page.

paulbrown1982:
- This allows the promo preview page to be viewable in the promo code tool.

cc @nlindblad @tomverran

### DIFF
--- a/app/filters/AffectXFrameOptionsHeader.scala
+++ b/app/filters/AffectXFrameOptionsHeader.scala
@@ -13,7 +13,7 @@ object AffectXFrameOptionsHeader extends Filter {
   def apply(nextFilter: RequestHeader => Future[Result])(requestHeader: RequestHeader): Future[Result] = {
     nextFilter(requestHeader).map { result =>
       if (requestHeader.path.startsWith("/q/")) {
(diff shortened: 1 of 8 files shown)


app/filters/AffectXFrameOptionsHeader.scala
```diff
@@ -0,0 +1,22 @@
...
+import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import play.api.libs.ws.WS
+import play.api.mvc._
+
+import scala.concurrent.Future
+
+object AffectXFrameOptionsHeader extends Filter {
+  def apply(nextFilter: RequestHeader => Future[Result])(requestHeader: RequestHeader): Future[Result] = {
+    nextFilter(requestHeader).map { result =>
+      if (requestHeader.path.startsWith("/q/")) {
```
tomverran: I'd be slightly more comfortable with an X-Frame-Options-Override header or something as opposed to matching on the URL

paulbrown1982: That's a great idea, doing that now.

tomverran:
👍

PR #469, opened 2016-06-10
Add /month to cta text to remove any moment of doubt

paulbrown1982:
Also improved template around the description so <p> is used in both sides of the fold.
cc @tomverran


tomverran:
👍

PR #476, opened 2016-06-16
More copy tweaks to help with conversion

paulbrown1982:
- Moved the DGA85 bit into the promoCode section to keep things tidier, and because it just wasn't right being in the header.
- Put the sort code before the account number.
- Added the direct debit logo.
- Put the lock symbol on the first section too.
- Tidied up HTML a bit for address fields.

https://trello.com/c/lMC63OmG/32-update-digipack-checkout-page-to-improve-conversion

cc @AWare


AWare:
:+1: 
Had a look at these changes yesterday, hopefully we'll see an improvement in conversions. :)

PR #477, opened 2016-06-22
Add delivery data

tomverran:
Currently subscriptions uses one object to model the form you submit but we need to split it such that you always submit your personal information but if you're signing up to paper we expect different fields to digipack

As such there's now a SubscribeRequest object to replace SubscriptionData which contains always your personal details and then either digipack details or paper details

For better type safety we also changed the model to have an entire plan in it rather than just the ID, so we can ensure that the digipack form gets a digipack plan, this then required the form to know about the catalog service

### DIFF
--- a/app/model/SubscriptionData.scala
+++ b/app/model/SubscriptionData.scala
@@ -1,12 +1,14 @@
 package model
 
 import com.gu.i18n.Country.UK
-import com.gu.i18n.{Title, Country, CountryGroup}
+import com.gu.i18n.{Country, CountryGroup, Title}
 import com.gu.identity.play.IdUser
 import com.gu.memsub.Subscription.ProductRatePlanId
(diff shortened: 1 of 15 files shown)


app/model/SubscriptionData.scala
```diff
@@ -52,12 +54,24 @@ case class PersonalData(first: String,
...
+case class DigipackData(
+   plan: DigipackPlan[BillingPeriod]
+)
+
+case class PaperData(
+  startDate: LocalDate,
+  deliveryAddress: Address,
+  deliveryInstructions: Option[String],
+  plan: ProductRatePlanId,
+  includesDigipack: Boolean,
```
tomverran: this model is not final yet but its a starting point

paulbrown1982:
👍

PR #479, opened 2016-06-24
More email checks

tomverran:
Double check we're not about to create subscriptions when we know identity is going to fail

Also if identity or exact target _does_ fail we still show the thank you page as by that point we've signed you up so you have a valid (if orphaned) account


nlindblad:
👍

PR #481, opened 2016-06-27
Form population

tomverran:
The main checkout form in Subscriptions is rendered just with HTML but is populated using data from a `Form[SubscriptionData]` Using play forms allows the following:

- Easy display of validation errors
- Easy repopulation of the form with what the user submitted if errors do occur

Since we're submitting the form with Javascript and do all validation client side (and throw a 500 if the backend validation is out of sync) we don't actually gain anything by passing a Form[SubscriptionData] to the view but we do suffer from the following disadvantages:

- Any more advanced modelling (e.g. populating with Either[PaperData, DigipackData] is tricky
- It isn't possible to pass only parts of the data to view fragments - e.g. `SubscriptionData` contains an `Address` but you can't obviously get a `Form[Address]` from a `Form[SubscriptionData]` so the address view fragment needs the entire form
- No static analysis / type safety on the fields we actually read to populate the form - Since to get a value you call `form.get("fieldname")` you can't do any refactoring on the names of fields etc as they're all only referred to as strings


tomverran:
Niklas says 👍

PR #482, opened 2016-06-27
Remove duplicate promo code field

tomverran:


tomverran:
Niklas once again gives his 👍

PR #487, opened 2016-06-28
Fix the promise polyfill

tomverran:
When we switched from bower to NPM we copied across the libraries we use but unfortunately promise-polyfill places its Promise object onto the object loading it, rather than specifically the window

As such it used to work when we used it via bower but quietly does nothing now we use NPM


nlindblad:
👍 Well spotted!

PR #489, opened 2016-06-29
Paper form

tomverran:
- Introduce new SubscriptionsForm model to populate the form with
- Support for both delivery and billing address (which is not used at the moment)
- Paper form locked behind Google Auth for now
- Extra form fields for start date and delivery instructions
- Split out address into reusable view fragment
- Tested end-to-end from form to Zuora

Sorry this PR is quite massive but @paulbrown1982 would you mind taking a look

### DIFF
--- a/app/services/TouchpointBackend.scala
+++ b/app/services/TouchpointBackend.scala
@@ -8,7 +8,7 @@ import com.gu.memsub.services.{PaymentService => CommonPaymentService, _}
 import com.gu.monitoring.{ServiceMetrics, StatusMetrics}
 import com.gu.salesforce.SimpleContactRepository
 import com.gu.stripe.StripeService
-import com.gu.subscriptions.{DigipackCatalog, Discounter}
+import com.gu.subscriptions.{DigipackCatalog, Discounter, PaperCatalog}
 import com.gu.zuora
 import com.gu.zuora.{rest, soap}
 import configuration.Config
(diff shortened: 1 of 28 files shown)


paulbrown1982:
👍

app/services/TouchpointBackend.scala
```diff
@@ -111,6 +113,7 @@ case class TouchpointBackend(environmentName: String,
                              catalogService : api.CatalogService,
                              zuoraService: zuora.api.ZuoraService,
                              subscriptionService: SubscriptionService[DigipackCatalog],
+                             subscriptionServicePaper: Option[SubscriptionService[PaperCatalog]],
```
paulbrown1982: This is temporary until we have the rate plans in prod Zuora.

assets/javascripts/modules/checkout/promoCode.js
```diff
@@ -82,8 +81,8 @@ define([
     }
 
     function validate() {
-        var country = countryChoice.getCurrentCountryOption().value.trim(),
-            promoCode  = $inputBox.val().trim();
+        var country = formElements.BILLING.$COUNTRY_SELECT.val().trim(),
```
paulbrown1982: Just double check whether this should be delivery or billing address - i.e. do whatever Membership does.

tomverran: So as discussed - for digipack any physical goods will go to your billing address, for paper we'll change this to use your delivery address country

PR #492, opened 2016-07-01
Validate non emptiness of delivery address fields clientside

tomverran:
This will currently break the backend code as the fields get disabled if you choose to use your billing address but I shall fix that in a separate PR


nlindblad:
👍

PR #493, opened 2016-07-01
Fix login url to maintain promo code on redirect back

paulbrown1982:
Fixed login url callback to maintain promo code and country.

This should help conversion a tiny bit.

cc @tomverran


tomverran:
👍

PR #494, opened 2016-07-01
If no delivery address is submitted then use the billing address

tomverran:
The frontend disables the delivery address form fields if you leave "Use my billing address" selected so we need to adjust the backend such that if it gets no delivery address it has a go at getting an address from the billing address instead


paulbrown1982:
👍

PR #495, opened 2016-07-01
Send billing / delivery address to identity and salesforce

tomverran:
This is quite painful due to delivery address not being in personal data


paulbrown1982:
👍

PR #496, opened 2016-07-02
Make the global-nav not lowercase.

AWare:
old: 
image

NEW:
image

I'm not sure why this was lowercase. What do you all think?


paulbrown1982:
👍

PR #498, opened 2016-07-04
Product specific thankyou page

paulbrown1982:
- Carried forward the product rate plan description to the thank you page.
- Added new URL routes for the paper flow (these might change depending on how we want to do analytics)
- Added button on /test-users page to test paper flows.

Still to do:

- Deeplinking to a package
- Emails
- Redirects/confirmation of the digipack checkout URLs once analytics have been tweaked.
cc @tomverran

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -58,10 +58,14 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
     SubscriptionsForm.subsForm.bind(Map("promoCode" -> promoCode.fold("")(_.get)))
 
   def renderPaperCheckout(countryGroup: CountryGroup, promoCode: Option[PromoCode]) = AuthorisedTester.async { req =>
-    renderCheckout(countryGroup, promoCode, productFamily = Paper).apply(req)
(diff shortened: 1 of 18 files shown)


app/controllers/Checkout.scala
```diff
@@ -57,11 +57,36 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
...
 
+  def getCheckoutRouteForProductFamily(countryGroup: CountryGroup, promoCode: Option[PromoCode], product: ProductFamily) = {
+    product match {
+      case Paper => routes.Checkout.renderPaperCheckout(countryGroup, promoCode)
+      case Digipack => routes.Checkout.renderDigipackCheckout(countryGroup, promoCode)
+      case _ => routes.Homepage.index()
+    }
+  }
+
+  def renderProductCheckout(countryGroup: CountryGroup, promoCode: Option[PromoCode], product: ProductFamily) = AuthorisedTester.async { req =>
```
tomverran: am I right in thinking that the intention is to replace renderPaper / renderDigipack with calls to this when we release paper & sort out analytics?

PR #500, opened 2016-07-05
Fix React loading

nlindblad:
Gets rid of console error and makes sure React is only loaded for paper checkout.

**The error in question**:

image

Fixed by following [this](http://dev.topheman.com/make-your-react-production-minified-version-with-webpack/).

React is now loaded async on the paper checkout page only:

**Digipack checkout**:
screen shot 2016-07-05 at 17 41 59

**Paper checkout**:
screen shot 2016-07-05 at 17 42 36


paulbrown1982:
👍

PR #502, opened 2016-07-06
Move around addresses

tomverran:
## before

image

## after

image

image

## paper bonus

image

image

image


nlindblad:
LGTM 👍  Nice to present people with a shorter form to begin with.

paulbrown1982:
Looks good. To make it great, I think we should put Country under the postcode, rather than being at the top.

paulbrown1982:
👍

PR #503, opened 2016-07-06
Restrict date picker and change date format

nlindblad:
Restriction is now 5 days into the future and then two weeks onward.

Changed the date format to DD MMMM YYYY.

image


paulbrown1982:
Let's make the date format be the English-friendly: d MMMM YYYY,  e.g. 3 March 2016

paulbrown1982:
👍

PR #506, opened 2016-07-07
Fixing country select and county review

paulbrown1982:
- Moved country select up above the fields it affects. This creates a better UX when filling out the form.
- For digipack and paper: Fixed bug where county/state was missing in review address section.
- For paper subs: Added note to confirm that billing address is same as delivery.
- Made review sections be in same order as form.

picture 142
picture 141

cc @tomverran @nlindblad


nlindblad:
👍

PR #510, opened 2016-07-11
Fix country name in analytics event

nlindblad:
**Before**:
```Map(..., country -> Some(Country(GB,United Kingdom,List())), ...)```

**After**:
```Map(..., country -> United Kingdom, ...)```


tomverran:
👍

PR #511, opened 2016-07-11
Add a character limit to the delivery instructions

paulbrown1982:
Add a character limit to the delivery instructions, by making it a React module.

cc @nlindblad @tomverran


nlindblad:
Nice! 👍

PR #512, opened 2016-07-13
Use new subs modelling

tomverran:
This leaves the subscriptions checkout entirely ignorant as to exactly what it is subscribing you to, with the only constraint being it must be a rate plan within one of the whitelisted products.

Now it turns out this leaves the checkout actually _too_ ignorant because it now no longer cares whether you submit an address or not, but I'll fix that later.


app/controllers/Checkout.scala
```diff
@@ -216,26 +205,26 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
...
       ratePlanId <- session.get(RatePlanId)
       currencyStr <- session.get(SessionKeys.Currency)
       currency <- Currency.fromString(currencyStr)
-    } yield (subsName, ratePlanId, currency)
+    } yield (subsName, plan, currency)
 
     sessionInfo.fold {
-      Redirect(getCheckoutRouteForProductFamily(product = product)).withNewSession
-    } { case (subsName, ratePlanId, currency) =>
+      Redirect(routes.Homepage.index()).withNewSession
```
tomverran: Hopefully this is okay - we just go to the homepage if you visit the thank you page with no plan to simplify the logic here, because product family doesn't differentiate between paper & digipack

conf/routes
```diff
@@ -31,8 +31,8 @@ GET         /checkout/lookupPromotion        controllers.Checkout.validatePromoC
 
 # Checkout pages
 GET         /checkout                        controllers.Checkout.renderCheckout(countryGroup: CountryGroup ?= CountryGroup.UK, promoCode: Option[PromoCode])
-GET         /checkout/thank-you              controllers.Checkout.thankYouDigipack
-GET         /checkout/paper/thank-you        controllers.Checkout.thankYouPaper
+GET         /checkout/thank-you              controllers.Checkout.thankYou()
+GET         /checkout/paper/thank-you        controllers.Checkout.thankYou()
```
paulbrown1982: Will this /paper/ endpoint go once we put paper properly live?

paulbrown1982:
👍

conf/routes
tomverran: yep

PR #514, opened 2016-07-14
Membership common 234

tomverran:
This expands the use of the PaperCatalog which now also contains digipack plans and simplifies some code as a result


paulbrown1982:
👍

PR #519, opened 2016-07-14
Exact target changes for paper delivery

paulbrown1982:
Put the relevantly formatted JSON message onto the subs-welcome-email queue, plus a field to determine which type of message it is, so that the workflow service can broker it to the right Exact 
Target Data Extension.

cc @mario-galic @tomverran


tomverran:
👍

mario-galic:
👍

PR #520, opened 2016-07-15
Address population

tomverran:
Populate billing address with billing address only and delivery address with delivery address only, also only hide the billing address on paper if it isn't filled in to prevent it being silently overwritten


paulbrown1982:
👍

paulbrown1982:
👍

PR #523, opened 2016-07-15
Ensure the rate plan description is re-rendered

paulbrown1982:
- Ensure the rate plan description is re-rendered when the user changes their package.
- Improved the layout of the test users page.

cc @tomverran


tomverran:
👍

PR #524, opened 2016-07-18
Only United Kingdom is available for home delivery subs.

paulbrown1982:
- Overrode the derived country group when it's a home delivery subscription.
- Refactored how the arguments which get passed in to the checkout template are determined.

cc @tomverran

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -69,22 +69,23 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
         .fold({ case (plan, group) => Left(PlanList(plan, group.betterThan(plan).productPlans.sortBy(_.priceGBP.amount):_*)) },
               { case (plan, group) => Right(PlanList(plan, group.betterThan(plan).productPlans.sortBy(_.priceGBP.amount):_*)) })
 
-      val productPlan = planListEither.fold(identity, identity).default
       val personalData = user.map(PersonalData.fromIdUser)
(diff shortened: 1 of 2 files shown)


app/controllers/Checkout.scala
```diff
@@ -73,22 +69,31 @@ object Checkout extends Controller with LazyLogging with ActivityTracking with C
...
-      val countryGroupWithDefault = countryOpt.fold(countryGroup)(c => CountryGroup.byCountryCode(c.alpha2).getOrElse(countryGroup))
-      val country = countryOpt orElse countryGroupWithDefault.defaultCountry
-
-      val desiredCurrency = countryGroupWithDefault.currency
-      val supportedCurrencies = planListEither.fold(identity, identity).default.currencies
-      val defaultCurrency = if (supportedCurrencies.contains(desiredCurrency)) desiredCurrency else GBP
+      val productData = ProductPopulationData(user.map(_.address), planListEither)
+      val preselectedCountry = personalData.flatMap(data => data.address.country)
+      val countryGroupForPreselectedCountry = preselectedCountry.flatMap(c => CountryGroup.byCountryCode(c.alpha2))
+      val determinedCountryGroup = if (productPlan.products.seq.contains(Delivery)) {
```
tomverran: voucher won't contain delivery - you could check whether the checkout is for physical products with `planListEither.isRight`

paulbrown1982: I wasn't going to worry about voucher right now, to assist with testing, but as it makes the code simpler to be generic, I'll make it work for voucher now too.

PR #526, opened 2016-07-18
Exact target paper fix

paulbrown1982:
- Fixing issues with welcome 1 where I typed the wrong field names
- Missed fields delivery_instructions and date_of_first_paper

cc @tomverran


tomverran:
👍

PR #527, opened 2016-07-18
Support tracking codes on the subs checkout

tomverran:
It doesn't look like the subs checkout ever worked with tracking codes(!)

Hence this PR

 - Puts tracking codes from `?promoCode=blah` into the session
 - Fills in `suppliedPromoCode` with the tracking code if there's no promo code
 - Filters out tracking codes on the thank you page


paulbrown1982:
👍

PR #528, opened 2016-07-19
Paper packshots

tomverran:
I'm not entirely sure how great these images look but they can be changed before going live

image


nlindblad:
👍

PR #529, opened 2016-07-19
Added a postcode check for home delivery subs. 

paulbrown1982:
- Added a postcode check for home delivery subs. 
- Auto-uppercased any typed in chars. (Delivery only for now)
- Added Isle of Man to the country drop down list for voucher subs.
- Fixed the footer email address on thank you page.

cc @tomverran @nlindblad

### DIFF
--- a/app/model/HomeDeliveryPostCodes.scala
+++ b/app/model/HomeDeliveryPostCodes.scala
@@ -1,0 +1,81 @@
+package model
+
+import java.lang.Math.min
+
+object HomeDeliveryPostCodes {
+  private val London = Set(
+    "E1", "EC1", "N1", "NW1", "SE1", "SW1", "W1", "WC1",
+    "E10", "EC2", "N10", "NW10", "SE10", "SW10", "W10", "WC2",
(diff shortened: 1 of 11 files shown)


app/model/HomeDeliveryPostCodes.scala
```diff
@@ -0,0 +1,81 @@
...
+    "GU25",
+    "GU3",
+    "GU4"
+  )
+
+  private val availableDistricts = London ++ M25 ++ Outer
+
+  def findDistrict(postCode: String): Option[String] = {
+    val sanitised = postCode.toUpperCase.trim
+    if (sanitised.isEmpty) None else availableDistricts.find(d => d.startsWith(sanitised) || sanitised.startsWith(d))
```
paulbrown1982: I wonder if there's a faster way to do this?

tomverran:
👍

app/model/HomeDeliveryPostCodes.scala
tomverran: Almost certainly there is but it probably isn't worth it really

PR #532, opened 2016-07-21
Removed the dynamically driven packages till the prod rate plans are loaded

paulbrown1982:
Updated the test screen prices to include weekly amounts too (quick QA), plus removed the dynamically driven packages till the prod rate plans are loaded.

cc @tomverran @nlindblad


tomverran:
👍

PR #533, opened 2016-07-21
Fix promo code

paulbrown1982:
- Fixed bug where the wrong package names were displayed when setting or resetting a promotion.
- Only show Sunday delivery start dates for SUNDAY+ package.

cc @tomverran @nlindblad


tomverran:
👍

PR #539, opened 2016-07-25
Allow users to re-subscribe after cancellation

mario-galic:
@tomverran @paulbrown1982 

[Trello card](https://trello.com/c/39uNlLYm/50-users-that-cancelled-cannot-re-subscribe)


tomverran:
👍

PR #540, opened 2016-07-25
Fix bug with identity callback and update direct debit address

paulbrown1982:
- Fixed a bug where the callback URL for Sign in had a wrong countryGroup parameter and caused a 400 error.
- Updated the address for the direct debit, and tidied up the styling around that section.

cc @tomverran


tomverran:
👍

PR #543, opened 2016-08-01
Tweaks for Google Analytics ready for migration

paulbrown1982:
- Make the tracking ID configurable.
- Enable the linker.
- Added eventbrite linker (to be consistent with other membership sites).
- Used Alexis's suggested sandbox UA account when in dev mode.

cc @tomverran

### DIFF
--- a/assets/javascripts/modules/analytics/ga.js
+++ b/assets/javascripts/modules/analytics/ga.js
@@ -21,6 +21,9 @@ define(['utils/cookie',
             'cookieDomain': guardian.googleAnalytics.cookieDomain
         });
 
+        ga('require', 'linker');
+        ga('linker:autoLink', ['eventbrite.co.uk'] ); // for consistency with the rest of membership sites
+
         /**
          * Enable enhanced link attribution
          * https://support.google.com/analytics/answer/2558867?hl=en-GB
--- a/assets/javascripts/modules/analytics/ga.js
+++ b/assets/javascripts/modules/analytics/ga.js
@@ -16,9 +16,19 @@ define(['utils/cookie',
         })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
         /*eslint-enable */
 
-        ga('create', 'UA-51507017-5', 'auto');
-        ga('set', 'dimension1', identitySignedIn.toString());
-        ga('set', 'dimension2', identitySignedOut.toString());
+        ga('create', guardian.googleAnalytics.trackingId, {
+            'allowLinker': true,
+            'cookieDomain': guardian.googleAnalytics.cookieDomain
+        });
+
+        /**
+         * Enable enhanced link attribution
+         * https://support.google.com/analytics/answer/2558867?hl=en-GB
+         */
+        ga('require', 'linkid', 'linkid.js');
+
+        ga('set', 'dimension1', identitySignedIn.toString());   // deprecated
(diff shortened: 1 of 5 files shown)


conf/DEV.conf
```diff
@@ -30,3 +30,5 @@ qa {
 
 //aws.queue.welcome-email = "subs-welcome-email-dev"
 aws.queue.welcome-email = "subs-welcome-email"
+
+google.analytics.tracking.id="UA-51507017-4"
```
paulbrown1982: "Membership Dev" property

assets/javascripts/modules/analytics/ga.js
```diff
@@ -16,9 +16,22 @@ define(['utils/cookie',
...
+        ga('linker:autoLink', ['eventbrite.co.uk'] ); // for consistency with the rest of membership sites
+
+        /**
+         * Enable enhanced link attribution
+         * https://support.google.com/analytics/answer/2558867?hl=en-GB
+         */
+        ga('require', 'linkid', 'linkid.js');
+
+        ga('set', 'dimension1', identitySignedIn.toString());   // deprecated
+        ga('set', 'dimension2', identitySignedOut.toString());  // deprecated
```
cb372: Alexis explained this to me this morning. I guess you'll start sending `identitySignedIn` as a different custom dimension when we have decided the index for it?

cb372:
Looks good to me!

assets/javascripts/modules/analytics/ga.js
paulbrown1982: identitySignedOut is actually the set of people who have an identity account but are not signed in (have a GU_SO cookie and no GU_U cookie). It's not the same set as the inverse of those who are signed in (have a GU_U cookie).

paulbrown1982: This is useful to subs because it has an "Already have an account?" link which those users may or may not click. Also anyone in that set who tries to register without logging in will get a "You already have an account" error message. This scenario is obviously on the backlog for us to improve somehow.

cb372: OK, so it's not actually deprecated, but is only useful for subs? It might be worth clarifying the situation in the comment.

paulbrown1982: It's deprecated against that dimension ID - we've agreed it's fine to lose the existing data under dimension2 and start again on a new dimension number.

paulbrown1982: It is also useful for membership too - because we currently [annoyingly] redirect all non-logged-in customers to the "Create your Guardian account" Identity page not the "Sign in to the Guardian" page when they click a Join Now CTA. Grr.

paulbrown1982: Action to change the dimension names for identitySignedIn and identitySignedOut on this ticket: https://trello.com/c/3EeN2Ivg/575-ensure-omniture-google-analytics-tracking-works-ok-across-the-product-set

ajosephides: Going forward, and something I am putting in place on other properties, is that we have 2 custom dimensions that get passed through: identityID and isUserLoggedOn. Identity gets passed as a User scope and isUserLoggedOn on a hit level scope. 
In combination these should satisfy the deprecation of identitySignedOut

PR #545, opened 2016-08-04
Fix post code and delivery date selectors

paulbrown1982:
- Switch businessAdd back to normal days add, as it was disadvantaging weekend customers.
- Make the available days in the CustomDatePicker flex when the user upgrades or downgrades their package. Correcting the selected date to the first available date if it's no longer applicable.
- Added the day of the week to the 'Date of first paper' display to make it more obvious.
- Update the 'Date of first paper' on the Review details section too (with a lot of whitespace collateral)
- Fixed postcode interaction and formatting so it works same in Chrome as it does in Firefox.

cc @tomverran


tomverran:
👍

PR #546, opened 2016-08-04
Improve logging on sendETDataExtensionRow

mario-galic:
@tomverran

### DIFF
--- a/app/services/ExactTargetService.scala
+++ b/app/services/ExactTargetService.scala
@@ -50,14 +50,14 @@ trait ExactTargetService extends LazyLogging {
       validPromotion: Option[ValidPromotion[NewUsers]],
       purchaserIds: PurchaserIdentifiers): Future[DataExtensionRow] = {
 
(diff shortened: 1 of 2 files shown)


app/services/ExactTargetService.scala
```diff
@@ -83,14 +86,32 @@ trait ExactTargetService extends LazyLogging {
           promotionDescription = promotionDescription
         )
       )
+    }
+
+    for {
+      sub <- getZuoraPaidSubscription()
```
tomverran: Now you're using functions to return futures these steps won't happen concurrently, before that the futures were begun outside of the for comprehension

app/services/ExactTargetService.scala
```diff
@@ -83,14 +86,32 @@ trait ExactTargetService extends LazyLogging {
...
+      pm <- getZuoraPaymentMethod()
```
tomverran: This is also kinda reminiscent of PaymentService#getPaymentDetails

tomverran:
👍

PR #547, opened 2016-08-04
Added the billing schedule to the suspend page

paulbrown1982:
Added the billing schedule to the suspend page.
Also removed already-taken dates from the calendar so the user won't accidentally pick them.
Made the red banner show when in UAT mode.
Moved the partial templates into a fragments sub-package.
Tidied up some unused copy-pasted CSS.

cc @tomverran


tomverran:
hmm has this been incorporated into later PRs?

paulbrown1982:
I'm changing this currently, so will close till complete

PR #549, opened 2016-08-05
Fixed bug when paper sub contains digipack

paulbrown1982:
Fixed bug where flag was never true to indicate whether a paper sub contains a digipack bolt-on, and so 'instructions' part of ExactTarget email template is not being rendered.

cc @tomverran


tomverran:
👍

PR #553, opened 2016-08-05
Email billing schedule after suspend

paulbrown1982:
- Send the ExactTarget data extension JSON onto an SQS queue, using the new HolidaySuspensionBillingScheduleDataExtensionRow for formatting.
- Separate out building of data extension row from the enqueuing of the message.
- Use the new getByContactId and BillingSchedule.rollup from membership-common-0.255

cc @tomverran @mario-galic

### DIFF
--- a/app/controllers/AccountManagement.scala
+++ b/app/controllers/AccountManagement.scala
@@ -114,7 +114,7 @@ object AccountManagement extends Controller with LazyLogging {
       oldBS <- EitherT(tpBackend.commonPaymentService.billingSchedule(sub, 12).map(_ \/> "Error getting billing schedule"))
       result <- EitherT(tpBackend.suspensionService.addHoliday(newHoliday, oldBS)).leftMap(userFacingError)
       newBS <- EitherT(tpBackend.commonPaymentService.billingSchedule(sub, 12).map(_ \/> "Error getting billing schedule"))
-      newHolidays <- EitherT(tpBackend.suspensionService.getHolidays(sub.name))
+      newHolidays <- EitherT(tpBackend.suspensionService.getHolidays(sub.name)).leftMap(_ => "Error getting holidays")
       pendingHolidays = pendingHolidayRefunds(newHolidays)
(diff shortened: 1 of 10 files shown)


app/controllers/AccountManagement.scala
```diff
@@ -118,15 +118,22 @@ object AccountManagement extends Controller with LazyLogging {
       pendingHolidays = pendingHolidayRefunds(newHolidays)
```
tomverran: This wasn't your code but there's a call to getHolidays in this file that returnsString \/ Seq[Holiday] where the string is going to be some Zuora JSON we failed to deserialise, and it'll be printed out in BadRequest unless we leftMap it to something like Internal Error first

app/model/exactTarget/DataExtensionRow.scala
```diff
@@ -163,10 +167,60 @@ object PaperHomeDeliveryWelcome1DataExtensionRow {
...
+           packageName: String,
+           billingSchedule: BillingSchedule,
+           numberOfSuspensionsLinedUp: Int,
+           daysUsed: Int,
+           daysAllowed: Int
+         ): HolidaySuspensionBillingScheduleDataExtensionRow = {
+
+    val (thereafterBill, trimmedSchedule) = BillingSchedule.rolledUp(billingSchedule)
+    val count = new AtomicInteger()
+    val futureBills = trimmedSchedule.flatMap(bill => {
```
tomverran: You may find zipWithIndex helpful here - it takes a list[A] and maps it to a List[(Int, A)] where the Int is the index

paulbrown1982: Ok I'll fix that when I rebase https://github.com/guardian/subscriptions-frontend/pull/547 to master and resubmit it.

app/controllers/AccountManagement.scala
paulbrown1982: Ok I'll fix that when I rebase https://github.com/guardian/subscriptions-frontend/pull/547 to master and resubmit it.

tomverran:
👍

PR #554, opened 2016-08-05
Fix date of first paper not being 5 days ahead

paulbrown1982:
😢  Forgot to add , 'days' after changing it from addBusinessDays
cc @tomverran


tomverran:
👍 

As low latency as our delivery services are we do need slightly more than 5ms notice

PR #556, opened 2016-08-08
Billing schedule on first page

paulbrown1982:
- Put the billing schedule into its own view template, and also added the rollup logic.
- Added the template to the suspend page and the checkout page. This meant I could remove the promotion-applying-to-plan-price logic - much happier that we won't tell the customer the wrong thing now.
- Removed already-taken dates from the calendar so  the user won't accidentally pick them.
- Added a 1-year-ahead limit to the start date.
- Added the red banner to the top of the page so we are reminded what environment we are in.


tomverran:
👍

PR #558, opened 2016-08-08
Added new dimensions to Google Analytics.

paulbrown1982:
- Added new dimensions to Google Analytics.
- Made GA load only after the ophan js promise has returned (successfully or not) as it prefers to have the ophan pageViewId.
- Used the 'membership' ophan platform for now, Alex will create 'subscriptions' when he creates 'contributions'

cc @AWare @ajosephides @cb372

### DIFF
--- a/assets/javascripts/modules/analytics/ga.js
+++ b/assets/javascripts/modules/analytics/ga.js
@@ -18,32 +18,35 @@ define(['modules/analytics/analyticsEnabled',
         })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
         /*eslint-enable */
 
-        // Do JellyFish first as they don't want our custom dimensions
-        ga('create', 'UA-44575989-1', 'auto', 'jellyfishGA');
-        ga('jellyfishGA.send', 'pageview');
-
-        ga('create', guardian.googleAnalytics.trackingId, {
+        // Ours
+        ga('create', {
+            'trackingId': guardian.googleAnalytics.trackingId,
+            'name': 'subscriptionsGA',
             'allowLinker': true,
             'cookieDomain': guardian.googleAnalytics.cookieDomain
         });
 
-        ga('require', 'linker');
-        ga('linker:autoLink', ['eventbrite.co.uk'] ); // for consistency with the rest of membership sites
+        ga('subscriptionsGA.require', 'linkid');
+        ga('subscriptionsGA.require', 'linker');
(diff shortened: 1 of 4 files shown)


assets/javascripts/modules/analytics/ga.js
```diff
@@ -30,12 +36,14 @@ define(['utils/cookie',
          */
         ga('require', 'linkid', 'linkid.js');
 
-        ga('set', 'dimension1', identitySignedIn.toString());   // deprecated
```
ajosephides: Did you want to do this immediately in this first iteration?
Or can we leave it in for the time being to check against new data. Turning it off from now might affect a dashboard that is in use.

assets/javascripts/modules/analytics/ga.js
```diff
@@ -16,6 +18,10 @@ define(['utils/cookie',
         })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
         /*eslint-enable */
 
+        // Do JellyFish first as they don't want our custom dimensions
+        ga('create', 'UA-44575989-1', 'auto', 'jellyfishGA');
```
ajosephides: Since we have 2 trackers - ours and jellyfish can we explicitly call our tracker on the custom dimension sets.
This way we can have the pageview sends together at the end.

paulbrown1982: We have to make 2 separate calls they don't combine. JellyFish's tracker also redirects 3 times...

https://developers.google.com/analytics/devguides/collection/analyticsjs/creating-trackers#working_with_multiple_trackers

ajosephides: That's cool.
I didn't explain myself too well. Currently our default tracker doesn't have a name - we could it something like membershipPropertyTracker
We then set each of the custom dimensions with the relevant tracker. So in this case we would have something like:
ga('membershipPropertyTracker.set', 'dimension5', 'subscriptions');
etc...
We can then be more explicit with what goes to what tracker without having to worry about having the Jellyfish pageview send go first.
This is what is being done on Frontend.
Does that make sense?

paulbrown1982: Yup, that's a good idea.

tomverran:
👍  when everyone is happy

paulbrown1982:
cc @ajosephides I've added a name to the sub's tracker, etc.

ajosephides:
👍

PR #561, opened 2016-08-09
Removed snowplow analytics

paulbrown1982:
Removed snowplow analytics beaconing because we and the business typically use Google Analytics and Omniture for insights into subs-frontend conversion, etc, and so we (sadly) rarely refer to the subs snowplow dashboard. The opposite is the case for Membership.

cc @tomverran @nlindblad @mario-galic @dominickendrick @JuliaBellis


tomverran:
👍

PR #562, opened 2016-08-09
Updates and improvements

paulbrown1982:
Copy and SEO improvements.
Put back the JellyFish promo benefits list.
Updated curl.js to the latest (but still old) version.
cc @JuliaBellis @tomverran @nlindblad @mario-galic


nlindblad:
Nice! 👍

PR #563, opened 2016-08-10
Login link in the email lookup error message

paulbrown1982:
- Added a link to the error message which appears when the email address is already assigned to an Identity account. This should hopefully encourage users to really log in.
- Made the URL for the email validation come from reverse routing than be specified in the JavaScript file.
- Trimmed the postcode lookup to just 4 characters as that's the longest district code in the lookup - adds a bit more privacy. Also added a cache header to that response.


nlindblad:
👍

PR #565, opened 2016-08-10
More Google Analytics dimensions and better redirects

paulbrown1982:
- Added the product name, the zuora ID and the internal campaign code dimensions when relevant, to Google Analytics pageviews.
- Also fixed redirects for INTCMP which added two copies of the parameter to the query string when redirecting, so instead we just preserve all query parameters when redirecting.  All ready for the new Google Analytics 'utm_' external campaign codes.
- Changed how the homepage and /digital edition redirects work, removing the edition enumerations from the routes and controller files.

cc @ajosephides @AWare @rtyley

### DIFF
--- a/app/views/checkout/thankyou.scala.html
+++ b/app/views/checkout/thankyou.scala.html
@@ -23,11 +23,15 @@
 <script type="text/javascript">
     guardian.pageInfo.slug="GuardianDigiPack:Order Complete";
     guardian.pageInfo.productData = {
+        // Omniture - TODO - fix the type field
         source: 'Subscriptions and Membership',
         type: 'GUARDIAN_DIGIPACK',
         eventName: 'purchase',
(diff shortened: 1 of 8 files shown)


app/views/checkout/thankyou.scala.html
```diff
@@ -23,11 +23,15 @@
 <script type="text/javascript">
     guardian.pageInfo.slug="GuardianDigiPack:Order Complete";
     guardian.pageInfo.productData = {
+        // Omniture - TODO - fix the type field
```
AWare: Does this comment need to be re-added?

paulbrown1982: I need to talk to @JuliaBellis about what the enum values will be for non-digipack.

AWare:
👍

ajosephides:
👍

PR #567, opened 2016-08-11
Rolling back curl.js as the new one handles plugins differently

paulbrown1982:
My bad. I've reverted.
cc @nlindblad


nlindblad:
👍

PR #568, opened 2016-08-11
Removed internal uses of INTCMP codes

paulbrown1982:
Removed internal uses of INTCMP codes, these should be for navigating across sites only. Google Analytics has the linkid plugin installed to solve thie "which link did the customer use" problem.

cc @ajosephides @nlindblad


paulbrown1982:
I won't merge this till I've spoken to @JuliaBellis

nlindblad:
👍

paulbrown1982:
Am going to solve this another way - the same way as has been solved on dotcom.

PR #570, opened 2016-08-12
Initial work for voucher checkout page

nlindblad:
**Paper**:

screen shot 2016-08-12 at 12 51 04

**Voucher**:

screen shot 2016-08-12 at 12 51 12

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -12,6 +12,7 @@
 @import views.support.ProductPopulationData
 @import views.support.ProductPopulationData._
 @import com.gu.memsub.Current
+@import com.gu.memsub.Delivery
 @import com.gu.subscriptions.ProductPlan
 @import com.gu.i18n.Currency
 @import com.gu.i18n.Country
@@ -157,7 +158,11 @@
                                     Delivery address
                                 </legend>
                                 <div class="field-panel__fields">
-                                    @fragments.checkout.fieldsPaper(productData, countriesWithCurrency)
+                                    @if(productData.plans.list.contains(Delivery)) {
+                                        @fragments.checkout.fieldsPaper(productData, countriesWithCurrency)
+                                    }else{
+                                        @fragments.checkout.fieldsVoucher(productData, countriesWithCurrency)
+                                    }
                                     @continueButton("#paymentDetails", "js-checkout-delivery-details-submit")
                                 </div>
                                 <div class="field-panel__edit">
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -158,7 +158,7 @@
                                     Delivery address
                                 </legend>
                                 <div class="field-panel__fields">
-                                    @if(productData.plans.list.contains(Delivery)) {
+                                    @if(productData.plans.default.products.seq.contains(Delivery)) {
                                         @fragments.checkout.fieldsPaper(productData, countriesWithCurrency)
                                     }else{
                                         @fragments.checkout.fieldsVoucher(productData, countriesWithCurrency)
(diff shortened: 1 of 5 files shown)


app/views/checkout/payment.scala.html
```diff
@@ -157,7 +158,11 @@
                                     Delivery address
                                 </legend>
                                 <div class="field-panel__fields">
-                                    @fragments.checkout.fieldsPaper(productData, countriesWithCurrency)
+                                    @if(productData.plans.list.contains(Delivery)) {
```
nlindblad: @tomverran Looks like this is actually looking through all the products, not just the one currently displayed. That means it will always default to the paper fields.

I'm trying to figure out the correct way of seeing whether we're serving the voucher checkout page without resorting to looking at the slug.

app/views/checkout/payment.scala.html
```diff
@@ -157,7 +158,11 @@
                                     Delivery address
                                 </legend>
                                 <div class="field-panel__fields">
-                                    @fragments.checkout.fieldsPaper(productData, countriesWithCurrency)
+                                    @if(productData.plans.default.products.seq.contains(Delivery)) {
+                                        @fragments.checkout.fieldsPaper(productData, countriesWithCurrency)
+                                    }else{
```
paulbrown1982: This needs to be an else if otherwise it'll show for digital pack checkout.

assets/javascripts/modules/checkout/voucherFields.jsx
```diff
@@ -0,0 +1,57 @@
...
+import CustomDatePicker from '../react/customDatePicker'
+import CharacterCountedTextArea from './characterCountedTextArea'
+import reviewDetails from './reviewDetails'
+
+require('react-datepicker/dist/react-datepicker.css');
+
+const MAX_WEEKS_AVAILABLE = 4;
+
+function filterDate() {
+    return (date) => date.day() === 1;
```
paulbrown1982: This is fine for a first pass, but we should consider making it show the relevant day to the chosen package. e.g. Everyday and Sixday shows Monday. Weekend shows Saturday and Sunday+ shows Sunday.

assets/javascripts/modules/checkout/voucherFields.jsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function getFirstSelectableDate() {
+    var now = moment.utc();
+    var currentWeekday = now.weekday();
+    var currentHour = now.hour();
+    var mostRecentMonday = moment().startOf('isoWeek');
+    var weeksToAdd = currentWeekday >= 3 && currentHour >= 6 ? 3 : 2;
```
tomverran: It would also be good to break these numbers out at least into consts to explain what this check does

app/views/checkout/payment.scala.html
nlindblad: The `@if(productData.planEither.isRight)` above it (line 154) should take care of that. It was already in place to switch on digital pack vs. home delivery.

nlindblad: screen shot 2016-08-15 at 10 12 07

assets/javascripts/modules/checkout/voucherFields.jsx
nlindblad: Yeah, that might be more true to what you are actually getting, e.g. you'll see the first day you can actually use your voucher to get the paper. I'll keep it as it is and create a new PR for that specific change once this one is merged.

nlindblad:
Tested in UAT, I went through the flow for voucher locally and I can confirm it shows up correctly in Zuora with the picked date.

paulbrown1982:
👍

PR #571, opened 2016-08-12
Support supplier codes to facilitate inbound channel tracking.

paulbrown1982:
Changes required to handle supplier codes to facilitate inbound channel tracking.

The idea is that the call centre would bookmark: https://subscribe.theguardian.com/s/:supplierCode to redirect to the homepage, or have a catalog of deep links to the checkout: https://subscribe.theguardian.com/checkout?supplierCode=:supplierCode  and then all subscriptions made by them would be tracked against their supplier code as the code would be stored in their browser session. 

- The query parameter or path value always wins over the session.
- This functionality works irrespective to promo codes and tracking promo codes which are for outbound channel tracking.
- The change requires Zuora to add a custom field called SupplierCode__c and for it to get replicated everywhere like the PromotionCode__c custom field.
- I also tidied up the error messages in Binders.scala and renamed suppliedPromoCode to just promoCode in SubscriptionData.scala

### DIFF
--- a/app/controllers/Binders.scala
+++ b/app/controllers/Binders.scala
@@ -1,7 +1,7 @@
 package controllers
 
 import com.gu.i18n.{Country, CountryGroup}
-import com.gu.memsub.ProductFamily
+import com.gu.memsub.{ProductFamily, SupplierCode, SupplierCodeBuilder}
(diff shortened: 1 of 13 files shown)


tomverran:
👍

paulbrown1982:
Will merge on Monday.

PR #573, opened 2016-08-12
Delivery mma styles

tomverran:
This brings the delivery MMA page closer to the voucher MMA page in style

image

I took this screenshot on a vertical monitor hence the billing schedule overflow


nlindblad:
👍

PR #575, opened 2016-08-15
Hardcode Node to 6.2.1 for build

nlindblad:
Makes the build more repeatable and we know this version exists. Ping @paulbrown1982 and @tomverran


tomverran:
👍

PR #576, opened 2016-08-15
Adding prefix to supplier redirect so we can filter in GA

paulbrown1982:
This helps us identify requests that have come from the supplier redirect, by making the campaign codes have a prefix. This is the same as what we do for the non-landing-page promotions redirect.

cc @nlindblad @tomverran


nlindblad:
👍

PR #578, opened 2016-08-15
Stripe TLSv1.2 double check and tidy up

paulbrown1982:
- Updated to latest membership common to pick up the latest 2.x version of OkHTTP (2.7.2 upgrading to [2.7.5](https://github.com/square/okhttp/blob/master/CHANGELOG.md)), as part of the Stripe TLSv1.2 check. We won't update to OkHTTP v3x yet.
- Also updated the OkHTTP dependency used by the acceptance test suite to the same, and ensured it was brought in to the test scope only.
- Finally, removed other unused dependencies and config - these were used by the snowplow analytics which was removed earlier.

cc @tomverran @nlindblad


nlindblad:
👍

PR #581, opened 2016-08-16
Move plan sensitive date filter to its own module and use it for voucher

nlindblad:
Voucher now adheres to showing the first day you could actually get the paper using your voucher:

screen shot 2016-08-16 at 14 10 36
screen shot 2016-08-16 at 14 09 43


paulbrown1982:
👍 Nice!

PR #582, opened 2016-08-16
Remove the DLA special copy

paulbrown1982:
Just realised in this mode it was not adjusting the quote down on screen - the alternate bit is only applicable for incentive promotions. The promotion is still getting applied correctly to the sub.

cc @tomverran @JuliaBellis


tomverran:
👍

PR #584, opened 2016-08-17
This should be the net value available for you to use, not the total you're allowed to take.

paulbrown1982:
picture 173

cc @tomverran


tomverran:
I am _loving_ the descriptiveness of the screenshots here!
👍

PR #586, opened 2016-08-18
Fix section payment jump

nlindblad:
This fixes the bug where pressing enter at the beginning of the collapsed form shows they payment section, which is a very confusing behaviour that lead users astray.

A more long term solution is to let enter trigger the partial validation for each form section (the equivalent of clicking on "Continue"), but the way the form is currently laid out, there is not enough information about how the sections fit together to do that nicely. Let's discuss how we might improve this.


tomverran:
👍

paulbrown1982:
👍 +cc @mario-galic

mario-galic:
👍

PR #588, opened 2016-08-19
Copy tweaks as Sunday+ customers have £1 delivery not 50p

paulbrown1982:
Also removed some unused imports.
cc @tomverran @nlindblad @JuliaBellis


tomverran:
👍

PR #591, opened 2016-08-25
Improve instructions for setup

johnduffell:
@nlindblad @tomverran


nginx/setup.sh
```diff
@@ -12,4 +14,5 @@ aws s3 cp s3://identity-local-ssl/sub-thegulocal-com-exp2017-03-31.key ${GU_KEYS
 sudo ln -fs ${GU_KEYS}/ $NGINX_HOME/keys
 
 sudo nginx -s stop
+sleep 1
```
johnduffell: the above is async so we need to give it time

tomverran:
👍

PR #593, opened 2016-08-30
Polish review and confirm

nlindblad:
- Hide *Delivery instructions* in case of voucher.

- Fix JS to mirror e-mail correctly in *Review and confirm* when the user is logged in.


tomverran:
:+1:

PR #596, opened 2016-09-05
Add different date filtering for voucher and home delivery

nlindblad:
6 and 7 day subscriptions can start on Mondays and weekend has to start on Saturdays.

**Voucher Weekend**:
screen shot 2016-09-05 at 15 23 23

**Voucher Everyday and Sixday**:
screen shot 2016-09-05 at 15 23 08


tomverran:
👍

PR #600, opened 2016-09-12
membership common mk4

tomverran:
This does the following

- Rewires subs to use the new subscription / catalog models
- Adds a product catalog check to the healthcheck to match membership
- Turns TouchpointBackend into a trait of lazy vals - no more `ExceptionInInitializer` errors
- reduces lines of code!


app/views/account/voucher.scala.html
```diff
@@ -28,7 +27,7 @@ <h3 class="mma-section__header">
         <h3 class="mma-section__header">
           Your billing schedule
         </h3>
-        @views.html.account.fragments.billingSchedule(billingSchedule, subscription.currency)
+        @views.html.account.fragments.billingSchedule(billingSchedule, subscription.plan.charges.price.currencies.head)
```
tomverran: The difficulty of safely digging a currency out of a paid subscription plan is quite annoying but its something that can be sorted out in the future I think

tomverran:
@johnduffell what better to take your mind off that last pull request than another very similar one? 😄

johnduffell:
thanks, I'll take a look when I run out of other material to read, presently I'm looking at this: https://www.theguardian.com/info/2014/jul/10/guardian-weekly-print-subscription-services-terms-conditions 😆

app/actions/ActionRefiners.scala
```diff
@@ -1,29 +0,0 @@
...
-import scalaz.OptionT
-import org.joda.time.DateTime
-
-
-object ActionRefiners {
-
-  def noSubscriptionAction(onSubscription: Subscription => Result): ActionFilter[Request]  = new ActionFilter[Request] {
-    override def filter[A](request: Request[A]): Future[Option[Result]] = {
-      authenticatedUserFor(request).fold[Future[Option[Result]]] {Future.successful(None)} { user =>
-        val tp = TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)(request).backend
```
tomverran: So this is what we used to use to detect duplicate digipack subs if you were signed in

But for paper we use the same checkout so this is no longer good, we don't want to kick you out just because you're subscribed to digipack

app/controllers/AccountManagement.scala
```diff
@@ -71,19 +73,20 @@ object AccountManagement extends Controller with LazyLogging {
...
+      sub = subEither.fold(identity, identity)
       allHolidays <- OptionT(tpBackend.suspensionService.getHolidays(sub.name).map(_.toOption))
-      billingSchedule <- OptionT(tpBackend.commonPaymentService.billingSchedule(sub, 12))
-      suspendableDays = Config.suspendableWeeks * sub.plan.products.without(Delivery).size
-    } yield {
-      sub.plan.products.seq.contains(Delivery).fold({
+      billingSchedule <- OptionT(tpBackend.commonPaymentService.billingSchedule(sub.id, numberOfBills = 12))
+      suspendableDays = Config.suspendableWeeks * sub.plan.charges.benefits.size
+    } yield
+      subEither.fold({ s =>
```
johnduffell: `s` should be `deliverySubscription`

app/controllers/AccountManagement.scala
```diff
@@ -71,19 +73,20 @@ object AccountManagement extends Controller with LazyLogging {
...
         val pendingHolidays = pendingHolidayRefunds(allHolidays)
-        val suspendedDays = SuspensionService.holidayToSuspendedDays(pendingHolidays, sub.plan.products.physicalProducts.list)
-        Ok(views.html.account.suspend(sub, pendingHolidayRefunds(allHolidays), billingSchedule, suspendableDays, suspendedDays, errorCodes))
-      },{
-        Ok(views.html.account.voucher(sub, billingSchedule))
+        val suspendedDays = SuspensionService.holidayToSuspendedDays(pendingHolidays, s.plan.charges.days.toList)
+        Ok(views.html.account.suspend(s, pendingHolidayRefunds(allHolidays), billingSchedule, suspendableDays, suspendedDays, errorCodes))
+      },{ s =>
```
johnduffell: `s` should be `voucherSubscription`

app/controllers/AccountManagement.scala
```diff
@@ -71,19 +73,20 @@ object AccountManagement extends Controller with LazyLogging {
...
     (for {
-      sub <- OptionT(subscription)
+      subEither <- OptionT(subscription)
+      sub = subEither.fold(identity, identity)
       allHolidays <- OptionT(tpBackend.suspensionService.getHolidays(sub.name).map(_.toOption))
-      billingSchedule <- OptionT(tpBackend.commonPaymentService.billingSchedule(sub, 12))
-      suspendableDays = Config.suspendableWeeks * sub.plan.products.without(Delivery).size
-    } yield {
-      sub.plan.products.seq.contains(Delivery).fold({
+      billingSchedule <- OptionT(tpBackend.commonPaymentService.billingSchedule(sub.id, numberOfBills = 12))
```
johnduffell: good use of a named parameter

app/controllers/AccountManagement.scala
tomverran: seems github doesn't hide comments on outdated diffs now, but I fixed this in 4cf5501

app/controllers/Checkout.scala
```diff
@@ -42,6 +43,9 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
   def checkoutService(implicit res: TouchpointBackend.Resolution): CheckoutService =
     res.backend.checkoutService
 
+  def getBetterPlans[A <: CatalogPlan.Paid](plan: A, others: List[A]) =
```
johnduffell: better for who? 😛

johnduffell: should that be more expensive?

app/controllers/Checkout.scala
```diff
@@ -218,14 +223,11 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
     implicit val tpBackend = resolution.backend
-
-    import tpBackend.catalogService._
-
     val session = request.session
 
     val sessionInfo = for {
       subsName <- session.get(SubsName)
-      plan <- session.get(RatePlanId).map(ProductRatePlanId).flatMap(p => paperCatalog.find(p))
+      plan <- session.get(RatePlanId).map(ProductRatePlanId).flatMap(p => catalog.allSubs.flatten.find(_.id == p))
```
johnduffell: I'm surptised that ProductRatePlanId works without a .apply on the end

app/controllers/Checkout.scala
```diff
@@ -260,18 +262,19 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
-          ratePlanPrice.ratePlanId.get -> ratePlanPrice.plan.prettyPricingForDiscountedPeriod(discountPromo, currency)
+          ratePlanPrice.ratePlanId.get -> ratePlanPrice.chargeList.prettyPricingForDiscountedPeriod(discountPromo, currency)
         }.toMap
       }
     }
 
     tpBackend.promoService.findPromotion(promoCode).fold {
       NotFound(Json.obj("errorMessage" -> s"Sorry, we can't find that code."))
     } { promo =>
+      println(promo.appliesTo, prpId)
```
johnduffell: 😆

app/controllers/Checkout.scala
```diff
@@ -260,18 +262,19 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
     import views.support.Pricing._
 
     def getAdjustedRatePlans(promo: AnyPromotion): Option[Map[String, String]] = {
-      case class RatePlanPrice(ratePlanId: ProductRatePlanId, plan: DigipackPlan[BillingPeriod])
+      case class RatePlanPrice(ratePlanId: ProductRatePlanId, chargeList: PaidChargeList)
       promo.asDiscount.map { discountPromo =>
-        catalog.planMap.map { case (key, value) => RatePlanPrice(key, value) }.map { ratePlanPrice =>
+        catalog.allSubs.flatten.map(p => RatePlanPrice(p.id, p.charges)).map { ratePlanPrice =>
```
johnduffell: p is a ratePlan I think?

app/controllers/Checkout.scala
tomverran: a more expensive plan is better for everyone!

app/controllers/Checkout.scala
tomverran: oh gosh

app/controllers/Checkout.scala
tomverran: it is, though I kinda take the view that if the scope of a variable is as short as this one it is reasonable to use short names

johnduffell: I guessed, actually it's not clear what comes out of the flattening otherwise I'd agree

tomverran: ah fair enough, I shall make it `plan`

app/controllers/Management.scala
```diff
@@ -40,10 +43,23 @@ object Management extends Controller with LazyLogging {
     override def ok = exec()
   }
 
+  class EitherTest[A](name: String, exec: () => List[String] \/ A) extends Test {
+    override def messages = exec().swap.toOption.toList.flatten
+    override def ok: Boolean = exec().isRight
```
johnduffell: do you really want it to do the exec twice if it fails? might need to put the result of exec into a lazy val to avoid reevaluating it

johnduffell: or return the whole thing as an object and use this as a factory, if it does need reevaluating every time it's used as a whole

tomverran: so I didn't want to really, but the idea of these tests is that they get run repeatedly to determine whether the instance is up, so a lazy val result wouldn't work too well

I guess maybe making the EitherTest return some ADT of Okay or Errors might make more sense

tomverran: oh or a scalaz.validation so they can be tested concurrently!

app/controllers/PromoLandingPage.scala
```diff
@@ -27,14 +27,17 @@ object PromoLandingPage extends Controller {
...
+      else
+        Some(views.html.promotion.landingPage(edition, catalog.digipack.month, promoCode, promotionWithLandingPage, Config.Zuora.paymentDelay, PegdownMarkdownRenderer))
     } yield Ok(html)).getOrElse(Redirect("/digital" ? ("INTCMP" -> s"FROM_P_${promoCode.get}")))
   }
 
   def preview() = GoogleAuthenticatedStaffAction { implicit request =>
     Form(Forms.single("promoJson" -> Forms.text)).bindFromRequest().fold(_ => NotFound, { jsString =>
       Json.fromJson[AnyPromotion](Json.parse(jsString)).asOpt.flatMap(_.asDigipack)
-      .map(p => views.html.promotion.landingPage(edition, catalog, p.codes.headOption.getOrElse(PromoCode("")), p, Config.Zuora.paymentDelay, PegdownMarkdownRenderer))
+      .map(p => views.html.promotion.landingPage(edition, catalog.digipack.month, p.codes.headOption.getOrElse(PromoCode("")), p, Config.Zuora.paymentDelay, PegdownMarkdownRenderer))
```
johnduffell: there's a lot on one line, perhaps we can `val` up some of the parameters to make it neater?

tomverran: yeah I do agree though I'm a bit reluctant to make too many changes to code in this PR lest any issues with the refactor in general be hidden by issues introduced in this PR

app/views/support/Pricing.scala
```diff
@@ -4,17 +4,17 @@ import com.gu.i18n.{Currency, GBP}
...
 
-  implicit class PlanWithPricing(plan: PaidPlan[Status, BP]) {
-    lazy val gbpPrice = plan.pricing.getPrice(GBP).get
+  implicit class PlanWithPricing[+A <: PaidChargeList](plan: A) {
+    lazy val gbpPrice = plan.price.getPrice(GBP).get
 
-    def unsafePrice(currency: Currency) = plan.pricing.getPrice(currency).getOrElse(
-      throw new NoSuchElementException(s"Could not find a price in $currency for plan ${plan.name}")
+    def unsafePrice(currency: Currency) = plan.price.getPrice(currency).getOrElse(
+      throw new NoSuchElementException(s"Could not find a price in $currency for charge list")
```
johnduffell: not sure NoSuchElementException is that useful here, but it doesn't harm things

johnduffell:
have a merge and see what happens! 👍 🐎

PR #602, opened 2016-09-13
More checkout polishing

nlindblad:


app/views/fragments/checkout/reviewDetails.scala.html
```diff
@@ -34,7 +34,7 @@
             <div class="review-details__list">
                 <ul class="o-unstyled-list">
                     <li class="js-checkout-review-delivery-address sessioncamhidetext"></li>
-                    <li><span class="review-details__detail">Week of first paper:</span> <span class="js-checkout-review-delivery-start-date"></span></li>
+                    <li><span class="review-details__detail">Voucher book start date:</span> <span class="js-checkout-review-delivery-start-date"></span></li>
```
tomverran: Hmm does this get used for delivery too?

tomverran: it does not!

tomverran:
👍

app/views/support/PlanOps.scala
```diff
@@ -50,7 +50,7 @@ object PlanOps {
 
     def hasDigitalPack: Boolean = in.products.seq.contains(Digipack)
 
-    def phone: String = "+44 (0) 330 333 6767"
+    def phone: String = "0330 333 6767"
```
paulbrown1982: There are multiple instances of the phone number, throughout a multitude of templates and view classes. We should probably write a story to tidy that up into one helper class with the methods: phoneInternational and phoneUk

nlindblad: I created a card for it.

paulbrown1982:
👍

PR #605, opened 2016-09-15
Revert "membership common mk4"

tomverran:
Creating this just in case!

Reverts guardian/subscriptions-frontend#600


tomverran:
I dare to dream this is actually okay

PR #606, opened 2016-09-15
Fix voucher people subscribing to delivery

tomverran:
I _knew_ there would be something


johnduffell:
👍 🚤

paulbrown1982:
Would have been portentous if there wasn't anything.

PR #612, opened 2016-09-19
when calculating promotion pricing only look at the plans it applies to

tomverran:
This does fix https://sentry.io/the-guardian/subscriptions/issues/159796709/ because paper plans are getting looked at when digipack promotions get applied and they're missing some currencies


nlindblad:
👍

PR #613, opened 2016-09-19
Track google events after action complete

paulbrown1982:
The previous pull request beaconed Google Analytics events at impression time. This did not make sense when compared to a similar action with changing rate plans, and also in relation to the surrounding text on the funnel page. Therefore, to improve this event tracking further, I:

- Renamed the tracking.js module to be impressionTracking.js which is more aligned with its meaning. The google 'events' will be taken out of this file and put into their own module.
- Created a new 'eventTracking' module which beacons "completed" event actions to Google Analytics. Added the function calls to all the relevant places: e.g. (Continue ->) button event handlers.
- Removed the Google event action call from the impressionTracking module - that now works like it used to, only beaconing 'page view events' to Omniture.
- Added a new custom metric (metric2) to the Google Analytics events which is the elapsed number of seconds, the user took from page impression, to the time it took to complete that section.
- Added a new custom dimension (dimension13) which indicates whether the user is a customer service agent.

cc @ajosephides @johnduffell @tomverran @nlindblad @JuliaBellis


assets/javascripts/modules/checkout/eventTracking.js
```diff
@@ -0,0 +1,36 @@
+define(['modules/analytics/ga', 'modules/checkout/ratePlanChoice'], function (ga, ratePlanChoice) {
+    'use strict';
+
+    var STARTED = Date.now();
+
+    function getElapsedTimeMs() {
```
ajosephides: Does this not return time in seconds not milliseconds? If so should the function name be altered to reflect this?

paulbrown1982: Updated to getElapsedSeconds()

tomverran:
👍

PR #614, opened 2016-09-19
More confirmation page polishing

nlindblad:
- Hide phone number field if phone number is not present:

screen shot 2016-09-19 at 16 13 58
screen shot 2016-09-19 at 16 13 21


tomverran:
👍

PR #616, opened 2016-09-21
Use the new ACM certificate for the subs ELB

paulbrown1982:
- Use the new ACM certificate for the subs ELB
- Also use the new Fn:Sub function!
@nlindblad @tomverran @johnduffell


johnduffell:
great! much smarter than concatenation... 👍 🚀

PR #629, opened 2016-10-07
Removing SessionCam trial code

paulbrown1982:
- Removing SessionCam trial code, and field obfuscation classes. It was only ever a trial.

cc @JuliaBellis @AWare


app/views/fragments/checkout/reviewDetails.scala.html
```diff
@@ -10,9 +10,9 @@
         <span class="review-details__heading">Your details</span>
         <div class="review-details__list">
```
JuliaBellis: LGTM :see_no_evil: JB

PR #634, opened 2016-10-12
StripeCheckout #1 [for review only] 

AWare:
Here's a first look at getting stripe checkout working with subscriptions. 
image
image
The stripe checkout flow is shown when the user clicks continue from the payment stage. 

Remaining: 

- [ ] UX review
- [ ] Guardian logo
- [ ] Repeat payment testing
- [ ] Button text
- [ ] Really thorough testing


AWare:
---> #641

PR #641, opened 2016-10-18
Stripe checkout

AWare:
Stripe checkout for review. 

todo: 
- [ ] ~~don't show continue button for card?~~
- [x] modify acceptance tests
- [x] push membership-common changes
- [x] rebase 😔


mandr8:
+1

app/views/checkout/payment.scala.html
```diff
@@ -199,7 +211,7 @@
                 </div>
 
                 @* ===== Payment ===== *@
-                <div id="paymentDetails" class="field-panel is-collapsed js-fieldset-payment-details">
+                <div id="paymentDetails" class="field-panel is-collapsed js-fieldset-payment-details" @if(stripeCheckoutTest){data-country="GB"}>
```
paulbrown1982: This might have to be reworked in the Guardian Weekly checkout page.

assets/javascripts/modules/analytics/ga.js
```diff
@@ -6,6 +6,8 @@ define(['modules/analytics/analyticsEnabled',
     'use strict';
 
     var _EVENT_QUEUE = [];
+    var experienceIsSet = 'stripeCheckout' in guardian;
```
paulbrown1982: I think these should go into init() as that's guaranteed to be called after domcontentloaded, not executed at parse time like this function might be. There might be a race condition if other javascript set 'stripeCheckout' (even though Scala sets it in the template)

assets/javascripts/modules/checkout/stripeCheckout.js
```diff
@@ -0,0 +1,49 @@
+/* global guardian, StripeCheckout */
+
+define(['bean',    'modules/checkout/formElements'
+],function(bean,formElements){
+    function openCheckout() {
+        return new Promise(function (resolve, reject) {
+            var handler;
+            if (!('StripeCheckout' in window)) {
+                console.error('handle this with raven');
+                //TODO: raven
```
paulbrown1982: Are you going to create a Trello ticket to cover this TODO?

app/views/fragments/ABTest.scala
```diff
@@ -0,0 +1,19 @@
...
+import scala.util.Random
+
+object ABTest {
+  val TestIdCookieName = "gu.subscriptions.test.id"
+
+  def testIdFor[A](implicit request: Request[A]): Int = {
+
+    request.getQueryString("stripe").flatMap((str) => if (str == "checkout") {Some(2)} else {Some(1)}).
+      getOrElse(((request.cookies.get(TestIdCookieName) map (_.value.toInt)).
+        getOrElse(Random.nextInt(Int.MaxValue))))
```
paulbrown1982: If you use 0 and 1 for on/off then you can do Random.nextInt(2) here and be guaranteed to get a number back that you can support.

def nextInt(n : Int) : Int

    Returns a pseudorandom, uniformly distributed int value between 0 (inclusive) and the specified value (exclusive), drawn from this random number generator's sequence.

test/acceptance/pages/Checkout.scala
```diff
@@ -166,4 +184,64 @@ case class Checkout(testUser: TestUser, endpoint: String = "checkout") extends P
...
+    val container = name("stripe_checkout_app")
+    val cardNumber = id("card_number")
+    val cardExp = id("cc-exp")
+    val cardCvc = id("cc-csc")
+    val submitButton = id("submitButton")
+
+    def setInTest(inTest: Boolean) = {
+
+      Driver.addCookie(ABTest.TestIdCookieName, if (inTest) {
+        "2"
```
paulbrown1982: See 0/1 comment about Random.nextInt(2) below

paulbrown1982:
👍 🚀 🚢

PR #644, opened 2016-10-20
Refresh promotions faster

paulbrown1982:
Updating to the latest membership-common in order to refresh promotions faster - every 15 seconds instead of every 5 minutes. Membership will remain at 5 minutes (the DynamoPromoCollection empty-argument default) for now.

I think we have enough head room to go straight for 15s. If not I can double capacity, because it's not too expensive at the moment. 

picture 216

picture 217

cc @jacobwinch


paulbrown1982:
Things look fine after the change from 5 mins to 15 seconds. I also lowered the write credit amount from 5 to 2. The CPU on the subscriptions-frontend boxes hasn't got high either.

picture 224

picture 225

PR #650, opened 2016-10-26
Billing changes

pvighi:
changes: 
- all products now accept billing addresses from any country
- direct debit now depends on the billing address being in the UK and the currency being GBP (even if the delivery address is not in the UK)
- "pay in £GBP" checkbox is now ~~hidden~~ disabled if the user got past the billing address step to avoid problems where the currency change affects the availability of direct debit


app/views/fragments/checkout/fieldsPersonal.scala.html
```diff
@@ -1,10 +1,8 @@
-@import views.support.CountryOps._
-@import views.support.AddressValidationRulesOps._
 @import views.support.CountryWithCurrency
 @import com.gu.i18n.Title
 @import model.PersonalData
 
-@(form: Option[PersonalData], countriesWithCurrency: List[CountryWithCurrency])
+@(form: Option[PersonalData])
```
pvighi: countriesWithCurrency didn't seem to be used for anything here so I removed it

assets/javascripts/modules/checkout/fieldSwitcher.js
```diff
@@ -4,63 +4,127 @@ define([
...
-    var overrideCurrency = function(currency) {
-        latestCurrencyAddressModel.currency = currency;
-        switchCurrency(latestCurrencyAddressModel);
+
+    var redrawCurrencyOverride = function (currentState) {
+        $('.js-currency-override-checkbox').attr('checked', false);
+
+        var currencySelector = $('.js-currency-override-label, .js-currency-selector');
+
+        if (currentState.localization.currency == 'USD') {
```
paulbrown1982: Note, way may change this to if (currentState.localization.currency != 'GBP') in future if we want to allow AUS, NZA and EUR customers to pay in GBP too. CAD customers are already currently excluded due to being in the other zone.

assets/javascripts/modules/checkout/fieldSwitcher.js
```diff
@@ -91,82 +154,75 @@ define([
...
-                currency: currentCountryOption.attr('data-currency-choice'),
-                country: currentCountryOption.val(),
-                ratePlanId: ratePlanChoice.getSelectedRatePlanId()
+                currency: currency,
+                country: currentCountryOption.val()
             };
         };
 
-        var updateGuardianPageInfo = function(model) {
+        var updateGuardianPageInfo = function (currentState) {
```
paulbrown1982: This can probably be deleted as it's only used by Omniture.

app/model/PaymentValidation.scala
```diff
@@ -1,12 +1,12 @@
...
 
-import com.gu.i18n.Country
+import com.gu.i18n.{Country, Currency, GBP}
 
 object PaymentValidation {
   def validateDirectDebit(subscriptionData: SubscriptionData): Boolean = {
     subscriptionData.paymentData match {
       case _: DirectDebitData =>
-        subscriptionData.personalData.address.country.contains(Country.UK)
+        subscriptionData.personalData.address.country.contains(Country.UK) && subscriptionData.currency == GBP
```
paulbrown1982: Nice fix!

app/controllers/Checkout.scala
```diff
@@ -92,44 +92,47 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
       def renderCheckoutFor(planList: PlanList[CatalogPlan.ContentSubscription]) = {
 
         val supportedCurrencies = planList.default.charges.price.currencies
+        val billingCountriesWithCurrencyOrGDP =  CountryWithCurrency.whitelisted(supportedCurrencies, GBP)
```
paulbrown1982: s/billingCountriesWithCurrencyOrGDP/billingCountriesWithCurrencyOrGBP/g

PR #653, opened 2016-10-27
Promo terms and conditions

paulbrown1982:
- Added a new landing page type under /p/<promo_code> for newspaper subscription promotions (not a very pretty one - it just reuses the grey bars from the standard catalog pages - I had to break apart some 'shipping' templates to reuse them)
- Added a new /p/[promo_code]/terms endpoint which shows a summary of the promotion and what products it applies to,  including the new long terms and conditions. The page will remain permanent and not redirect when the promotion expires or hasn't started yet.
- Lots of refactoring of the terms and conditions sections, ready for how they want to manage the new process. Nothing removed, just sentences shifted around a bit or re-titled.
- Eventually we can add previewing capability to the newspaper landing page, but that's deliberately absent right now.
- There's a bit of collateral from membership-common's renaming PromoWithDigipackLandingPage to PromoWithDigitalPackLandingPage

picture 237

cc @johnduffell @pvighi @AWare @jacobwinch @jayceb1

### DIFF
--- a/app/controllers/PromoLandingPage.scala
+++ b/app/controllers/PromoLandingPage.scala
@@ -21,17 +21,16 @@ object PromoLandingPage extends Controller {
   val catalog = tpBackend.catalogService.unsafeCatalog
   val edition = DigitalEdition.UK
 
-  def render(promoCodeStr: String) = CachedAction { implicit request =>
+  def render(promoCodeStr: String) = CachedAction { _ =>
     val evaluateStarts = tpBackend.environmentName == "PROD"
     val promoCode = PromoCode(promoCodeStr)
 
     (for {
       promotion <- tpBackend.promoService.findPromotion(promoCode)
-      promotionWithLandingPage <- promotion.asDigipack
-      html <- if ((evaluateStarts && promotionWithLandingPage.starts.isAfterNow) || promotionWithLandingPage.expires.exists(_.isBeforeNow))
-        None
+      html <- if (((evaluateStarts && promotion.starts.isBeforeNow) || promotion.expires.exists(_.isAfterNow)) && promotion.asDigipack.isDefined)
+        Some(views.html.promotion.landingPage(edition, catalog.digipack.month, promoCode, promotion.asDigipack.get, Config.Zuora.paymentDelay, PegdownMarkdownRenderer))
       else
-        Some(views.html.promotion.landingPage(edition, catalog.digipack.month, promoCode, promotionWithLandingPage, Config.Zuora.paymentDelay, PegdownMarkdownRenderer))
+        Some(views.html.promotion.termsPage(promoCode, promotion, PegdownMarkdownRenderer))
     } yield Ok(html)).getOrElse(Redirect("/" ? (internalCampaignCode -> s"FROM_P_${promoCode.get}")))
   }
 
@@ -42,4 +41,5 @@ object PromoLandingPage extends Controller {
       .fold[Result](NotFound)(p => Ok(p).withHeaders(HandleXFrameOptionsOverrideHeader.HEADER_KEY -> s"ALLOW-FROM ${Config.previewXFrameOptionsOverride}"))
     })
   }
+
 }
--- a/app/controllers/PromoLandingPage.scala
+++ b/app/controllers/PromoLandingPage.scala
@@ -1,19 +1,19 @@
 package controllers
 import actions.CommonActions._
-import com.gu.memsub.promo._
 import com.gu.memsub.promo.Formatters.PromotionFormatters._
 import com.gu.memsub.promo.Promotion._
+import com.gu.memsub.promo._
 import com.netaporter.uri.dsl._
 import configuration.Config
 import filters.HandleXFrameOptionsOverrideHeader
 import model.DigitalEdition
-
 import play.api.data.{Form, Forms}
 import play.api.libs.json.Json
 import play.api.mvc._
+import play.twirl.api.Html
 import services.TouchpointBackend
-import views.support.PegdownMarkdownRenderer
 import utils.Tracking.internalCampaignCode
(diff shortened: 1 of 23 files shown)


app/controllers/PromoLandingPage.scala
```diff
@@ -1,45 +1,66 @@
...
     } yield Ok(html)).getOrElse(Redirect("/" ? (internalCampaignCode -> s"FROM_P_${promoCode.get}")))
   }
 
   def preview() = GoogleAuthenticatedStaffAction { implicit request =>
     Form(Forms.single("promoJson" -> Forms.text)).bindFromRequest().fold(_ => NotFound, { jsString =>
-      Json.fromJson[AnyPromotion](Json.parse(jsString)).asOpt.flatMap(_.asDigipack)
-      .map(p => views.html.promotion.landingPage(edition, catalog.digipack.month, p.codes.headOption.getOrElse(PromoCode("")), p, Config.Zuora.paymentDelay, PegdownMarkdownRenderer))
-      .fold[Result](NotFound)(p => Ok(p).withHeaders(HandleXFrameOptionsOverrideHeader.HEADER_KEY -> s"ALLOW-FROM ${Config.previewXFrameOptionsOverride}"))
+      Json.fromJson[AnyPromotion](Json.parse(jsString)).asOpt.flatMap(_.asDigitalPack)
+        .map(p => views.html.promotion.digitalpackLandingPage(edition, catalog, p.codes.headOption.getOrElse(PromoCode("")), p, PegdownMarkdownRenderer))
```
paulbrown1982: Yup, I've created a story to tackle that after - the Acquisition team are happy with that.

https://trello.com/c/6cDjeShh/62-enable-previewing-for-newspaper-landing-page

app/views/fragments/promotion/subscriptionTermsAndConditions.scala.html
```diff
@@ -1,18 +1,28 @@
...
-  @promotion.asFreeTrial.fold(defaultFreeTrialDays.getDays.toString)(_.promotionType.duration.getDays.toString)
-  days from receipt of subscriber ID, up to and including the day before your first payment falls due. At the end of the free trial,
-  the subscription is charged at @promotion.asIncentive.map { p => standard price (currently @formatPrice(catalog, promotion) a month) }
-  @promotion.asDiscount.map { p => the special price of @formatPrice(catalog, promotion) a month } unless cancelled.
-  Requires Internet connection (additional charges may apply) and an Apple, Android or Kindle Fire device.
-  For full details of the digital pack and its terms and conditions - see
-  <a class="u-link" href="//www.theguardian.com/digital-subscriptions-terms-conditions" target="_blank">theguardian.com/digital-subscriptions-terms-conditions</a>.
-</p>
+@(catalog: Catalog, promotion: AnyPromotion)
+@promotion.asDigitalPack.fold {
```
johnduffell: I prefer not to use fold on options espectially if we're not using the content of it, I'd find it clearer to use `if (option.isDefined) {something} else {something}` then new people can read more easily what's going on (and it is more clear that it's an option rather than a list etc.
If we are using the content of the option, another solution is to do a map{}.getOrElse{} although even that is less clear.  Or a pattern match would be really clear but in a play template might just fail to compile!

app/views/promotion/newspaperLandingPage.scala.html
```diff
@@ -0,0 +1,92 @@
...
+@import com.gu.memsub.subsv2.Catalog
+@import model.Subscriptions._
+
+@(catalog: Catalog, promoCode: PromoCode, promotion: PromoWithNewspaperLandingPage, md: MarkdownRenderer)
+@voucherPlans = @{catalog.voucher.list.filter(p => promotion.appliesTo.productRatePlanIds.contains(p.id))}
+@deliveryPlans = @{catalog.delivery.list.filter(p => promotion.appliesTo.productRatePlanIds.contains(p.id))}
+@anyPromotion = @{asAnyPromotion(promotion)}
+@title = @{"Subscribe to the Guardian and Observer"}
+@hash = @{s"#${promotion.landingPage.defaultProduct}"}
+@products = @{
```
johnduffell: that's a lot of code to inline into a template, I'd be very tempted to pull it out into an object so it can compile separately

app/views/support/LandingPageOps.scala
```diff
@@ -35,6 +39,24 @@ object LandingPageOps {
     def getIncentiveTermsAndConditions: String = {
       promotion.asIncentive.fold("") { p => p.promotionType.termsAndConditions.mkString }
     }
+    def getIncentiveLegalTerms: String = {
+      promotion.asIncentive.fold("") { p => p.promotionType.legalTerms.mkString }
```
johnduffell: would it be clearer:
`promotion.asIncentive.map(_.promotionType.legalTerms.mkString).getOrElse("")`

app/views/support/LandingPageOps.scala
```diff
@@ -35,6 +39,24 @@ object LandingPageOps {
...
+    }
   }
 
+  def planToOptions(promoCode: PromoCode, promotion: AnyPromotion)(in: CatalogPlan.Paid): SubscriptionOption = {
+    val planPrice = promotion.asDiscount.fold(in.charges.gbpPrice)(adjustPrice(in, _))
+    val saving = promotion.asDiscount.fold(in.saving)(_ => None)
```
johnduffell: if isdefined then None else Some(in.saving) ? so saving happens when there isn't a discount?

johnduffell:
👍 great work! 🐎

app/views/promotion/newspaperLandingPage.scala.html
paulbrown1982: I've added this as a task to the preview story which will revisit this file.

PR #659, opened 2016-10-28
Let you use the manage page to log in to weekly too

johnduffell:
So as per https://github.com/guardian/membership-common/pull/347

This PR uses the new approach of pattern matching to get hold of whether the subscription is weekly rather than delivery or voucher.  We get a load of warnings from the compiler due to the horrible way I've done the pattern matching so we need to improve it.

We can remove the warnings by doing an asInstanceOf instead of pattern matching an erased type parameter, but the only reason we don't get a warning is because we know asInstanceOf isn't safe, rather than it being a surprise.  I would prefer a way to lock the type parameters together so the compiler would at least check the code makes sense.

```
[warn] /Users/jduffell/ws/subscriptions-frontend/app/controllers/AccountManagement.scala:113: non-variable type argument com.gu.memsub.subsv2.PaidSubscriptionPlan[com.gu.memsub.Product.Delivery,com.gu.memsub.subsv2.PaperCharges] in type pattern com.gu.memsub.subsv2.Subscription[com.gu.memsub.subsv2.PaidSubscriptionPlan[com.gu.memsub.Product.Delivery,com.gu.memsub.subsv2.PaperCharges]] is unchecked since it is eliminated by erasure
[warn]         case SubMatch(Product.Delivery, charges: PaperCharges, sub: Subscription[PaidSubscriptionPlan[Product.Delivery, PaperCharges]]) =>
[warn]                                                                     ^
[warn] /Users/jduffell/ws/subscriptions-frontend/app/controllers/AccountManagement.scala:117: non-variable type argument com.gu.memsub.subsv2.PaidSubscriptionPlan[com.gu.memsub.Product.Voucher,com.gu.memsub.subsv2.PaperCharges] in type pattern com.gu.memsub.subsv2.Subscription[com.gu.memsub.subsv2.PaidSubscriptionPlan[com.gu.memsub.Product.Voucher,com.gu.memsub.subsv2.PaperCharges]] is unchecked since it is eliminated by erasure
[warn]         case SubMatch(Product.Voucher, _, sub: Subscription[PaidSubscriptionPlan[Product.Voucher, PaperCharges]]) =>
[warn]                                                ^
[warn] /Users/jduffell/ws/subscriptions-frontend/app/controllers/AccountManagement.scala:119: non-variable type argument com.gu.memsub.subsv2.PaidSubscriptionPlan[com.gu.memsub.Product.Weekly,com.gu.memsub.subsv2.PaidCharge[com.gu.memsub.Weekly.type,com.gu.memsub.BillingPeriod]] in type pattern com.gu.memsub.subsv2.Subscription[com.gu.memsub.subsv2.PaidSubscriptionPlan[com.gu.memsub.Product.Weekly,com.gu.memsub.subsv2.PaidCharge[com.gu.memsub.Weekly.type,com.gu.memsub.BillingPeriod]]] is unchecked since it is eliminated by erasure
[warn]         case SubMatch(_: Product.Weekly, _, sub: Subscription[PaidSubscriptionPlan[Product.Weekly, PaidCharge[Weekly.type, BillingPeriod]]]) =>
[warn]                                                  ^
```

I slightly prefer the asInstanceOf apart from the fact it's not together with the rest of the pattern match, but those warnings are not really acceptable.

Anyway we can discuss monday and come to a conclusion.  Then we can potentially delete the either method we already have.

After login holding page:
image

@pvighi @paulbrown1982 @jacobwinch

### DIFF
--- a/app/views/account/details.scala.html
+++ b/app/views/account/details.scala.html
@@ -4,28 +4,28 @@
     subscriberId: Option[String]
 )(implicit request: RequestHeader, flash: Flash, touchpointBackendResolution: services.TouchpointBackend.Resolution)
 
-@main("Cancel your papers when you're away | The Guardian", edition = UK, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
+@main("Manage your subscription | The Guardian", edition = UK, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
 
     <main class="page-container gs-container">
 
         <section class="suspend-container">
 
             <div class="suspend-header">
-                <h1 class="suspend-header__title">Cancel your papers while you're away</h1>
+                <h1 class="suspend-header__title">Manage your subscription</h1>
             </div>
 
             <div class="prose">
                 <p>
-                    You can use this service to temporarily suspend delivery of your newspapers.
-                    You will need your ​​Subscriber ID​, which you can find on any correspondence.
+                    To manage your subscription, you will need your ​​Subscriber ID​, which you can find on any correspondence.<br>
+                    You will also need your billing information, which we accept in either capitals or lower case.
                 </p>
                 <p>
-                    Please give us at least 5 days notice, and we will reduce the first payment after you go away by the value of the cancelled papers, including delivery.
-                </p>
-                <p>
-                    If you do not have your Subscriber ID to hand, or you have any other customer service enquir​ies,
-                    please contact us on: <a href="tel:+44 330 333 6767">+44 (0) 330 333 6767</a>
-                    or email <a href="mailto:homedelivery@@theguardian.com">homedelivery@@theguardian.com</a>.
+                    If you do not have your Subscriber ID to hand, or you have any other customer service enquir​ies, you can get in touch.
+                    <ul>
+                        <li>Home delivery, please contact us on: <a href="tel:+44 330 333 6767">+44 (0) 330 333 6767</a>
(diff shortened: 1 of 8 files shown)


paulbrown1982:
The changes to the details() page will need to go in another pull request as product and commercial teams will need to approve the wording, however for the rest, nice work - 👍

johnduffell:
ok regarding the warnings from downcasting, collectively we have attached this problem from every direction
This one is neat but doesn't have exhaustiveness checking so if we add a new product, it will happily work until runtime
image
And this one would produce a warning if it stops being exhaustive, but the `case <product>` has to be the same `sub.as<product>` otherwise it will just go into the getOrElse.
image
The latter I slightly prefer due to the exhaustiveness checking, but it is a little less neat.

app/views/account/details.scala.html
```diff
@@ -4,28 +4,28 @@
     subscriberId: Option[String]
 )(implicit request: RequestHeader, flash: Flash, touchpointBackendResolution: services.TouchpointBackend.Resolution)
 
-@main("Cancel your papers when you're away | The Guardian", edition = UK, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
+@main("Manage your subscription | The Guardian", edition = UK, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
```
paulbrown1982: Just a reminder to put this change into a separate PR, and so we can discuss what kind of multi-product copy we need, with Julia and maybe the Commercial team. This is a great starting point though.

PR #661, opened 2016-10-31
Fix holiday refund calculations

paulbrown1982:
Fixed 2 suspend and resume bugs and made 1 improvement:

1. The refund amount being wrong on a migrated sub's last term iff they have pro-rated charges. (note the After screenshot is still wrong on this issue as we have refunded the customer in a different way)
2. The number of suspendable days and the number of suspended days was wrong for multi-day subs
3. Now shows what days are being delivered on for a multi-day sub.

What will actually get billed:

picture 246

Before - the suspend and resume page was wrong:

picture 247

After - the suspend and resume page is right:

picture 245

cc @johnduffell @JuliaBellis @jacobwinch @pvighi @AWare @jayceb1


johnduffell:
👍 good fixes! 🚀

PR #666, opened 2016-11-02
Run acceptance test on voucher flow and add temporary hack for Stripe

jacobwinch:
This PR re-factors an acceptance-test so that it runs against voucher instead of home delivery, now that we handle all of this traffic ourselves.

It also introduces a temporary fix for the (currently broken) Stripe test, which now uses xpath to identify elements on Stripe Checkout form (as opposed to ids), because the ids on Stripe checkout are no longer set consistently. This is likely to be flakey - and should only be used temporarily - I will also contact Stripe and see if there is a better long-term solution for this.

### DIFF
--- a/test/acceptance/CheckoutSpec.scala
+++ b/test/acceptance/CheckoutSpec.scala
@@ -175,93 +175,12 @@ class CheckoutSpec extends FeatureSpec with Browser
       }
     }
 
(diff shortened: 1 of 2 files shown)


jacobwinch:
@johnduffell

johnduffell:
brilliant, that's great that you worked out the xpath! hope stripe will have a better idea though 👍 🚀

PR #670, opened 2016-11-03
Salesforce scheduled downtime

paulbrown1982:
Added a 'Down for maintenance' page. I've scheduled some downtime to match the Salesforce maintenance window this Sunday at 2am-5am.

picture 260

cc @johnduffell @jacobwinch @pvighi @AWare @JuliaBellis @jayceb1

PR #671, opened 2016-11-07
Revert "Salesforce scheduled downtime"

paulbrown1982:
Reverts guardian/subscriptions-frontend#670 

The code is not generic like the Membership version, and has served its purpose. 

cc @johnduffell


johnduffell:
👍 thanks! very neat revert too 😄

PR #676, opened 2016-11-07
bug with "pay in £GBP" checkbox 

pvighi:
fixed bug introduced in latest js refactor:
 If the "Pay in £GBP" checkbox was selected then the first time the country was changed the currency remained as £ and the checkbox would hide.

It should now reset the checkbox state when changing countries and the appropriate currency should be selected.

### DIFF
--- a/assets/javascripts/modules/checkout/fieldSwitcher.js
+++ b/assets/javascripts/modules/checkout/fieldSwitcher.js
@@ -106,8 +106,6 @@ define([
     };
 
     var redrawCurrencyOverride = function (currentState) {
-        $('.js-currency-override-checkbox').attr('checked', false);
-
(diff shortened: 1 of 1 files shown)


assets/javascripts/modules/checkout/fieldSwitcher.js
```diff
@@ -177,6 +175,10 @@ define([
         };
 
         var refresh = function () {
+            if (determinesPriceBanding) {
+                $('.js-currency-override-checkbox').attr('checked', false);
+            }
+            
             var currentState = getCurrentState();
```
pvighi: we need to reset the checkbox state before getting the current state otherwise it will always return the currency as GBP if it is checked

PR #679, opened 2016-11-08
Guardian Weekly - moving Canada to Zone B

paulbrown1982:
- Moved Canada to Zone B Rest of the world. 
- Also updated the UI links and some copy on the /weekly page.

E.g.

picture 280

cc @pvighi @johnduffell @JuliaBellis @jacobwinch @AWare


pvighi:
Looks good. But I checked out the branch and It's not switching to CAD when selecting Canada in Guardian Weekly zone B, this will probably work when we add a CAD price in Zuora for Zone B.
Other than that 👍

PR #686, opened 2016-11-16
when renewing Weekly, output salesforce email if available

johnduffell:
Further to https://github.com/guardian/membership-common/pull/354
This uses the above PR to get migrated weekly subscribers from salesforce, and only send their email address to the logged in page if they have one so that we can update it only if it's not present.

@pvighi @paulbrown1982 @AWare

### DIFF
--- a/app/services/ExactTargetService.scala
+++ b/app/services/ExactTargetService.scala
@@ -29,6 +29,7 @@ import views.support.Pricing._
 import scala.concurrent.Future
 import scala.reflect.internal.util.StringOps
 import scala.util.{Failure, Success, Try}
+import scalaz.syntax.std.option._
 
 /**
   * Sends welcome email message to Amazon SQS queue which is consumed by membership-workflow.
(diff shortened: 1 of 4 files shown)


app/services/ExactTargetService.scala
```diff
@@ -151,27 +152,32 @@ class ExactTargetService(
...
-          packageName = packageName,
-          billingSchedule = billingSchedule,
-          numberOfSuspensionsLinedUp,
-          daysUsed,
-          daysAllowed
-        )
+        salesforceContact.email.toRightDisjunction(s"no email in salesforce for ${subscription.id.get}").map { email =>
+          HolidaySuspensionBillingScheduleDataExtensionRow(
+            email = StringOps.oempty(email).headOption,
+            saltuation = constructSalutation(salesforceContact.title, salesforceContact.firstName, Some(salesforceContact.lastName)),
```
pvighi: not something introduced by this PR but this should probably be salutation instead of saltuation, right?

app/services/ExactTargetService.scala
```diff
@@ -151,27 +152,32 @@ class ExactTargetService(
...
+            subscriptionName = subscription.name.get,
+            subscriptionCurrency = subscription.plan.charges.price.currencies.head,
+            packageName = packageName,
+            billingSchedule = billingSchedule,
+            numberOfSuspensionsLinedUp,
+            daysUsed,
+            daysAllowed
```
pvighi: another thing not introduced in this PR but I would stick to by name params for all of them

johnduffell:
updated as per @pvighi suggestions and @AWare pointing out I was using the none-ness of email in two different ways!

pvighi:
👍

PR #687, opened 2016-11-16
countries and currency validation

pvighi:
On different parts of the renewal flow we're going to need to be aware of which currencies and country combinations are valid for each plan. This is currently hard coded in the checkout controller method that renders the subscription flow.

This PR :

- Removes the code to a common location in order to use it for the renewal flow

- Reuses it to provide server side currency and countries validation to the subscription flow

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -36,24 +36,8 @@ import scalaz.std.option._
 import scalaz.std.scalaFuture._
 import scalaz.syntax.applicative._
 import scalaz.{NonEmptyList, OptionT}
-
+import model.ContentSubscriptionPlanOps._
 object Checkout extends Controller with LazyLogging with CatalogProvider {
-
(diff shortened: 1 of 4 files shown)


app/controllers/Checkout.scala
```diff
@@ -90,65 +74,31 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
-            fallbackCountry = countryGroup.defaultCountry,
-            fallbackCurrency = GBP
-          )
-          case Product.WeeklyZoneB => getSettings(
-            availableDeliveryCountries = Some(CountryWithCurrency.whitelisted(supportedCurrencies, USD, weeklyZoneBGroups)),
-            availableBillingCountries = allCountriesWithCurrencyOrGBP,
-            fallbackCountry = None,
-            fallbackCurrency = USD
-          )
+          case Product.Digipack => getSettings(countryGroup.defaultCountry, GBP)
```
paulbrown1982: 👍 much clearer

PR #688, opened 2016-11-17
tell the user we'll take their card details later

johnduffell:
it was a little confusing that everything went blank where you expected to put the card deatils in.

Behold my client side dev skills....
image

@AWare @pvighi @paulbrown1982


paulbrown1982:
👍 👍 👍

PR #697, opened 2016-11-22
Handle removal of Ireland country group

paulbrown1982:
- Updated how the countryGroup is handled on the checkout page. It now accepts upper or lower case country ISO codes as well as country group IDs.
- Providing a country code will find the enclosing country-group AND set the default country within that group to the provided code. Falling back to UK as normal if an enclosing country group cannot be found.
- This change has been made because the Ireland country group has been removed from membership-common and the /checkout?countryGroup=ie is still a heavily linked-to URL (primarily at JellyFish).

cc @rtyley


paulbrown1982:
Requires https://github.com/guardian/membership-common/pull/358

rtyley:
:+1: this is feels better to me than Ireland becoming it's own country group :+1: :+1: :+1:

PR #698, opened 2016-11-23
Moved 'Converted' custom event/metric to being fired off the thank you page

paulbrown1982:
Moved 'Converted' custom event/metric to being fired off the thank you page.

cc @ajosephides @mkopka


mkopka:
looks good to me - thanks for helping

PR #704, opened 2016-11-25
pass renewal plans info to the client side

johnduffell:
In order to let the user choose what plans to renew onto we have to read the catalog and insert the pricing and product rate plan is into the page.

This relies on the changes in https://github.com/guardian/membership-common/pull/360

@paulbrown1982 @AWare @pvighi @jacobwinch

### DIFF
--- a/app/controllers/AccountManagement.scala
+++ b/app/controllers/AccountManagement.scala
@@ -5,11 +5,11 @@ import _root_.services.TouchpointBackend
 import _root_.services.TouchpointBackend.Resolution
 import actions.CommonActions._
 import com.github.nscala_time.time.OrderingImplicits.LocalDateOrdering
-import com.gu.memsub.Subscription.Name
+import com.gu.memsub.Subscription.{ProductRatePlanId, Name}
 import com.gu.memsub.services.GetSalesforceContactForSub
 import com.gu.memsub.subsv2.SubscriptionPlan.{Delivery, WeeklyPlan}
 import com.gu.memsub.subsv2._
-import com.gu.memsub.{BillingSchedule, Product}
+import com.gu.memsub.{BillingPeriod, Price, BillingSchedule, Product}
 import com.gu.subscriptions.suspendresume.SuspensionService
 import com.gu.subscriptions.suspendresume.SuspensionService.{BadZuoraJson, ErrNel, HolidayRefund, PaymentHoliday}
 import com.gu.zuora.soap.models.Queries.Contact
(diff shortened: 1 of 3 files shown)


app/controllers/AccountManagement.scala
```diff
@@ -134,6 +134,21 @@ object ManageWeekly extends LazyLogging {
...
+  def sequence[A](list: List[Either[String, A]]): Either[String, List[A]] = {
+    val errors = list collect {
+      case Left(x) => x
+    }
+    if (errors.nonEmpty)
+      Left(errors.mkString(", "))
+    else
+      Right(list collect {
+        case Right(x) => x
+      })
```
pvighi: What about doing something like this? 

list.partition(_.isLeft) match {
  case (Nil, successResponses) => res.map { case Right(success) => success }
  case (errors, _) => errors.mkString(", ")
}

app/views/account/renew.scala.html
```diff
@@ -5,13 +5,29 @@
...
+    <script>
+
+        guardian.plans = [
+            @plans.map { plan =>
+                {
+                    id: '@plan.id.get'
+                    price: '@plan.price.prettyAmount'
+                    currency: '@plan.price.currency'
+                    billingPeriod: '@plan.billingPeriod.noun'
+                },
```
johnduffell: @AWare is this format about right for client side?

AWare: Yeah, it should be alright. I'll figure out what I need when I've got the promo stuff done.

app/views/account/renew.scala.html
```diff
@@ -5,13 +5,29 @@
...
 )(implicit r: RequestHeader, touchpointBackendResolution: services.TouchpointBackend.Resolution)
 
 @main("Your Guardian Weekly subscription | The Guardian", bodyClasses = List("is-wide"), edition = UK, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
 
+    <script>
+
+        guardian.plans = [
+            @plans.map { plan =>
+                {
+                    id: '@plan.id.get'
```
paulbrown1982: Needs commas after every K:V.

app/views/account/renew.scala.html
```diff
@@ -5,13 +5,29 @@
...
 
+    <script>
+
+        guardian.plans = [
+            @plans.map { plan =>
+                {
+                    id: '@plan.id.get',
+                    price: '@plan.price.prettyAmount',
+                    currency: '@plan.price.currency',
+                    billingPeriod: '@plan.billingPeriod.noun',
```
paulbrown1982: The last one isn't supposed to have a comma...

PR #705, opened 2016-11-28
More data for tax

paulbrown1982:
* Added the card country and IP country to custom fields on a subscription.
* Wire up request data to Subscribe command. Just need to get the cardCountry from the form now.

cc @AWare @johnduffell @pvighi @ajosephides @jacobwinch

### DIFF
--- a/app/utils/RequestCountry.scala
+++ b/app/utils/RequestCountry.scala
@@ -1,13 +1,12 @@
 package utils
 
 import com.gu.i18n.CountryGroup
-import com.netaporter.uri.Uri
-import com.netaporter.uri.dsl._
 import play.api.mvc.Request
 
 object RequestCountry {
   implicit class RequestWithFastlyCountry(r: Request[_]) {
-    def getFastlyCountry = r.headers.get("X-GU-GeoIP-Country-Code").flatMap(CountryGroup.byFastlyCountryCode)
+    def getFastlyCountryGroup = r.headers.get("X-GU-GeoIP-Country-Code").flatMap(CountryGroup.byFastlyCountryCode)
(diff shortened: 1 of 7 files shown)


app/utils/RequestCountry.scala
```diff
@@ -1,13 +1,12 @@
...
 import com.gu.i18n.CountryGroup
-import com.netaporter.uri.Uri
-import com.netaporter.uri.dsl._
 import play.api.mvc.Request
 
 object RequestCountry {
   implicit class RequestWithFastlyCountry(r: Request[_]) {
-    def getFastlyCountry = r.headers.get("X-GU-GeoIP-Country-Code").flatMap(CountryGroup.byFastlyCountryCode)
+    def getFastlyCountryGroup = r.headers.get("X-GU-GeoIP-Country-Code").flatMap(CountryGroup.byFastlyCountryCode)
+    def getFastlyCountry = r.headers.get("X-GU-GeoIP-Country-Code").flatMap(CountryGroup.countryByCode)
```
johnduffell: why is the above by"fastly"countrycode, but this one is just by code? are they both fastly codes really?

johnduffell:
apart from the comment looks good! 👍 🚤

app/utils/RequestCountry.scala
paulbrown1982: Agreed, there is a bad name in membership-common. The cost/benefit ratio of fixing that is quite large as it'll require releasing all 3 apps which are dependent on membership-common. So I've raised a ticket in Trello: https://trello.com/c/MZwyk9HK/89-fix-countrygroup-byfastlycountrycode-api-nomenclature-in-membership-common

PR #682, opened 2016-11-09
Remove Outer set of postcodes

paulbrown1982:
Apparently we should not be offering delivery to the Outer set of postcodes.

cc @johnduffell @jacobwinch @pvighi @JuliaBellis @AWare

PR #712, opened 2016-12-02
Use an official Home Delivery postcode whitelist

paulbrown1982:
Use an official postcode whitelist, the old lists were out of date and included the Outer set of which only some are delivered to. This change replaces https://github.com/guardian/subscriptions-frontend/pull/682 which has been cancelled.

Moved the validation to the client side now that the number of postcode prefixes is much shorter.

cc @johnduffell @jacobwinch @pvighi @JuliaBellis @AWare


johnduffell:
looks good, although it does seem crazy we do this on the server side 😕 
👍

paulbrown1982:
cc @johnduffell

johnduffell:
yep looks good to me! 👍

PR #713, opened 2016-12-02
don't need to input username into identity in the tests now

johnduffell:
same fix as https://github.com/guardian/identity-frontend/pull/187 

Basically identity doesn't have a username field that needs inputting any more, so we shouldn't try to use it in the tests.

@guardian/membership-and-subscriptions


paulbrown1982:
👍

johnduffell:
not sure why this is failing, I reproduced it locally, but if I do it manually in a browser it works ok

PR #718, opened 2016-12-05
Promo renew 

AWare:
Some promo renew stuff. 
I'm still tidying it up so it looks nice, but here's a WIP for review.
image
image
image

### DIFF
--- a/assets/javascripts/modules/react/weeklyRenew.jsx
+++ b/assets/javascripts/modules/react/weeklyRenew.jsx
@@ -91,25 +91,32 @@ class WeeklyRenew extends React.Component {
             promosStatus: status.LOADING
         });
 
-        validatePromo(this.state.promoCode, this.props.country).then((a)=>{
-            let update = validatePromoForPlans(a,this.state.plans);
-            if(update){
+        validatePromo(this.state.promoCode, this.props.country).then((a) => {
+            let newPlans = validatePromoForPlans(a, this.state.plans);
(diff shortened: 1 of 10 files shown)


assets/javascripts/modules/react/weeklyRenew.jsx
```diff
@@ -82,7 +139,7 @@ class WeeklyRenew extends React.Component {
     }
 
     handleAccountHolder(e) {
-        let name = e.target.value.substring(0,18);
+        let name = e.target.value.substring(0, 18);
```
paulbrown1982: We should make this magic number a const.

assets/javascripts/modules/react/weeklyRenew.jsx
```diff
@@ -92,18 +149,26 @@ class WeeklyRenew extends React.Component {
...
+            this.setState({plan: plan.id});
         }
     }
 
-    handleError(error){
-        this.setState({error:error,loading:false});
+    getPlan() {
+        return this.state.plans.filter((plan) => {
+            return plan.id == this.state.plan
+        })[0]
```
paulbrown1982: What if the filter returns an empty array?

assets/javascripts/modules/react/weeklyRenew.jsx
AWare: I'll put it on the element as a maxlength.

PR #719, opened 2016-12-05
standalone registration needs display name

johnduffell:
This should fix the failing prout tests.

Added the code to enter the display name to the tests.  Turned out if there's a validation failure in identity, nothing checks that it actually registered, it just assumes it worked.  So I've added a redirect to subs frontend and a check it actually ended up there, so we know what went wrong next time it happens.

@jacobwinch @AWare

### DIFF
--- a/test/acceptance/pages/Register.scala
+++ b/test/acceptance/pages/Register.scala
@@ -3,9 +3,13 @@ package acceptance.pages
 import acceptance.util.{TestUser, LoadablePage, Browser, Config}
 
 class Register(val testUser: TestUser) extends LoadablePage with Browser {
-  val url = s"${Config.identityFrontendUrl}/register?skipConfirmation=true"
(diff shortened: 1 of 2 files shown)


test/acceptance/pages/Register.scala
```diff
@@ -3,10 +3,14 @@ package acceptance.pages
 import acceptance.util.{TestUser, LoadablePage, Browser, Config}
 
 class Register(val testUser: TestUser) extends LoadablePage with Browser {
-  val url = s"${Config.identityFrontendUrl}/register?skipConfirmation=true"
+  val url = s"${Config.identityFrontendUrl}/register?skipConfirmation=true&returnUrl=${Config.baseUrl}/uk"
 
   def hasLoaded(): Boolean = pageHasElement(createAccountButton)
 
+  def hasCreated(): Boolean = pageHasElement(subscriptionsUKDigital)
```
jacobwinch: As you say, I guess we might want a more robust way of checking that the Identity account has been created successfully, but this is an improvement on our current status!

PR #724, opened 2016-12-08
Adding Guardian Weekly Zone C rate plans

paulbrown1982:
- Adding Guardian Weekly Zone C rate plans, and updating those as what we are going to sell.
- Guardian Weekly Zone B will be what the migrated customers are set as, and the renewal options will be also be within that set.

See: https://github.com/guardian/membership-common/pull/374

cc @johnduffell @pvighi @AWare @jacobwinch

### DIFF
--- a/app/views/support/PlanOps.scala
+++ b/app/views/support/PlanOps.scala
@@ -50,7 +50,10 @@ object PlanOps {
 
     def isDigitalPack: Boolean = in.product == com.gu.memsub.Product.Digipack
 
-    def isGuardianWeekly: Boolean = in.product.isInstanceOf[Product.Weekly]
+    def isGuardianWeekly: Boolean = in.product match {
+      case _:Product.Weekly => true
+      case _ => false
+    }
 
     def hasPhysicalBenefits: Boolean = in.charges.benefits.list.exists(_.isPhysical)
 
--- a/app/views/support/PlanOps.scala
+++ b/app/views/support/PlanOps.scala
@@ -50,7 +50,7 @@ object PlanOps {
 
     def isDigitalPack: Boolean = in.product == com.gu.memsub.Product.Digipack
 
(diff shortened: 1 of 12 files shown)


app/views/support/PlanOps.scala
```diff
@@ -50,7 +50,7 @@ object PlanOps {
 
     def isDigitalPack: Boolean = in.product == com.gu.memsub.Product.Digipack
 
-    def isGuardianWeekly: Boolean = in.product == Product.WeeklyZoneA || in.product == Product.WeeklyZoneB // TODO is this right to include both?
+    def isGuardianWeekly: Boolean = in.product.isInstanceOf[Product.Weekly]
```
johnduffell: type safety would be lost I think? `==` will give a warning if the types can't ever match but would `isInstanceOf` do that?  another option is something like
```scala
in.product match {
  case Product.Weekly => true
  case _ => false
}```

File: app/views/testing/testUsers.scala.html [hunk 2]
```diff
@@ -22,6 +22,9 @@
         <div class="actions">
             <h2>Subscribe as a <em>guest</em> user to:</h2>
             @for(group <- products) {
+                @group.headOption.map { plan =>
+                    <h4>@plan.product.name.capitalize:</h4>
```
johnduffell: good idea!

johnduffell:
yep, looks good, just the whole zoneb and zonec being the same thing that's confusing, if we can make a ZoneBLoyalty and a ZoneB it would be slightly nicer
👍 🚅

paulbrown1982:
I've raised https://trello.com/c/9bZ92SUh/100-rename-the-product-zones-to-something-more-appropriate to cover this after the migration. Right now we need to move on and get the renewal and acquisition promotions added to the tool.

PR #725, opened 2016-12-13
Add a week to the voucher book delivery day

paulbrown1982:
Add a week to the voucher book delivery day. This will be for a few weeks while we wait for there to be dual reports created in Salesforce for renewing subs vs new subs. Right now there's only one report generating the fulfilment files and it must have only one date set - and those customers want their books 3 weeks in advance not 2.

Fixed a few var-should-be-const errors which IntelliJ flagged up.

cc @AWare @johnduffell @pvighi @jacobwinch @jayceb1


johnduffell:
👍 fair enough

PR #726, opened 2016-12-13
handling of non renewable weekly subs in /manage

pvighi:
* only show data for the newest plan attached to the sub in /manage
* only show renewal option if the newest plan is a one off plan
* just in case validate on the renewal logic that the sub is renewable before doing any processing


paulbrown1982:
👍

PR #727, opened 2016-12-15
Handle responses from the weekly renewal endpoint

pvighi:
* added a basic thank you page if the weekly renewal is successful
* display a message if the renewal cannot be completed (maybe we could add something to the error displaying logic so that it add information about who to call or email if things go wrong)

Like with the rest of the renewal flow, this implements the correct behaviour but some additional work needs to be done to make sure it looks as expected as well. 
@paulbrown1982 @johnduffell


pvighi:
It's pretty basic:
thank-you-renew

app/controllers/AccountManagement.scala
```diff
@@ -286,22 +287,26 @@ object AccountManagement extends Controller with LazyLogging {
...
     SessionSubscription.subscriptionFromRequest flatMap { maybeSub =>
       val response = for {
         sub <- maybeSub.toRightDisjunction("no subscription in request")
         renewableSub <- validateRenewable(sub)
         renew <- parseRenewalRequest(request, tpBackend.catalogService.unsafeCatalog)
       } yield {
-        tpBackend.checkoutService.renewSubscription(renewableSub, renew).map(res => Ok(res.id))
+        tpBackend.checkoutService.renewSubscription(renewableSub, renew).map(res => Ok(Json.obj("redirect" -> routes.AccountManagement.renewThankYou.url)))
+        .recover{
+          case e: Throwable => InternalServerError(jsonError("Unexpected error while renewing subscription."))
```
johnduffell: some logging would be needed here I think

app/controllers/AccountManagement.scala
```diff
@@ -314,5 +319,18 @@ object AccountManagement extends Controller with LazyLogging {
...
+  def renewThankYou() = accountManagementAction.async { implicit request =>
+    implicit val resolution: TouchpointBackend.Resolution = TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)
+    implicit val tpBackend = resolution.backend
+    import model.SubscriptionOps._
+
+    val res = for {
+    subscription <- OptionT( SessionSubscription.subscriptionFromRequest)
+    billingSchedule <- OptionT(tpBackend.commonPaymentService.billingSchedule(subscription.id, numberOfBills = 13))
+    }yield {
+      Ok(thankyou_renew(subscription.latestPlan,billingSchedule, resolution))
```
johnduffell: not really conventional to use _ in views, would normally be package.name.lowerCamelCase

app/views/account/fragments/yourDetails.scala.html
```diff
@@ -6,16 +6,15 @@
 @import scala.collection.immutable.List.empty
 @import com.gu.salesforce.Contact
 @import model.SubscriptionOps._
+@import org.joda.time.LocalDate
 
-@(subscription: Subscription[SubscriptionPlan.PaperPlan], chosenPaperDays: List[PaperDay] = empty, maybeContact: Option[Contact] = None)(extra: Html = Html(""))
+@(subscriptionName: String, plan: SubscriptionPlan.PaperPlan, maybeStartDate: Option[LocalDate], chosenPaperDays: List[PaperDay] = empty, maybeContact: Option[Contact] = None)(extra: Html = Html(""))
```
johnduffell: good - less painful way to get around the Subscription of some type erased plan problem

johnduffell:
Sounds good to me regarding the wording!

There are some weird spacing going on in a few places in the code, but functionality wise I think it's really solid 👍 🎉

app/views/account/fragments/yourDetails.scala.html
paulbrown1982: https://en.wikipedia.org/wiki/Rule_of_least_power FTW!

app/controllers/AccountManagement.scala
paulbrown1982: Seconded.

conf/routes
```diff
@@ -54,6 +54,8 @@ GET         /manage/suspend                     controllers.AccountManagement.re
 POST        /manage/suspend                     controllers.AccountManagement.processSuspension
 GET         /manage/signout                     controllers.AccountManagement.logout
 POST        /manage/renew                       controllers.AccountManagement.processRenewal
+GET         /manage/renew/thank-you             controllers.AccountManagement.renewThankYou()
```
paulbrown1982: Why is there () at the end of this controller def?

paulbrown1982:
👍

pvighi:
In case of any server side error while renewing the sub it will display this error message:
error

paulbrown1982:
👍 looks great!

PR #729, opened 2016-12-15
Fix bug with bill to country display

jacobwinch:
Noticed a small copy bug on the /manage page.

Before:
oldcountry

After:
newcountry


jacobwinch:
Closing this because I'm making some more significant changes which make this redundant.

PR #731, opened 2016-12-16
Move renew/view details logic into controller

jacobwinch:
This PR moves the logic which decides whether or not we show the Renew section out of the view and into the controller.

It also tidies up the views a bit (although I'm sure there are still improvements which can be made to this!):

Details page (i.e. auto-renewing subs):

weeklydetails

Renew page:

weeklyrenew

@paulbrown1982 @pvighi @johnduffell @AWare

### DIFF
--- a/app/views/account/weeklyDetails.scala.html
+++ b/app/views/account/weeklyDetails.scala.html
@@ -15,19 +15,6 @@
     @helper.javascriptRouter("jsRoutes")(
         routes.javascript.Promotion.validate
     )
-    <script>
-
-        guardian.plans = [
-            @plans.map { plan =>
-            {
-                id: '@plan.id.get',
-                price: '@plan.price'
-            },
-            }
-        ]
-
-    </script>
-
     <main class="page-container gs-container">
         <section class="suspend-container">
 
--- a/app/views/account/weeklyDetails.scala.html
+++ b/app/views/account/weeklyDetails.scala.html
@@ -1,0 +1,60 @@
+@import com.gu.memsub.BillingSchedule
+@import com.gu.memsub.subsv2.SubscriptionPlan
+@import com.gu.memsub.subsv2.Subscription
+@import model.DigitalEdition.UK
+@import com.gu.salesforce.Contact
+@import com.gu.i18n.Country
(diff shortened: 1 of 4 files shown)


app/views/account/weeklyDetails.scala.html
```diff
@@ -0,0 +1,60 @@
...
+    subscription: Subscription[SubscriptionPlan.WeeklyPlan], billingSchedule: BillingSchedule, contact: Contact, billToCountry: Country, plans: List[WeeklyPlanInfo]
+)(implicit r: RequestHeader, touchpointBackendResolution: services.TouchpointBackend.Resolution)
+
+@main("Your Guardian Weekly subscription | The Guardian", bodyClasses = List("is-wide"), edition = UK, touchpointBackendResolutionOpt = Some(touchpointBackendResolution)) {
+    @helper.javascriptRouter("jsRoutes")(
+        routes.javascript.Promotion.validate
+    )
+    <script>
+
+        guardian.plans = [
```
johnduffell: this is just passing to the react form so it knows what to renew people onto so we problably don't need it when we're not allowing renewals

app/views/account/weeklyRenew.scala.html
```diff
@@ -48,47 +48,20 @@ <h3 class="mma-section__header">
...
+                <h3 class="mma-section__header">Renew your subscription</h3>
 
-                @if(subscription.renewable){
+                <p>
+                @if(subscription.termEndDate.isAfter(LocalDate.now)) {
+                    Your current subscription is due to expire on @prettyDate(subscription.termEndDate).
+                } else {
+                    Your subscription expired on @prettyDate(subscription.termEndDate).
+                }
+                Please complete the form below to renew your subscription. You can find your Promo Code on any correspondence from us.
```
johnduffell: do we always send a promo code? not sure if this would be on all the letters anyway?  maybe no need to mention promo codes separately

paulbrown1982: Let's just leave out that last sentence.

johnduffell:
good work splitting things out, very sensible 👍 🚀

paulbrown1982:
👍 Looks great!

PR #733, opened 2016-12-19
update membership common to find only current subs

johnduffell:
also ship https://github.com/guardian/membership-common/pull/383 here just to be on the safe side
@paulbrown1982 @pvighi @AWare


paulbrown1982:
👍

PR #738, opened 2016-12-20
Promo code field

johnduffell:
mainly for @AWare in case he needs to update the react side for renewals, it should info the details into the log when one is submitted
cc @paulbrown1982 @pvighi


paulbrown1982:
👍

johnduffell:
@AWare added the field, if it doesn't make sense from your end we can change the reader, but at least it gives us something to get going with

PR #741, opened 2016-12-22
pass in a list of rate plan for renewal instead of one

johnduffell:
This just uses the non empty list renewal instead of the single item renewal call.

@pvighi @paulbrown1982

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.330-SNAPSHOT",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.329",
+    "com.gu" %% "membership-common" % "0.330-SNAPSHOT",
```
johnduffell: https://github.com/guardian/membership-common/pull/387

pvighi:
retroactive 👍

PR #742, opened 2016-12-22
fixed promo code serialisation for renewals

pvighi:
The current deserialisation code expects to deserialise promo codes in the format : {"get":"promoCodeValue"}
instead the frontend is posting them as plain strings in the Renewal JSON, so this change alters the reader to construct a PromoCode case class from just a String.

@johnduffell @AWare @paulbrown1982

### DIFF
--- a/app/model/Renewal.scala
+++ b/app/model/Renewal.scala
@@ -37,7 +37,14 @@ class RenewalReads(catalog: Catalog) {
         case None => JsError("invalid payment data type")
       }
   }
-  implicit val promoReads = Json.reads[PromoCode]
+
+  implicit val promoReads = new Reads[PromoCode] {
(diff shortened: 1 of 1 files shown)


johnduffell:
good spot, I did wonder about that 😊 
👍

AWare:
If PromoCode in membership-common extended AnyVal, this wouldn't be needed.

paulbrown1982:
@AWare yup, we can do this in the New Year. It would require Membership-frontend regression testing.

👍  @pvighi

paulbrown1982:
Sorry, actually @pvighi if the string is empty, the reader should return a None.

pvighi:
@paulbrown1982 :  alex will change the client side code so that it leaves out the promo code parameter instead of sending it as an empty string when there is no promo code

PR #745, opened 2016-12-22
Fixed bug in suspend page based which wrongly ignored zero sum invoices

paulbrown1982:
I think this was when we did the billing schedule another way.

cc @johnduffell


johnduffell:
good work! 👍 😨

PR #746, opened 2017-01-03
Serialise the title to salesforce

AWare:
Add title into Salesforce account.

cc @ajosephides @jayceb1


paulbrown1982:
Here is the same change in Membership-workflow: https://github.com/guardian/membership-workflow/commit/946e033c7c8e722f870587f5a66eb7d7b17cf883. Note it's deliberately not added to Membership-frontend as it does not collect a title as part of the sign up flow.

PR #747, opened 2017-01-04
Remove billing period from gw checkout title

paulbrown1982:
Tweaking the checkout page title for Guardian Weekly to reflect the type of product being bought not the exact package name. 

- Parts of the page still use the original rate plan names so I've created a packageName method on the view helper class and use that in all existing places except the page title. This makes things much more consistent across HTML templates and Scala now.
- I also stopped the title changing as the radios change (more mobile friendly). 
- Added appropriate alt text to the packshot images.

picture 1

cc @johnduffell @AWare


johnduffell:
👍 great! 🚤

PR #751, opened 2017-01-05
Promotion nomenclature fixes and field ordering

paulbrown1982:
- Refactored the 'promo' field names nomenclature into their respective 'promotion', 'promoCode' or 'promotionalPrice'. 'promo' should not be used on its own.
- Re-ordered the form so that the Promo code field comes first above the list of plans.
- Changed copy of terms and conditions link and make it open in a new page.

cc @AWare @johnduffell

### DIFF
--- a/app/views/account/weeklyRenew.scala.html
+++ b/app/views/account/weeklyRenew.scala.html
@@ -61,7 +61,7 @@
                 </p>
 
                 <div class="js-weekly-renew" data-email="@contact.email"
-                data-country="@billToCountry.alpha2" data-promotion="test" />
+                data-country="@billToCountry.alpha2" data-promo-code="" />
             </section>
             <section class="mma-section">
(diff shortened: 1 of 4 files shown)


app/views/account/weeklyRenew.scala.html
```diff
@@ -61,7 +61,7 @@ <h3 class="mma-section__header">Renew your subscription</h3>
                 </p>
 
                 <div class="js-weekly-renew" data-email="@contact.email"
-                data-country="@billToCountry.alpha2" data-promotion="test" />
+                data-country="@billToCountry.alpha2" data-promo-code="" />
```
johnduffell: ah I did wonder why that was happening!

johnduffell:
looks good, although it's still inconsistent we use "promo" and "promotion" interchangably
👍

paulbrown1982:
@johnduffell agreed we should never use 'promo' alone for anything. I didn't change the 'Promo*' directive or function names in this change, only variables and attributes.

PR #754, opened 2017-01-05
put the delivery country code in the element

johnduffell:
The client side needs the delivery country code for validating the promo codes.  This puts it in the div so we can use it.

I also moved the currency to the div as well as it's better not to proliferate those global variables, thanks for the suggestion Paul.

@AWare @paulbrown1982

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.334",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.334",
+    "com.gu" %% "membership-common" % "0.336-SNAPSHOT",
```
johnduffell: https://github.com/guardian/membership-common/pull/395

PR #755, opened 2017-01-05
Format promotional prices in the right currency

paulbrown1982:
- Added a requested currency hint to both promotion validation AJAX URLs in the checkout and renewal page. Builds upon https://github.com/guardian/subscriptions-frontend/pull/754 for the renewal page.
- Fixed a bug where the promo code was not passed to the checkout if there was no preselected country, because it wouldn't validate. Now we fall back to using the provided one iff the code exists in the DB.
- Altered how the AJAX URL for the promoCode form in the checkout is generated , using the Play javascriptRouter like renewal does (no longer via data-lookup-url=""). This is good because the promoCode.js module doesn't need to know the request parameter names.
- Made the promo code lookup on the checkout read the country from the delivery country, falling back to the billing country.
- Some refactoring of the validatePromo* JavaScript function names in the renewal section, moving them from renew.es6 to promoCode.es6.

cc @johnduffell @AWare

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -120,10 +120,7 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
         val trackingCodeSessionData = validatedPromoCode.right.toSeq.flatten
         val supplierCodeSessionData = resolvedSupplierCode.map(code => Seq(SupplierTrackingCode -> code.get)).getOrElse(Seq.empty)
         val productData = ProductPopulationData(user.map(_.address), planList)
-        val promoCodeExists = for {
-          code <- promoCode
-          promotion <- tpBackend.promoService.findPromotion(code)
-        } yield code
+        val promoCodeExists = promoCode.filter(tpBackend.promoService.findPromotion(_).isDefined)
 
         Ok(views.html.checkout.payment(
           personalData = personalData,
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -110,22 +110,27 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
         val digitalEdition = model.DigitalEdition.getForCountry(countryAndCurrencySettings.defaultCountry)
 
         // either a code to send to the form (left) or a tracking code for the session (right)
-        val promo: Either[PromoCode, Seq[(String, String)]] = (promoCode |@| countryAndCurrencySettings.defaultCountry) (
+        val countryToValidatePromotionAgainst = countryAndCurrencySettings.defaultCountry orElse determinedCountryGroup.defaultCountry
+        val validatedPromoCode: Either[PromoCode, Seq[(String, String)]] = (promoCode |@| countryToValidatePromotionAgainst) (
           tpBackend.promoService.validate[NewUsers](_: PromoCode, _: Country, planList.default.id)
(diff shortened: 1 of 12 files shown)


app/controllers/Checkout.scala
```diff
@@ -110,22 +110,27 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
         val digitalEdition = model.DigitalEdition.getForCountry(countryAndCurrencySettings.defaultCountry)
 
         // either a code to send to the form (left) or a tracking code for the session (right)
```
johnduffell: this comment is in slightly the wrong place now

app/controllers/Checkout.scala
```diff
@@ -110,22 +110,27 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
-        val promo: Either[PromoCode, Seq[(String, String)]] = (promoCode |@| countryAndCurrencySettings.defaultCountry) (
+        val countryToValidatePromotionAgainst = countryAndCurrencySettings.defaultCountry orElse determinedCountryGroup.defaultCountry
+        val validatedPromoCode: Either[PromoCode, Seq[(String, String)]] = (promoCode |@| countryToValidatePromotionAgainst) (
```
johnduffell: I know it's not your change, but a `Seq[(String, String)]` is a tracking code?

app/controllers/Checkout.scala
```diff
@@ -110,22 +110,27 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
 
         val resolvedSupplierCode = supplierCode orElse request.session.get(SupplierTrackingCode).map(SupplierCode) // query param wins
-        val trackingCodeSessionData = promo.right.toSeq.flatten
+        val trackingCodeSessionData = validatedPromoCode.right.toSeq.flatten
         val supplierCodeSessionData = resolvedSupplierCode.map(code => Seq(SupplierTrackingCode -> code.get)).getOrElse(Seq.empty)
         val productData = ProductPopulationData(user.map(_.address), planList)
+        val promoCodeExists = for {
+          code <- promoCode
+          promotion <- tpBackend.promoService.findPromotion(code)
+        } yield code
```
johnduffell: this whole for comprehension is just doing this I think?:
```scala
promoCode.filter { tpBackend.promoService.findPromotion(_).isDefined }
```

app/controllers/Promotion.scala
```diff
@@ -20,20 +20,21 @@ import views.support.{BillingPeriod => _}
 
 object Promotion extends Controller with LazyLogging with CatalogProvider {
 
-  def getAdjustedRatePlans(promo: AnyPromotion, country:Country)(implicit tpBackend:TouchpointBackend): Option[Map[String, String]] = {
+  def getAdjustedRatePlans(promo: AnyPromotion, country:Country, requestedCurrency: Option[Currency])(implicit tpBackend:TouchpointBackend): Option[Map[String, String]] = {
+    def impliedCurrency = CountryGroup.byCountryCode(country.alpha2).getOrElse(CountryGroup.UK).currency
```
johnduffell: in some ways I'd prefer to see the getOrElses all in the same place to make the order of precendence clearer

johnduffell:
looks good, some really good use of variable names, not so sure about the client side stuff, and there's a lot in one PR so it's less clear what relates to what but overall it's good! 👍 🚀

app/controllers/Promotion.scala
paulbrown1982: Logged in Trello: https://trello.com/c/coarykrm/108-tidy-up-code-in-checkout-scala-for-trackingcode

PR #756, opened 2017-01-06
Use the latest membership-common for extra field in NewpaperLandingPage 

paulbrown1982:
Use the latest membership-common which has an extra Markdown field in the NewpaperLandingPage model that can be used to override the promotion description on the Newspaper promo landing page. See https://github.com/guardian/memsub-promotions/pull/90 for where the field will be edited.

picture 6

cc @johnduffell @AWare

### DIFF
--- a/app/views/promotion/newspaperLandingPage.scala.html
+++ b/app/views/promotion/newspaperLandingPage.scala.html
@@ -48,7 +48,12 @@
         @fragments.page.header(title)
 
         <section class="promotion-description">
-            <p>@promotion.landingPage.description.filter(_.isEmpty).getOrElse(promotion.description)</p>
+            @promotion.landingPage.description.fold {
+                <p>@promotion.description</p>
+            } { d =>
+                <div>@Html(md.render(d))</div>
+                <br/>
+            }
 
             @promotion.asIncentive.map { p =>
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -42,7 +42,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.334",
+    "com.gu" %% "membership-common" % "0.336-SNAPSHOT",
```
johnduffell: https://github.com/guardian/membership-common/pull/397

app/views/promotion/newspaperLandingPage.scala.html
```diff
@@ -48,7 +48,7 @@
         @fragments.page.header(title)
 
         <section class="promotion-description">
-            <p>@promotion.description</p>
+            <p>@promotion.landingPage.description.filter(_.isEmpty).getOrElse(promotion.description)</p>
```
johnduffell: it's an option of string, but that string might itself be empty - can we resolve that when it goes into the promotion object rather than flattening it out at the point of use?

johnduffell:
👍 looks good!

PR #743, opened 2016-12-22
Feed promotions to renew [wip]

AWare:
Also contains a not great attempt at passing the country through, but I really don't like it as it's no longer using a gu.Country and some of the direct-debit logic is less clear than before so it needs tidying after christmas.

PR #758, opened 2017-01-06
Feed promotions to renew

paulbrown1982:
Second go of https://github.com/guardian/subscriptions-frontend/pull/743 but with the healthcheck working properly!

The issue was a race condition with the healthcheck, such that it was creating a new future each time it mapped over the future that was the task we were really waiting for. The bootstrap time of those new futures meant they were not necessarily completed in time for the .isCompleted evaluation to return true.

cc @AWare


paulbrown1982:
👍

PR #760, opened 2017-01-06
Remove the savings % as these will change v soon, and also remove Le …

paulbrown1982:
Remove the Guardian Weekly homepage savings % as these will change v soon, and also removed Le Monde as they are no longer a publishing partner for Guardian Weekly.

Before:
picture 10
picture 13

After:

picture 14
picture 11

cc @johnduffell @AWare


johnduffell:
good just not sure about the bold! 👍 🚀

PR #763, opened 2017-01-06
upgrade of library broke renew

AWare:
A change to react-switch-button broke the renew js. 
Probably worth switching to yarn.


paulbrown1982:
We should create a ticket in Trello with options to stop this issue from potentially happening for other npm imports too.

PR #765, opened 2017-01-10
Fix renewal date starts

paulbrown1982:
A few changes to ensure the renewal dates and display of the renew funnel make sense.
Requires: https://github.com/guardian/membership-common/pull/398

- We always set the effective and acceptance date of the renewal amendment to be the end of the one-off term
- We now show when the sub will switch from the one-off to the renewed one if the renewal date is in the future (latestPlan stuff)
- We always add the new product with a future date if the renewal is in the future, or today's date. Never a past date.
- Updated copy on the thank you page and the weeklyDetails page.

picture 17
picture 15

cc @jacobwinch @johnduffell @AWare

PR #766, opened 2017-01-10
Check that we have salesforce auth in the healthcheck

AWare:
https://github.com/guardian/membership-common/pull/399
Implement healtcheck for ^.


AWare:
This will fail to compile and deploy because of membership common. https://github.com/guardian/subscriptions-frontend/pull/765 depends on this

PR #767, opened 2017-01-11
Check that we have salesforce auth in the healthcheck

AWare:
Move isAuthenticated into common

Update mem-common

https://github.com/guardian/subscriptions-frontend/pull/766


paulbrown1982:
👍

PR #770, opened 2017-01-18
Enable quarterly and annual subscribers to set holiday credits

paulbrown1982:
Fixed the restriction where holiday credit amounts could only be calculated for monthly subscribers. This previously meant quarterly and annual subscribers were advised to call the call centre.

There was a bug in subscriptions frontend where we weren't properly identifying the billing period for Echo Legacy subs which didn't have a paper on one or more days of the week, and instead showed the suspend form rather than the advice message. This meant some quarterly and annual customers could create a holiday suspension and the refund would be for a massive amount! The bug was in the avgPeriodBetweenInvoices method in suspend.scala.html. It should have filtered out invoices with a gross of zero. However rather than doing that fix, this change takes in the latest membership common which now actually calculates the refund correctly, so that restriction/call centre message can now be removed.

Requires: https://github.com/guardian/membership-common/pull/407

cc @johnduffell @jacobwinch @jayceb1 @AWare

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -77,7 +77,7 @@ resolvers ++= Seq(
     "scalaz-bintray" at "http://dl.bintray.com/scalaz/releases",
     Resolver.sonatypeRepo("releases"))
 
-addCommandAlias("devrun", "run -Dconfig.resource=DEV.conf 9200")
+addCommandAlias("devrun", "run -Dconfig.resource=PROD.conf 9200")
 addCommandAlias("fast-test", "testOnly -- -l Acceptance")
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -43,7 +43,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.345",
+    "com.gu" %% "membership-common" % "0.346-SNAPSHOT",
```
AWare: updates to 347 in https://github.com/guardian/subscriptions-frontend/pull/773

PR #774, opened 2017-01-24
Display any existing customer account credit balance in the first positive invoice

paulbrown1982:
Tweaks to use the new billingSchedule API which now takes an account as it applies any existing credit balance to the first positive invoice.

Requires: https://github.com/guardian/membership-common/pull/410

cc @johnduffell @pvighi @jacobwinch @AWare @jayceb1 

picture 2

### DIFF
--- a/app/controllers/AccountManagement.scala
+++ b/app/controllers/AccountManagement.scala
@@ -78,12 +78,12 @@ object ManageDelivery extends ContextLogging {
       form <- EitherT(Future.successful(SuspendForm.mappings.bindFromRequest().value \/> "Please check your selections and try again"))
       maybeDeliverySub <- EitherT(SessionSubscription.subscriptionFromRequest.map(_ \/> "Could not find an active subscription"))
       sub <- EitherT(Future.successful(maybeDeliverySub.asDelivery \/> "Is not a Home Delivery subscription"))
-      billCycleDay <- EitherT(tpBackend.zuoraService.getAccount(sub.accountId).map(_.billCycleDay).map(intToEither))
+      account <- EitherT(tpBackend.zuoraService.getAccount(sub.accountId).map(toScalaZRight))
       newHoliday = PaymentHoliday(sub.name, form.startDate, form.endDate)
       // 14 because + 2 extra months needed in calculation to cover setting a 6-week suspension on the day before your 12th billing day!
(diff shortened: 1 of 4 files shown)


pvighi:
LGTM 👍

app/controllers/AccountManagement.scala
```diff
@@ -78,12 +78,12 @@ object ManageDelivery extends ContextLogging {
       form <- EitherT(Future.successful(SuspendForm.mappings.bindFromRequest().value \/> "Please check your selections and try again"))
       maybeDeliverySub <- EitherT(SessionSubscription.subscriptionFromRequest.map(_ \/> "Could not find an active subscription"))
       sub <- EitherT(Future.successful(maybeDeliverySub.asDelivery \/> "Is not a Home Delivery subscription"))
-      billCycleDay <- EitherT(tpBackend.zuoraService.getAccount(sub.accountId).map(_.billCycleDay).map(intToEither))
+      account <- EitherT(tpBackend.zuoraService.getAccount(sub.accountId).map(toScalaZRight))
```
johnduffell:
```scala
account <- EitherT(tpBackend.zuoraService.getAccount(sub.accountId).map(toScalaZRight))
```
could be
```scala
account <- EitherT(tpBackend.zuoraService.getAccount(sub.accountId).map(\/-.apply))
```
or am I missing something?

johnduffell: ah maybe the type of the left is also important?

app/views/support/BillingScheduleOps.scala
```diff
@@ -5,6 +5,6 @@ import scalaz.syntax.std.boolean._
 object BillingScheduleOps {
   implicit class InvoiceOps(bill: Bill) {
     def isDiscounted(str: String): Option[String] =
-      bill.items.list.exists(_.amount < 0).option(str)
+      (bill.accountCredit.isDefined || bill.items.list.exists(_.amount < 0)).option(str)
```
johnduffell: isDiscounted doesn't return a boolean, which is surprising

johnduffell: I'd prefer to separate the boolean determining logic from the massaging types, so even if we want to keep the behaviour of this method, put it into a val and then use the val on the next line

johnduffell:
looks good, just a small comment about naming. I think the first comment might be just be the compiler needing keeping happy. 👍

PR #776, opened 2017-01-25
Fix the green billing schedule for non-discounted bills

paulbrown1982:
It should only be the first...
cc @johnduffell @jacobwinch @pvighi @AWare 

Before:
picture 3
After:
picture 4


jacobwinch:
👍

johnduffell:
oh dear, I did think we shouldn't have two representations for no credit.  we should have either None or 0

PR #777, opened 2017-01-25
stop the excess commas in delivery address

johnduffell:
with thanks to @pvighi for pointing out my lazy coding
from
image
to
image


pvighi:
👍

PR #778, opened 2017-01-25
Flow

AWare:
Now we can use flow in intellij and in es6/jsx modules. 
I've added it to the DirectDebit component. (and changed it into a stateless functional component) 

We can gradually add it throughout our ES6 code. 

It doesn't work with our AMD modules yet.


paulbrown1982:
👍 Nice!

PR #779, opened 2017-01-26
One off auto renew

pvighi:
When renewing a weekly subscription, don't set it to autoRenew unless the plan we are renewing it to is recurring.
This PR also has some small refactoring to remove duplication and avoid using isInstanceOf in every place of the code we checked if a plan is one off or recurring 
@johnduffell @paulbrown1982 @jacobwinch

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -275,8 +275,6 @@ class CheckoutService(identityService: IdentityService,
     val customerAcceptance = contractEffective
 
     def addPlan(contact: Contact) = {
-
-
       val newRatePlan = RatePlan(renewal.plan.id.get, None)
 
       val renewCommand = Renew(
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -285,7 +285,11 @@ class CheckoutService(identityService: IdentityService,
         contractEffectiveDate = contractEffective,
         customerAcceptanceDate = customerAcceptance,
         promoCode = None,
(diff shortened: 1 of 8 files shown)


app/services/CheckoutService.scala
```diff
@@ -276,7 +276,17 @@ class CheckoutService(identityService: IdentityService,
...
+
+      val renewCommand = Renew(
+        subscriptionId = subscription.id.get,
+        currentTermStartDate = subscription.termStartDate,
+        currentTermEndDate = subscription.termEndDate,
+        newRatePlans = NonEmptyList(newRatePlan),
+        contractEffectiveDate = contractEffective,
+        customerAcceptanceDate = customerAcceptance,
+        promoCode = None,
+        autoRenew = renewal.plan.charges.billingPeriod.isInstanceOf[RecurringPeriod])
```
pvighi: maybe it's my java past but I always want to make autoRenew a def in the Billing period trait instead of asking if something implements a trait like. But I guess this is the same as doing pattern matching, thoughts?

johnduffell: not quite, I think this code is doing a type check, wheras pattern matching checks at compile time if it's a valid check and whether it's exhaustive (assuming it's a sealed trait)
both are valid types of polymorphism, but one is inheritence and the other is extending from outside

johnduffell:
great! 👍 
cc @adamnfish this is where the whole debate able isInstanceOf vs pattern match came from

PR #781, opened 2017-01-26
Use the new wildcard cert for *.thegulocal.com, clean up Nginx config

rtyley:
Relying on the new CA-signed wildcard-cert that's provided by https://github.com/guardian/identity-platform/pull/149 means we can delete some code - there's no point in us having different certificates for our dev environment, and this matches the approach used in Membership, reducing some of the setup cost if you're working on both projects.

This is a repeat for Subs of these two Membership PRs:

* https://github.com/guardian/membership-frontend/pull/1441
* https://github.com/guardian/membership-frontend/pull/1447

cc @paulbrown1982 @mario-galic @AWare

### DIFF
--- a/nginx/setup.sh
+++ b/nginx/setup.sh
@@ -1,18 +1,9 @@
 #!/bin/bash
 
-GU_KEYS="${HOME}/.gu/keys"
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 NGINX_HOME=$(nginx -V 2>&1 | grep 'configure arguments:' | sed 's#.*conf-path=\([^ ]*\)/nginx\.conf.*#\1#g')
 
-echo this script needs root access and membership AWS credentials
-echo configuring nginx, please enter your sudo password if prompted
+echo this script needs root access to configure nginx, please enter your sudo password if prompted
 sudo mkdir -p $NGINX_HOME/sites-enabled
 sudo ln -fs $DIR/subscribe.conf $NGINX_HOME/sites-enabled/subscribe.conf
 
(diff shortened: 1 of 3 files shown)


nginx/setup.sh
```diff
@@ -1,18 +1,9 @@
...
-echo configuring nginx, please enter your sudo password if prompted
+echo this script needs root access to configure nginx, please enter your sudo password if prompted
 sudo mkdir -p $NGINX_HOME/sites-enabled
 sudo ln -fs $DIR/subscribe.conf $NGINX_HOME/sites-enabled/subscribe.conf
 
-aws s3 cp s3://identity-local-ssl/sub-thegulocal-com-exp2017-03-31-bundle.crt ${GU_KEYS}/ --profile membership
-aws s3 cp s3://identity-local-ssl/sub-thegulocal-com-exp2017-03-31.key ${GU_KEYS}/ --profile membership
-sudo ln -fs ${GU_KEYS}/ $NGINX_HOME/keys
-
-sudo nginx -s stop
```
rtyley: The auto-start of Nginx with `sudo nginx` is removed here because it actually causes an inconsistent state on Ubuntu 16.10, where the Nginx PID file can not be created.

rtyley:
This failure looks unrelated to my change :-(

image

https://travis-ci.org/guardian/subscriptions-frontend/builds/196902949#L632

rtyley:
The error looks like this on the Screencast:

image

I don't see this error myself on going to https://subscribe.theguardian.com/checkout - but I also am not a test-user (see #783), so it's possible that is part of the problem.

jacobwinch:
Sentry issue is here: https://sentry.io/the-guardian/subscriptions/issues/181087071/events/4866537246/

It looks test-user specific and unrelated to the changes in this PR.

PR #783, opened 2017-01-31
Use current Google Groups for auth, now that `subscriptions.dev` is dead

rtyley:
Looks like the Google Group for `"subscriptions.dev@guardian.co.uk"` was deleted (not sure when or why). I need access to https://subscribe.theguardian.com/test-users, and `membership.dev@guardian.co.uk` is the only group I'm sure I'm on, so switching to that!

cc @paulbrown1982 @johnduffell

### DIFF
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -23,7 +23,7 @@ object CommonActions {
 
   val AuthorisedTester = GoogleAuthenticatedStaffAction andThen requireGroup[GoogleAuthRequest](Set(
     "directteam@guardian.co.uk",
-    "subscriptions.dev@guardian.co.uk",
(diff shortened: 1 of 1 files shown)


app/actions/CommonActions.scala
```diff
@@ -23,7 +23,7 @@ object CommonActions {
 
   val AuthorisedTester = GoogleAuthenticatedStaffAction andThen requireGroup[GoogleAuthRequest](Set(
     "directteam@guardian.co.uk",
-    "subscriptions.dev@guardian.co.uk",
```
paulbrown1982: Please can you leave that in, and add membership.dev additionally? It was just that you were removed from subscriptions.dev - there might be others of us who are on that and not in membership.dev!

PR #784, opened 2017-01-31
Fix TestUser sensitivity to clock drift which had been throwing exceptions

rtyley:
The acceptance tests were intermittently failing - this was happened when the TestUser token was stamped with a time ahead of the clock time on the `subscriptions-frontend` appserver.

    "IllegalArgumentException: The end instant must be greater the start"

See https://github.com/guardian/identity-test-users/commit/410a7fec for the details of the fix, and accompanying tests that check for tolerance of clock drift.

I noticed this problem with https://github.com/guardian/subscriptions-frontend/pull/781#issuecomment-276336674, but [looking at Sentry](https://sentry.io/the-guardian/subscriptions/issues/181087071/events/4866537246/) there are quite a few errors of this format.

cc @jacobwinch

### DIFF
--- a/app/utils/TestUsers.scala
+++ b/app/utils/TestUsers.scala
@@ -1,6 +1,7 @@
 package utils
 
-import com.github.nscala_time.time.Imports._
+import java.time.Duration.ofDays
+
 import com.gu.identity.play.IdMinimalUser
 import com.gu.identity.testing.usernames.TestUsernames
 import configuration.Config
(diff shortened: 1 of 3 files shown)


jacobwinch:
👍

paulbrown1982:
👍 👍 nice spot

rtyley:
Well, the post-deployment testing passed, which was the intention!

image

Hopefully, this error will not re-occur.

PR #786, opened 2017-02-01
Inital weekly landing page stuff (wip)

AWare:
Requires:
https://github.com/guardian/membership-common/pull/416

Adds a landing page for weekly and makes a small change to prettyPricingForDiscountedPeriod

### DIFF
--- a/app/views/support/DiscountWeeklyRegions.scala
+++ b/app/views/support/DiscountWeeklyRegions.scala
@@ -1,0 +1,80 @@
+package views.support
+
+import com.gu.i18n.{Country, CountryGroup, Currency}
+import com.gu.memsub.Subscription
+import com.gu.memsub.promo.Promotion.PromoWithWeeklyLandingPage
+import com.gu.memsub.promo.{PromoCode, WeeklyLandingPage}
+import com.gu.memsub.subsv2.{Catalog, CatalogPlan}
+import com.netaporter.uri.Uri
+import com.netaporter.uri.dsl._
+import views.support.Pricing._
+//TODO: move this and revert to the old
+
+case class DiscountWeeklyRegion(title: String, description: String, url: Uri, ratePlans: Set[Subscription.ProductRatePlanId], countries: Set[Country]) {
+}
+
+object DiscountWeeklyRegion {
+
+  private val UK = CountryGroup.UK.countries.toSet
+  private val US = CountryGroup.US.countries.toSet
+  private val ZONEC = CountryGroup.allGroups.flatMap(_.countries).toSet.diff(UK union US)
+
+  def all(catalog: Catalog): List[DiscountWeeklyRegion] = List(DiscountWeeklyRegion(
+    title = "United Kingdom",
+    description = "Includes Isle of Man and Channel Islands",
+    url = Uri.parse(s"checkout/${catalog.weeklyZoneA.quarter.slug}").addParam("countryGroup", "uk"),
+    ratePlans = catalog.weeklyZoneA.toList.map(_.id).toSet,
+    countries = UK
+  ), DiscountWeeklyRegion(
+    title = "United States",
+    description = "Includes Alaska and Hawaii",
+    url = Uri.parse(s"checkout/${catalog.weeklyZoneA.quarter.slug}").addParam("countryGroup", "us"),
+    ratePlans = catalog.weeklyZoneA.toList.map(_.id).toSet,
+    countries = US
+  ), DiscountWeeklyRegion(
+    title = "Rest of the world",
+    description = "Posted to you by air mail",
+    url = Uri.parse(s"checkout/${catalog.weeklyZoneC.quarter.slug}"),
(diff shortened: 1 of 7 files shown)


app/views/support/Pricing.scala
```diff
@@ -21,10 +24,9 @@ object Pricing {
     def prettyPricing(currency: Currency) =
       s"${unsafePrice(currency).pretty} ${plan.billingPeriod.frequencyInMonths}"
 
-    def amount(currency: Currency) =
-      unsafePrice(currency).amount
+    def amount(currency: Currency) = unsafePrice(currency).amount
 
-    def prettyPricingForDiscountedPeriod(discountPromo: Promotion[PercentDiscount, Option, LandingPage], currency: Currency) = {
+    def prettyPricingForDiscountedPeriod[M[+A], A <: LandingPage](discountPromo: Promotion[PercentDiscount,M, A], currency: Currency) = {
```
AWare: This *was* parametrised in Option, which meant that we could not use it with a Promotion once it had fetched its landing page. This is the same type parameterisation as used in the Promotion case class in membership-common. 

It's not obvious that we should split Promotions into Option of Landing Page and Identity of LandingPage.

app/model/WeeklyRegions.scala
```diff
@@ -1,34 +1,75 @@
...
+
+    def plans:List[CatalogPlan.Paid] = catalog.weekly.flatten.filter(plan => promotion.appliesTo.productRatePlanIds.contains(plan.id))
+
+    for {
+      plan <- plans
+      discountPromo <- promotion.asDiscount.toList
+      currency <- plan.charges.currencies.toList
+      if (cs.contains(currency))
+    } yield {
+      plan.charges.prettyPricingForDiscountedPeriod[scalaz.Id.Id, WeeklyLandingPage](discountPromo, currency)
```
AWare: Id.Id is even more basic and instinctual than Id alone, and should be replaced with ~the superego~ something that doesn't conflict with Option so badly. 

We have to parameterise the function call here because Id.Id is the identity type, and the scala compiler needs to be persuaded that the emperor has new clothes, otherwise it will consider the "container" type of the LandingPage to be LandingPage. 

Thanks to @philwills for figuring out the types involved in this.  (And @johnduffell for helping me remain patient when trying to figure it out to begin with)

test/model/PrettyPricingForDiscountTest.scala
```diff
@@ -0,0 +1,8 @@
+package model
+
+/**
+  * Created by alex_ware on 03/02/2017.
```
paulbrown1982: No need for this as we have git!

assets/javascripts/modules/dropdown.js
```diff
@@ -0,0 +1,77 @@
...
+    function hideAllMenus() {
+        $(MENU_SELECTOR).addClass(HIDDEN_CLASS);
+    }
+
+    function show($elem) {
+        $elem.removeClass(HIDDEN_CLASS);
+    }
+
+    function init() {
+        console.log('hi');
```
paulbrown1982: Delete me!

assets/javascripts/modules/dropdown.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    function isEnabled(dropdown) {
+        return !$(dropdown).hasClass(DROPDOWN_DISABLED_CLASS);
+    }
+
+    // bonzo doesn't provide this so had to write my own
+    function closest(elem, className) {
+        do {
+            console.log(elem);
```
paulbrown1982: And me

app/views/support/DiscountWeeklyRegions.scala
```diff
@@ -0,0 +1,78 @@
...
+    val promotionCountries = promotion.appliesTo.countries
+
+    all(catalog).filter { region =>
+      val ratePlaninPromotion = promotionProductRatePlanIds.exists(plan => region.ratePlans.contains(plan))
+      val countryInPromotion = !region.countries.intersect(promotionCountries).isEmpty
+      ratePlaninPromotion && countryInPromotion
+    }
+  }
+
+  private def currencyFor(country: Country): Option[Currency] = CountryGroup.availableCurrency(Currency.all.toSet)(country)
```
paulbrown1982: A tiny performance tweak could be to store Currency.all.toSet in a val closing into this def.

paulbrown1982:
👍 looks good

assets/javascripts/modules/dropdown.js
johnduffell: I'm pretty sure on frontend we had a thing to fail linting if there were any console.log left in (becuse I did it all the time!) might be a time saver for subs too?

app/views/support/DiscountWeeklyRegions.scala
paulbrown1982: Sorry, ignore me, I didn't spot that this was a curried function.

paulbrown1982:
👍

PR #787, opened 2017-02-02
use labelled catalog from zuora

johnduffell:
use the labelled catalog, this means we can add 6 for 6 to the catalog and distinguish it from normal quarterly plans

@pvighi @paulbrown1982 @jacobwinch @AWare

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.353",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -43,7 +43,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.353",
+    "com.gu" %% "membership-common" % "0.354-SNAPSHOT",
```
johnduffell: https://github.com/guardian/membership-common/pull/419

PR #790, opened 2017-02-08
use membership common that understands the free id

johnduffell:
https://github.com/guardian/membership-common/pull/423 understand free id
edit:
https://github.com/guardian/membership-common/pull/424 don't read the untagged rateplans
https://github.com/guardian/membership-common/pull/425 use billing period as an object not a case class


jacobwinch:
👍

PR #792, opened 2017-02-10
Add Facebook pixel to checkout and thank you page

paulbrown1982:
Also removed the old TagManager, and removed JellyFish's Facebook pixel so as not to cause conflict - they weren't using it.

cc @johnduffell @pvighi @jacobwinch @AWare


johnduffell:
👍

PR #793, opened 2017-02-13
Create an "current offers" page and better handling for tracking codes

paulbrown1982:
- Cloned the UK and International homepages into /uk/offers and /int/offers page respectively, and removed the real promotions from the original home page.
- Added better redirect logic for the /p/<promo_code> page such that it will redirect to the promotion's relative page providing it has a hint within the list of appliesTo rate plans on where to land. Otherwise it falls back to the homepage as before.
- Fixed a few deprecation warnings by using the new Future-based promotion lookup

cc @jacobwinch @johnduffell @AWare @pvighi

### DIFF
--- a/app/controllers/PromoLandingPage.scala
+++ b/app/controllers/PromoLandingPage.scala
@@ -1,13 +1,16 @@
 package controllers
 import actions.CommonActions._
 import com.gu.i18n.{Country, CountryGroup}
+import com.gu.i18n.Country
+import com.gu.memsub.Digipack
 import com.gu.memsub.promo.Formatters.PromotionFormatters._
 import com.gu.memsub.promo.Promotion._
 import com.gu.memsub.promo._
 import com.netaporter.uri.dsl._
 import configuration.Config
+import controllers.SessionKeys.PromotionTrackingCode
 import filters.HandleXFrameOptionsOverrideHeader
-import model.DigitalEdition
+import model.DigitalEdition.UK
 import play.api.data.{Form, Forms}
 import play.api.libs.json.Json
 import play.api.mvc._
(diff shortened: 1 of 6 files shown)


app/controllers/PromoLandingPage.scala
```diff
@@ -1,61 +1,108 @@
...
+  private val digitalPackRatePlanIds = catalog.digipack.toList.map(_.id).toSet
+  private val allPaperPackages = catalog.delivery.list ++ catalog.voucher.list
+  private val paperOnlyPackageRatePlanIds = allPaperPackages.filter(_.charges.benefits.list.contains(Digipack)).map(_.id).toSet
+  private val paperPlusPackageRatePlanIds = allPaperPackages.filterNot(_.charges.benefits.list.contains(Digipack)).map(_.id).toSet
+  private val guardianWeeklyRatePlanIds = catalog.weeklyZoneA.toList.map(_.id).toSet ++ catalog.weeklyZoneC.toList.map(_.id).toSet
 
   private def isActive(promotion: AnyPromotion): Boolean = {
     val isTest = tpBackend.environmentName != "PROD"
-    isTest || (promotion.starts.isBeforeNow && !promotion.expires.exists(_.isBeforeNow))
+    isTest || (
```
johnduffell: kind of feel like this should be in the promotion itself, as other parts of the code would need to do this too

conf/routes
```diff
@@ -78,5 +78,9 @@ GET         /s/:supplierCodeStr              controllers.Homepage.supplierRedire
 # Paths from legacy redirects
 GET         /Voucher                         controllers.Homepage.index
 
+# Offers page or 404
+GET         /offers                          controllers.Offers.offers
+GET         /:edition/offers                 controllers.Offers.offersPage(edition: String)
```
johnduffell: the "or 404" is just if you put in an invalid edition right? so we probably don't need to mention here

johnduffell:
looks good to me! 👍 😃

PR #801, opened 2017-02-15
Fixed bug where the Digital Pack's edition hint was dropped from the …

paulbrown1982:
Fixed bug where the Digital Pack's edition hint was dropped from the international version of the homepage, as it now uses tracking promo codes which don't support it. I've added it back in via a query parameter, which overrides the GEO-IP logic in the DigitalPack redirect action.

cc @jacobwinch @AWare

### DIFF
--- a/app/controllers/DigitalPack.scala
+++ b/app/controllers/DigitalPack.scala
@@ -10,13 +10,20 @@ import utils.RequestCountry._
 
 object DigitalPack extends Controller {
 
-  private def redirectResult(digitalEdition: DigitalEdition)(implicit request: Request[AnyContent]) =
-    Redirect(routes.DigitalPack.landingPage(digitalEdition.id).url, request.queryString, SEE_OTHER)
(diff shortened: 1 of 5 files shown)


app/controllers/DigitalPack.scala
```diff
@@ -10,13 +10,20 @@ import utils.RequestCountry._
 
 object DigitalPack extends Controller {
 
-  private def redirectResult(digitalEdition: DigitalEdition)(implicit request: Request[AnyContent]) =
-    Redirect(routes.DigitalPack.landingPage(digitalEdition.id).url, request.queryString, SEE_OTHER)
+  private val queryParamHint = "edition"
+
+  private def redirectResult(digitalEdition: DigitalEdition)(implicit request: Request[AnyContent]) = {
+    val queryString = request.queryString.-(queryParamHint)
```
AWare: Weird use of .-()

Otherwise :+1:

PR #803, opened 2017-02-15
Weekly animation tweaks

AWare:
Add a 'from' price on the package, which animates away. 

image
image

### DIFF
--- a/app/views/support/Pricing.scala
+++ b/app/views/support/Pricing.scala
@@ -65,11 +65,8 @@ object Pricing {
     def headlinePricingForDiscountedPeriod[M[+A], A <: LandingPage](discountPromo: Promotion[PercentDiscount, M, A], currency: Currency) = {
       val originalAmount = unsafePrice(currency)
       val discountAmount = discountPromo.promotionType.applyDiscount(originalAmount, plan.billingPeriod)
-      discountPromo.promotionType.durationMonths.fold {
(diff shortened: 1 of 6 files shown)


app/views/support/Pricing.scala
```diff
@@ -62,7 +62,17 @@ object Pricing {
       }
     }
 
+    def headlinePricingForDiscountedPeriod[M[+A], A <: LandingPage](discountPromo: Promotion[PercentDiscount, M, A], currency: Currency) = {
+      val originalAmount = unsafePrice(currency)
+      val discountAmount = discountPromo.promotionType.applyDiscount(originalAmount, plan.billingPeriod)
+      discountPromo.promotionType.durationMonths.fold {
```
paulbrown1982: "From" could just be made an Option[String] with .mkString used in the s"..." so you don't have to repeat the full String template, or use a .fold on durationMonths

assets/javascripts/modules/animatedDropdown.es6
```diff
@@ -1,4 +1,4 @@
-const SHOW_CLASS = 'js-show';
+    const SHOW_CLASS = 'animate-show-shown';
```
paulbrown1982: Whitespace could be tidied here

paulbrown1982:
👍 there's some minor code tweaks you could make but they're not serious enough to block deployment if anything (like a demo etc) is waiting on this.

paulbrown1982:
👍

PR #804, opened 2017-02-15
Six for six

pvighi:
Simplest implementation to have something to work with.

major things we still need to fix:
* The 6 for 6 plan is visible to everyone in this version, we should be able to hide it unless a specific promotion is active
* Management shows the incorrect plan information sometimes 
*  In many places it refers to the plan as six issues o six weeks one off plan without mentioning the quarterly recurring plan that comes after

@johnduffell @paulbrown1982 @AWare

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -70,8 +70,8 @@ class CheckoutService(identityService: IdentityService,
 
     val updatedCommand = additionalRateplan.map { ratePlantoAdd =>
       val updatedRatePlans = ratePlantoAdd <:: originalCommand.ratePlans
-      val contractEffective = DateTime.now.toLocalDate.plusDays(defaultPaymentDelayt.getDays)
-      originalCommand.copy(ratePlans = updatedRatePlans, contractEffective = contractEffective, contractAcceptance = contractEffective.plusDays(42))
+      val updatedContractEffective = DateTime.now.toLocalDate.plusDays(defaultPaymentDelayt.getDays)
+      originalCommand.copy(ratePlans = updatedRatePlans, contractEffective = updatedContractEffective, contractAcceptance = updatedContractEffective.plusDays(42))
     }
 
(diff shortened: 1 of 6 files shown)


app/services/CheckoutService.scala
```diff
@@ -61,6 +61,22 @@ class CheckoutService(identityService: IdentityService,
     case _ => false
   }
 
+  def processIntroductoryPeriod(defaultPaymentDelayt: Days, originalCommand: Subscribe): Subscribe = {
```
johnduffell: defaultPaymentDelayt
what's the 't' for?

app/services/CheckoutService.scala
```diff
@@ -61,6 +61,22 @@ class CheckoutService(identityService: IdentityService,
...
+    val additionalRateplan = (originalCommand.ratePlans.head.productRatePlanId match {
+      case catalog.weeklyZoneA.sixWeeks.id.get => Some(catalog.weeklyZoneA.quarter)
+      case catalog.weeklyZoneC.sixWeeks.id.get => Some(catalog.weeklyZoneC.quarter)
+      case _ => None
+    }).map(plan => RatePlan(plan.id.get, None))
```
johnduffell: not here, but I don't like that we use `get` rather than `value` as it makes it more likely we'll miss a `None.get` in the code

johnduffell:
sure looks good to me! 👍

app/services/CheckoutService.scala
```diff
@@ -61,6 +61,22 @@ class CheckoutService(identityService: IdentityService,
...
+
+    val updatedCommand = additionalRateplan.map { ratePlantoAdd =>
```
johnduffell: I did wonder what a 'planto' was for a minute! 😝

app/services/CheckoutService.scala
pvighi: t for typo

PR #806, opened 2017-02-16
Remove `InsecureOkHttpClient` as we have proper certs in dev now

rtyley:
The `InsecureOkHttpClient` used in acceptance tests is unnecessary now, because we have a valid CA-signed certificate.

For this to work, update your `identity-platform` to the latest version (which includes https://github.com/guardian/identity-platform/pull/154 ) and re-run `./ngnix/setup.sh`.

This is an echo of https://github.com/guardian/membership-frontend/pull/1510

cc @AWare @Ap0c


paulbrown1982:
👍

PR #807, opened 2017-02-16
Weekly delay

AWare:
https://trello.com/c/jN1U5NUz/124-bug-guardian-weekly-date-picker

image

Also fixes a bug where `Option.map("string")` was of type `Option[char]`


AWare:
(removed multi-const on recommendation of our former colleague) 
http://breck-mckye.com/blog/2017/01/the-javascript-single-var-style-is-an-antipattern/

PR #809, opened 2017-02-17
Weekly discounted regions

AWare:
Avoids repeatedly calling `DiscountWeeklyRegions.plans` by moving this into the Discounted Region. This needed a bit of a refactor, and I think has resulted in simpler code. 

Having this also made it easier to make the region act as a link when only one rate plan was valid for that region.


app/views/support/WeeklyPromotion.scala
```diff
@@ -0,0 +1,85 @@
...
+
+object WeeklyPromotion {
+  private val UK = CountryGroup.UK.countries.toSet
+  private val US = CountryGroup.US.countries.toSet
+  private val ZONEC = CountryGroup.allGroups.flatMap(_.countries).toSet.diff(UK union US)
+  case class DiscountedPlan(pretty: String, headline: String, period: String, url: Uri)
+
+  case class DiscountedRegion(title: String, description: String, countries: Set[Country], discountedPlans: List[DiscountedPlan])
+
+  def forPromotion(promotion: PromoWithWeeklyLandingPage, promoCode: PromoCode, requestCountry: Country)(implicit catalog: Catalog): Seq[DiscountedRegion] = {
```
johnduffell: this is actually "discounted regions" for promotion

app/views/support/WeeklyPromotion.scala
```diff
@@ -0,0 +1,85 @@
...
+    def plans: List[CatalogPlan.Paid] =
+      catalogPlans.filter(catalogPlan => promotion.appliesTo.productRatePlanIds.contains(catalogPlan.id)).toList.sortBy(_.charges.gbpPrice.amount)
+
+    for {
+      plan <- plans
+      currency <- plan.charges.currencies.toList
+      if (currencies.contains(currency))
+    } yield {
+      val pretty = promotion.asDiscount
+        .map { discountPromo => plan.charges.prettyPricingForDiscountedPeriod[scalaz.Id.Id, WeeklyLandingPage](discountPromo, currency) }
```
johnduffell: usually I'd lay this out as
```scala
blah.map { value =>
  value.something
}
```

johnduffell:
great, the code doesn't do the result justice, you should put gifs in your PRs (use licecap or something) to show what it actually looks like! 🎆 🌟 
👍

AWare:
anim

PR #813, opened 2017-02-20
add release to raven

AWare:
this might make sentry a little better to use?

other option: use the webhook


jacobwinch:
It looks like we are already including the release number on Sentry issues (as a tag) - do we get any extra benefits by adding this as well?

AWare:
Sentry does release tagging, which might be helpful to help track js issues.
https://sentry.io/the-guardian/subscriptions-client-side/settings/release-tracking/

jacobwinch:
Cool, thanks - sounds good 👍

PR #817, opened 2017-02-24
Support for empty billing schedule in /manage

pvighi:
A quick fix for the problem where a subscription with no future payments scheduled cannot log in to the manage page.

Maybe a longer term solution for this could be to redefine billing schedules so that they allow empty schedules.

@johnduffell


johnduffell:
great, thanks for that! Agreed about the longer term fix, save having an Option[NonEmptyList] by having a List 👍

PR #820, opened 2017-02-24
fix JS errors when there's no default plan (zone c)

johnduffell:
So for zone C, the default plan for zone C might not have the same currency as the assumed currency in the backend.. This results in no plan appearing on the right hand side until a country is selected (which is fine) but then the promocode stuff throws an error.

This just adds a check so it won't trigger until there is a rate plan selected.

@AWare @jacobwinch @pvighi


johnduffell:
suddenly I wonder whether this is caused by the same issue where the zone a appears as UK 6 pounds and us 6 pounds (when it should be 6 dollars) so perhaps this is the wrong fix ... :thinking:

johnduffell:
closing as updating the catalog and redeploying has picked up the right currencies

PR #821, opened 2017-02-27
Six for six fixes

pvighi:
* Removed 6 for 6 from renewal options
* In /manage show a list of all current plans and a list of all upcoming plans. A plan will be considered current only if the start date is not in the future. 
* added duration to pretty pricing of one off plans. For example for the 1 year plan will now say "US$240 one off payment to cover 1 year" instead of just "US$240 one off payment"
* fixed problem with welcome email not being sent to 6 for 6 subscriptions and added the quarterly payments to the description. It will say now  something like "£6 one off payment to cover 6 weeks then £30 every 3 months"

@paulbrown1982 @johnduffell


app/model/SubscriptionOps.scala
```diff
@@ -36,6 +45,16 @@ object SubscriptionOps extends LazyLogging {
         nextPlan
       }
     }
+
+    def currentPlans = subscription.plans.list.filter(p => (p.start == now || p.start.isBefore(now)) && p.end.isAfter(now))
```
pvighi: This is what we use to separate current and upcoming plans in /manage. Should we check this in a different way?

johnduffell: you mean for the "last day is exclusive in zuora" problem?

PR #822, opened 2017-02-27
Fixed description of plans to which a promotion is applicable on promotion terms page

paulbrown1982:
Fixed description of plans to which a promotion is applicable on promotion terms page. I've tested it for other products like newspaper and digital pack and it looks good there too.

Before:

picture 21

After:

picture 20

cc @johnduffell @pvighi @jacobwinch @AWare


jacobwinch:
👍

PR #823, opened 2017-02-27
Six for six effective date

pvighi:
Make the two rate plans in the sub not overlap
@paulbrown1982 @jacobwinch

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -61,7 +61,7 @@ class CheckoutService(identityService: IdentityService,
     case _ => false
   }
 
-  def processIntroductoryPeriod(defaultPaymentDelayt: Days, originalCommand: Subscribe): Subscribe = {
+  def processIntroductoryPeriod(defaultPaymentDelay: Days, originalCommand: Subscribe): Subscribe = {
     val additionalRateplan = (originalCommand.ratePlans.head.productRatePlanId match {
       case catalog.weeklyZoneA.sixWeeks.id.get => Some(catalog.weeklyZoneA.quarter)
       case catalog.weeklyZoneC.sixWeeks.id.get => Some(catalog.weeklyZoneC.quarter)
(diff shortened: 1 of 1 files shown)


app/services/CheckoutService.scala
```diff
@@ -70,8 +70,12 @@ class CheckoutService(identityService: IdentityService,
 
     val updatedCommand = additionalRateplan.map { ratePlantoAdd =>
       val updatedRatePlans = ratePlantoAdd <:: originalCommand.ratePlans
-      val updatedContractEffective = DateTime.now.toLocalDate.plusDays(defaultPaymentDelayt.getDays)
-      originalCommand.copy(ratePlans = updatedRatePlans, contractEffective = updatedContractEffective, contractAcceptance = updatedContractEffective.plusDays(42))
+
+      val defaultDelayDays = defaultPaymentDelay.getDays
+      val introductoryPeriodDelayDays = if (defaultDelayDays > 0) defaultDelayDays - 1  else defaultDelayDays
```
pvighi: maybe we don't need to check if defaultDays>0. I think it should always be

app/services/CheckoutService.scala
```diff
@@ -70,8 +70,12 @@ class CheckoutService(identityService: IdentityService,
...
+
+      val updatedContractEffective = DateTime.now.toLocalDate.plusDays(introductoryPeriodDelayDays)
+      originalCommand.copy(ratePlans = updatedRatePlans, contractEffective = updatedContractEffective, contractAcceptance = updatedContractEffective.plusDays(43))
```
paulbrown1982: 👎 this will break for 12-for-12. Is there somewhere in the model you can get the 6 or 12 number from? Then you can do something like:
```
.plusWeeks(sub.plan.product.introductoryPeriodWeeks).plusDays(1)
```

app/services/CheckoutService.scala
jacobwinch: Yeah I can't think of a scenario where defaultPaymentDelay is going to be < 0

PR #826, opened 2017-02-28
Fix /manage for expired subs

jacobwinch:
By converting these vals to defs, we avoid calling .plan on an expired sub.

@pvighi


jacobwinch:
This will fix: https://sentry.io/the-guardian/subscriptions/issues/226280665/

pvighi:
👍

PR #829, opened 2017-02-28
Weekly lp launch tweaks

AWare:
- don't display the promotion description if we don't have landing page copy
- do display the default copy
- don't display so many currencies for row
- todo update default image


app/views/support/PlanOps.scala
```diff
@@ -37,7 +37,7 @@ object PlanOps {
         )
       case Weekly :: Nil =>
         ResponsiveImageGroup(
-          availableImages = ResponsiveImageGenerator("1961260fc68598c31c0b882d8f4da8e8ec34e7d0/0_0_1000_1333", Seq(375, 750, 1000), "png"),
```
paulbrown1982: cc @tomverran 😉

tomverran: I'm impressed that image survived this long!

PR #830, opened 2017-02-28
in checkout page show associated plan as a "and then"

johnduffell:
This is needed to sell 6 for 6

Before:::
image

After:::

image

depends on https://github.com/guardian/membership-common/pull/440
@paulbrown1982 

I still need to fix the thank you page though cc @pvighi


pvighi:
looks good to me 👍

johnduffell:
ah yes I've broken the available for checkout code, fixing it now

PR #832, opened 2017-03-01
Proper name for Guardian Weekly print subscription services

paulbrown1982:
cc @jacobwinch


jacobwinch:
Just noticed that this got merged into the ensure-correct-tsandcs-on-guardian-weekly-checkout branch (not master)

PR #840, opened 2017-03-01
Hide tracking promos

AWare:
Don't show promo material in checkout for a tracking promo.


AWare:
@paulbrown1982 promo is passed to zuora

PR #843, opened 2017-03-02
A few small copy tweaks to the Guardian Weekly checkout page

paulbrown1982:
- A few small tweaks to the Guardian Weekly checkout page to remove unnecessary bits of copy, whitespace etc.
- Renamed the product to The Guardian Weekly instead of Guardian Weekly
- Removed the repeating product name from each payment frequency (making GW like Digital Pack)
- Refactored the 'and then' generators into one helper macro.

Before:
picture 30

After:
picture 31

cc @johnduffell @AWare @pvighi @jacobwinch

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -24,24 +24,28 @@
     countryAndCurrencySettings: CountryAndCurrencySettings
 )(implicit request: RequestHeader)
 
+@andThenPricing(currency: Currency, associatedPlan: Option[CatalogPlan.Paid]) = {
+    @for(ap <- associatedPlan.map(_.charges.prettyPricing(currency))) {
+        and then @ap
+    }
+}
 @priceOption(plan: CatalogPlan.Paid, currency: Currency, associatedPlan: Option[CatalogPlan.Paid]) = {
     <label class="option" data-currency="@currency">
         <span class="option__input">
             <input
             type="radio" name="ratePlanId"
             data-option-label-prefix="@plan.prefix"
-            data-option-mirror-payment-default="@plan.charges.prettyPricing(currency) @for(ap <- associatedPlan.map(_.charges.prettyPricing(currency))) { and then @ap}"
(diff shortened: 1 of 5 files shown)


app/views/checkout/payment.scala.html
```diff
@@ -24,24 +24,28 @@
     countryAndCurrencySettings: CountryAndCurrencySettings
 )(implicit request: RequestHeader)
 
+@andThenPricing(currency: Currency, associatedPlan: Option[CatalogPlan.Paid]) = {
+    @for(ap <- associatedPlan.map(_.charges.prettyPricing(currency))) {
+        and then @ap
```
johnduffell: good idea!

johnduffell: doesn't even need to be a def, it could be a val

johnduffell:
👍 great!

app/views/support/Pricing.scala
```diff
@@ -57,7 +57,7 @@ object Pricing {
               s"${discountAmount.pretty} for 1 year, then ${originalAmount.pretty} every year thereafter"
             }
           case OneYear => s"${discountAmount.pretty} for 1 year"
-
+          case _ => ""
```
johnduffell: should it be "no discount"? this is if someone applies a discount to a 6Weeks plan?

app/views/support/Pricing.scala
```diff
@@ -80,6 +80,7 @@ object Pricing {
 
     def prettyName(currency: Currency): String = in.charges.benefits.list match {
       case Digipack :: Nil => planWithPricing.prettyPricing(currency)
+      case Weekly :: Nil => planWithPricing.prettyPricing(currency)
```
johnduffell: could you have used 
```scala
case Digipack :: Nil | Weekly :: Nil => planWithPricing.prettyPricing(currency)
```
?

johnduffell: or even
```scala
case (Digipack | Weekly) :: Nil => planWithPricing.prettyPricing(currency)
```
but not sure if they would both compile

johnduffell:
still looks good although a couple of things I'm not 100% sure about see the comments
👍

PR #849, opened 2017-03-03
Fix weekly landing page dollar handling

AWare:
Make USD show by default if more than one currency. 
(Also minor refactoring and tidying)

### DIFF
--- a/app/views/support/WeeklyPromotion.scala
+++ b/app/views/support/WeeklyPromotion.scala
@@ -116,36 +116,55 @@ object WeeklyPromotion {
 
   private val currencyFor = CountryGroup.availableCurrency(Currency.all.toSet) _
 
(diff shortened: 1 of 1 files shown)


app/views/support/WeeklyPromotion.scala
```diff
@@ -116,36 +116,55 @@ object WeeklyPromotion {
...
-      case _ => false
+  def plansForPromotion(maybePromotion: Option[PromoWithWeeklyLandingPage], promoCode: Option[PromoCode], countries: Set[Country], catalogPlans: Seq[CatalogPlan.Paid])(implicit catalog: Catalog): List[DiscountedPlan] = {
+    val currencies = for {
+      promotion <- maybePromotion.toList
+      promoCountry <- promotion.appliesTo.countries
+      if (countries contains promoCountry)
+      currency <- currencyFor(promoCountry)
+    } yield currency
+    val currency = currencies match {
+      case _ if (currencies.size == 1) => currencies.head
```
paulbrown1982: Maybe try something like: 
case c:Currency :: Nil => c

PR #852, opened 2017-03-03
Fix weekly landing page dollar handling

AWare:
Reverts guardian/subscriptions-frontend#850

And properly deals with currency


paulbrown1982:
👍 Will merge so it's fixed for the weekend.

PR #855, opened 2017-03-06
Use next available shipping date as start date when renewing an expired sub.

jacobwinch:
This PR:

- Enables us to calculate the next available shipping date when renewing an expired sub 
- Pulls in membership-common changes here: https://github.com/guardian/membership-common/pull/446
- Sets all three dates (contract effective, customer acceptance, term start date) as the next available shipping date (when renewing an expired sub).

@paulbrown1982 @pvighi @johnduffell


jacobwinch:
A-S00068107 (in UAT) is an example of an expired sub which has been renewed, if anyone wants to check that the JSON looks as expected at the end of the process...

PR #856, opened 2017-03-06
Fix promotion currency look up (and manage page)

AWare:
We were getting 500 errors on promos where the plan did not have the desired currency. (mostly due to zone B)
https://sentry.io/the-guardian/subscriptions/issues/199120847/

also making the json response body into a def so it's not eagerly calculated for incorrect promos

(and fixing my last pr, I left a few fixes in my checkin)

### DIFF
--- a/app/controllers/Promotion.scala
+++ b/app/controllers/Promotion.scala
@@ -32,7 +32,7 @@ object Promotion extends Controller with LazyLogging with CatalogProvider {
     case class RatePlanPrice(ratePlanId: ProductRatePlanId, chargeList: PaidChargeList)
     promo.asDiscount.map { discountPromo =>
     catalog.allSubs.flatten
-        .filter(plan => promo.appliesTo.productRatePlanIds.contains(plan.id))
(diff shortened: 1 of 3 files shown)


app/controllers/Promotion.scala
```diff
@@ -31,10 +31,11 @@ object Promotion extends Controller with LazyLogging with CatalogProvider {
 
     case class RatePlanPrice(ratePlanId: ProductRatePlanId, chargeList: PaidChargeList)
     promo.asDiscount.map { discountPromo =>
-      catalog.allSubs.flatten
+    catalog.allSubs.flatten
         .filter(plan => promo.appliesTo.productRatePlanIds.contains(plan.id))
+        .filter(plan => plan.charges.currencies contains currency)
```
jacobwinch: .contains(currency) for consistency?

jacobwinch: Or use infix for both

jacobwinch: But using both styles next to each other makes me uncomfortable

jacobwinch:
👍

PR #858, opened 2017-03-07
Thank you page 6 for 6

pvighi:
* added support for subscriptions with multiple plans in the thank you page
* pass the newly created subscription to the thank you page instead of the catalog plan
* a bunch of refactoring

**6 for 6 plans**
before :
be6for6
 after: 
af6issues

**Regular plans**
before: 
bedig
after:
afdi

@johnduffell @paulbrown1982


app/views/checkout/thankyou.scala.html
```diff
@@ -1,34 +1,33 @@
...
 
-@main(s"Confirmation | Subscribe to the ${plan.title} | The Guardian", bodyClasses=List("is-wide"), touchpointBackendResolutionOpt = Some(touchpointBackendResolution), product = Some(plan.product)) {
+
+@main(s"Confirmation | Subscribe to the ${subscription.firstPlan.title} | The Guardian", bodyClasses=List("is-wide"), touchpointBackendResolutionOpt = Some(touchpointBackendResolution), product = Some(subscription.firstProduct)) {
 
 <script type="text/javascript">
     guardian.pageInfo.slug="GuardianDigiPack:Order Complete";
     guardian.pageInfo.productData = {
-        numberOfMonths: @{plan.charges.billingPeriod.monthsInPeriod},
+        numberOfMonths: @{subscription.firstPlan.charges.billingPeriod.monthsInPeriod},
```
pvighi: I'm not sure what this is used for but numberOfMonth would not make sense for 6 for 6

paulbrown1982: I think it was a requirement for Omniture. Google Analytics does not use it, so it can go. Also delete line 22 in assets/scripts/modules/checkout/eventTracking.js.

app/model/SubscriptionOps.scala
```diff
@@ -3,22 +3,36 @@ package model
...
 
   type WeeklyPlanOneOff = PaidSubscriptionPlan[Product.Weekly, PaidCharge[Weekly.type, OneOffPeriod]]
 
   implicit class EnrichedPaidSubscription[P <: Paid](subscription: Subscription[P]) {
 
+    def isHomeDelivery: Boolean = containsPlanFor(Delivery)
+
+    def isVoucher: Boolean = containsPlanFor(Voucher)
+
+    def isDigitalPack: Boolean = containsPlanFor(com.gu.memsub.Product.Digipack)
```
johnduffell: you could import Product and just write `Product.Digipack`

app/model/SubscriptionOps.scala
```diff
@@ -3,22 +3,36 @@ package model
...
+
+    def isGuardianWeekly = subscription.plans.list.exists(plan => cond(plan.product) { case _: Product.Weekly => true })
+
     private def containsPlanFor(p:Product):Boolean = subscription.plans.list.exists(_.product == p)
 
+    def hasDigitalPack: Boolean = subscription.plans.list.exists(_.charges.benefits.list.contains(Digipack))
```
johnduffell: bear in mind plans includes expired ones as well, so this is more like `hasOrHadDigitalPack`

app/controllers/Checkout.scala
```diff
@@ -283,22 +283,18 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
 
     val sessionInfo = for {
       subsName <- session.get(SubsName)
-      plan <- session.get(RatePlanId).map(ProductRatePlanId).flatMap(p => catalog.allSubs.flatten.find(_.id == p))
-      ratePlanId <- session.get(RatePlanId)
-      currencyStr <- session.get(SessionKeys.Currency)
-      currency <- Currency.fromString(currencyStr)
       startDate <- session.get(StartDate)
-    } yield (subsName, plan, currency, startDate)
+    } yield (subsName, startDate)
```
paulbrown1982: The 'startDate' being in the session helps genericise this page as it gets round the fact that subscriptions of different products base their fulfilment and payment off different date triggers.

app/views/account/fragments/planList.scala.html
```diff
@@ -0,0 +1,34 @@
...
+@(plans: List[SubscriptionPlan.ContentSubscription], singularLabel: String, chosenPaperDays: List[PaperDay] = List.empty)
+
+@if(plans.length > 1) {
+    <dt class="mma-section__list--title">@{
+        singularLabel + "s"
+    }</dt>
+}
+@if(plans.length == 1) {
+    <dt class="mma-section__list--title">@singularLabel</dt>
+}
```
johnduffell: the dt part could be common, and this could be something like `@if(plans.length == 1) (singularLabel) else (singularLabel + "s")`

paulbrown1982:
👍 nice solution.

app/views/account/fragments/planList.scala.html
pvighi: I though about that but I don't want to output anything in case the plan list is empty.  I could do that and put everything inside a condition that checks that there are plans in the list

app/views/checkout/thankyou.scala.html
```diff
@@ -1,34 +1,32 @@
...
+@import model.SubscriptionOps._
 @import views.support.PlanOps._
-@import com.gu.memsub.subsv2.CatalogPlan
-@import com.gu.memsub.Product
 
-@(subscriptionName: String,
+@(  subscription: Subscription[ContentSubscription],
     guestAccountForm: Option[Form[model.GuestAccountData]] = None,
     touchpointBackendResolution: services.TouchpointBackend.Resolution,
-    plan: CatalogPlan.Paid, // we should take a subscription here
```
johnduffell: 🎉

app/views/support/PlanOps.scala
```diff
@@ -93,16 +81,24 @@ object PlanOps {
...
+        case _: Product.Weekly => "https://www.theguardian.com/help/2012/jan/19/guardian-weekly-faqs"
         case _ => "https://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions"
       }
 
+    def productType: String = product match {
+      case Delivery => "Home Delivery"
+      case Voucher => "Voucher Book"
+      case Product.Digipack => "Digital Pack"
+      case _: Product.Weekly => "Guardian Weekly"
+      case _ => "Unknown"
```
johnduffell: product.toString ?

johnduffell:
that's great! good work with all the refactoring 👍

app/views/support/PlanOps.scala
pvighi: I just moved this code from plan to here as it was..
I don't know if it is safe to change that, I'll have to check the js code that uses this.

app/views/account/fragments/planList.scala.html
johnduffell: if it's empty then it's plural right? 😆 but yeah it's only a small saving either way

PR #859, opened 2017-03-07
Fix one-off renewal with promo code + ordering of email fields

jacobwinch:
@paulbrown1982 @pvighi 

- Bump membership common to pull in: https://github.com/guardian/membership-common/pull/447
- Remove text from renewal thankyou page - this is incorrect if you renew with a promo code (I also think it's superfluous as we show the full billing schedule anyway).
- Correct the order of fields for GuardianWeeklyRenewalDataExtensionRow

### DIFF
--- a/app/views/account/thankYouRenew.scala.html
+++ b/app/views/account/thankYouRenew.scala.html
@@ -13,10 +13,7 @@
     <main class="page-container gs-container gs-container--slim">
         @fragments.page.header("Thank you", None, List("l-padded"))
         <section class="section-slice section-slice--bleed section-slice--limited">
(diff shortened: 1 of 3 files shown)


app/views/account/thankYouRenew.scala.html
```diff
@@ -13,10 +13,7 @@
     <main class="page-container gs-container gs-container--slim">
         @fragments.page.header("Thank you", None, List("l-padded"))
         <section class="section-slice section-slice--bleed section-slice--limited">
-            <p>Your Guardian Weekly subscription has been renewed. @plan.charges.billingPeriod match {
```
paulbrown1982: We could reinstate this at some point in the future.

PR #860, opened 2017-03-07
Renewal promo step up text

paulbrown1982:
Hopefully improve the text which appears in checkout, renewal and confirmation emails that describes the revised billing schedule when a promotion applies to a sub, by making it slightly more clear that the rate paid for after the promotional period has ended is at the standard rate, which the Ts&Cs hopefully say may be subject to any price rises after a sub's initial 1 year term.

Before:
picture 41

After:

picture 42

cc @jacobwinch @pvighi @johnduffell @jayceb1 @AWare

### DIFF
--- a/app/views/support/Pricing.scala
+++ b/app/views/support/Pricing.scala
@@ -36,14 +36,18 @@ object Pricing {
         plan.billingPeriod match {
           case Month =>
             val span = durationMonths
-            if (span > 1) {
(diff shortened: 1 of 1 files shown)


johnduffell:
great that's nice and clear ": 👍 🚀 
just the comment about a lot of duplication

johnduffell:
very clear! 👍 🐎

PR #861, opened 2017-03-07
Bugfix thankyou

pvighi:
prevent exception on thank you page with non gbp transactions

### DIFF
--- a/app/views/fragments/checkout/reviewPanel.scala.html
+++ b/app/views/fragments/checkout/reviewPanel.scala.html
@@ -15,7 +15,7 @@
 <div class="review-panel">
 
     <input type="hidden" name="subscriptionDetails"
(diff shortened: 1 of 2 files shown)


app/views/fragments/checkout/reviewPanel.scala.html
```diff
@@ -15,7 +15,7 @@
 <div class="review-panel">
 
     <input type="hidden" name="subscriptionDetails"
-           data-amount="@subscription.firstPlan.charges.gbpPrice.amount"
+           data-amount="0"
```
pvighi: @paulbrown1982  gbpprice was used here as well, is it ok to set it to 0 here also?

johnduffell:
👍

PR #867, opened 2017-03-08
tolerate empty billing schedules when renewing a sub

pvighi:
Asking for the billing schedule of a sub with no current plan seems to make Zuora return an empty response. This change allows the customer to see the thank you page after renewing a subscription even if there is no billing period.
When we have more time we should probably change the definition of BillingPeriod to allow empty response and remove the options on all the places where we handle Option[BillingPeriod].

As far as I know this only happens when renewing a sub that has no current plan but has a term end date in the future

@johnduffell @jacobwinch @paulbrown1982


johnduffell:
👍

PR #868, opened 2017-03-08
allow donor renewal

johnduffell:
donors couldn't renew before, this should handle it.

In the long term we need to make it only use the email in zuora billto, then it will handle gifts properly

@paulbrown1982 @pvighi @jacobwinch

### DIFF
--- a/app/views/account/weeklyRenew.scala.html
+++ b/app/views/account/weeklyRenew.scala.html
@@ -13,6 +13,7 @@
 @import com.gu.i18n.Currency
 @import com.gu.memsub.promo.PromoCode
 @import org.joda.time.LocalDate.now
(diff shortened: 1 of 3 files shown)


app/views/account/weeklyRenew.scala.html
```diff
@@ -57,7 +58,7 @@ <h3 class="mma-section__header">Renew your subscription</h3>
                 </p>
 
                 <div class="js-weekly-renew"
-                data-email="@contact.email"
+                data-email="@if(subscription.readerType == ReaderType.Direct) {@contact.email} else {}"
```
paulbrown1982: Need to remember to remove this once you get the email from Zuora's Bill To instead of the Salesforce contact.

PR #873, opened 2017-03-10
Revert "Temporarily hard-code the first available paper date for expired renewals"

jacobwinch:
Reverts guardian/subscriptions-frontend#871

Merge this on Monday.


jacobwinch:
@paulbrown1982 @johnduffell @pvighi @AWare 

We don't need this to be hardcoded any more - 24th March really is the next available paper date.

johnduffell:
👍 confusing there being in 2 places, client side and server

PR #874, opened 2017-03-10
Correct contract effective date for 6for6

paulbrown1982:
Use new membership-common API to alter the start date of the 6-for-6 rate plan charge in a different way, such that the contract effective date can be set to 'now' and be consistent with all other subs. Also the + - one day (aka 'starting the six for six on a Thursday') is no longer necessary as the fulfilment scripts can now dedupe.

Requires: https://github.com/guardian/membership-common/pull/450

https://trello.com/c/RETXPNdh/169-bug-six-weeks-rate-plans-use-the-contract-effective-date-incorrectly

cc @jacobwinch @johnduffell @pvighi @AWare @svillafe

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -68,7 +68,7 @@ object Config {
   object Zuora {
     private val stageConfig = config.getConfig("touchpoint.backend.environments").getConfig(stage)
 
-    val paymentDelay = Days.days(stageConfig.getInt("zuora.paymentDelayInDays"))
+    val defaultDigitalPackFreeTrialPeriod = Days.days(stageConfig.getInt("zuora.paymentDelayInDays"))
   }
 
   val subscriptionsUrl = config.getString("subscriptions.url")
--- a/app/model/exactTarget/DataExtensionRow.scala
+++ b/app/model/exactTarget/DataExtensionRow.scala
@@ -326,8 +326,7 @@ object GuardianWeeklyRenewalDataExtensionRow {
             contact: Contact,
             paymentMethod: PaymentMethod,
             email: String,
-            customerAcceptance: LocalDate,
-            contractEffective: LocalDate
+            newTermStartDate: LocalDate
            ): GuardianWeeklyRenewalDataExtensionRow = {
 
 
@@ -339,8 +338,8 @@ object GuardianWeeklyRenewalDataExtensionRow {
       title = contact.title.flatMap(Title.fromString),
       firstName = contact.firstName.getOrElse(""),
       lastName = contact.lastName,
-      startDate = contractEffective,
-      firstPaymentDate = customerAcceptance,
+      startDate = newTermStartDate,
+      firstPaymentDate = newTermStartDate,
       planName = planName,
       subscriptionName = subscriptionName,
       paymentMethod = paymentMethod,
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -310,20 +310,12 @@ class CheckoutService(identityService: IdentityService,
       }
     }
 
-<<<<<<< HEAD
     val currentVersionExpired = subscription.termEndDate.isBefore(now).withContextLogging("currentVersionExpired")
-    // For a renewal, all dates should be identical. If the sub has expired, this date should be fast-forwarded to the next available paper date
-    val startDateForRenewal = {
-      if (currentVersionExpired) CheckoutService.determineFirstAvailablePaperDate(now) else subscription.termEndDate
-    }.withContextLogging("startDateForRenewal")
-    val contractEffective = startDateForRenewal
-    val customerAcceptance = startDateForRenewal
-=======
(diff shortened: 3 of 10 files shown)


app/model/exactTarget/DataExtensionRow.scala
```diff
@@ -339,8 +338,8 @@ object GuardianWeeklyRenewalDataExtensionRow {
       title = contact.title.flatMap(Title.fromString),
       firstName = contact.firstName.getOrElse(""),
       lastName = contact.lastName,
-      startDate = contractEffective,
-      firstPaymentDate = customerAcceptance,
+      startDate = newTermStartDate,
```
paulbrown1982: The fulfilment (re)start date and renewal first payment date are always the same date for renewals.

app/configuration/Config.scala
```diff
@@ -68,7 +68,7 @@ object Config {
   object Zuora {
     private val stageConfig = config.getConfig("touchpoint.backend.environments").getConfig(stage)
 
-    val paymentDelay = Days.days(stageConfig.getInt("zuora.paymentDelayInDays"))
+    val defaultDigitalPackFreeTrialPeriod = Days.days(stageConfig.getInt("zuora.paymentDelayInDays"))
```
paulbrown1982: I didn't like the name collision of the Digital Pack free trial period length and the payment delay when orchestrating a subscription.

app/services/CheckoutService.scala
```diff
@@ -71,21 +72,20 @@ class CheckoutService(identityService: IdentityService,
     case _ => false
   }
 
-  def processIntroductoryPeriod(defaultPaymentDelay: Days, originalCommand: Subscribe): Subscribe = {
+  def processIntroductoryPeriod(fulfilmentDelay: Days, originalCommand: Subscribe): Subscribe = {
+    val introductoryPeriodDelayDays = Math.max(fulfilmentDelay.getDays, 0)
+    val sixWeeksPlanStartDate = originalCommand.contractEffective.plusDays(introductoryPeriodDelayDays)
```
pvighi: this is the start of the quarterly plan not the six weeks one right?

app/services/CheckoutService.scala
```diff
@@ -71,21 +72,26 @@ class CheckoutService(identityService: IdentityService,
...
-    }).map(plan => RatePlan(plan.id.get, None))
-
-    val updatedCommand = additionalRateplan.map { ratePlantoAdd =>
-      val updatedRatePlans = ratePlantoAdd <:: originalCommand.ratePlans
-
-      val defaultDelayDays = defaultPaymentDelay.getDays
-      val introductoryPeriodDelayDays = defaultDelayDays - 1
+  private val allSixWeekAssociations = List(catalog.weekly.zoneA.associations, catalog.weekly.zoneC.associations).flatten
+
+  // This method is not genericised because the '6' is not stored in the association.
```
johnduffell: so the association will be renamed to linked or soldTogether or something, and then possibly the 6 added

app/services/CheckoutService.scala
```diff
@@ -71,21 +72,26 @@ class CheckoutService(identityService: IdentityService,
...
+  def processSixWeekIntroductoryPeriod(fulfilmentDelay: Days, originalCommand: Subscribe): Subscribe = {
+    val introductoryPeriodDelayDays = Math.max(fulfilmentDelay.getDays, 0)
+    val sixWeeksPlanStartDate = originalCommand.contractEffective.plusDays(introductoryPeriodDelayDays)
+    val maybeSixWeekAssociation = allSixWeekAssociations.find(_._1.id == originalCommand.ratePlans.head.productRatePlanId)
```
johnduffell: also we will make the associations a case class so we don't need to break apart a tuple

app/services/CheckoutService.scala
```diff
@@ -71,21 +72,26 @@ class CheckoutService(identityService: IdentityService,
...
-    val updatedCommand = additionalRateplan.map { ratePlantoAdd =>
-      val updatedRatePlans = ratePlantoAdd <:: originalCommand.ratePlans
-
-      val defaultDelayDays = defaultPaymentDelay.getDays
-      val introductoryPeriodDelayDays = defaultDelayDays - 1
+  private val allSixWeekAssociations = List(catalog.weekly.zoneA.associations, catalog.weekly.zoneC.associations).flatten
+
+  // This method is not genericised because the '6' is not stored in the association.
+  def processSixWeekIntroductoryPeriod(fulfilmentDelay: Days, originalCommand: Subscribe): Subscribe = {
+    val introductoryPeriodDelayDays = Math.max(fulfilmentDelay.getDays, 0)
```
johnduffell: why would fulfilmentDelay.getDays be negative?

johnduffell: introductoryPeriodDelayDays should be daysUntilFirstIssue?

johnduffell:
👍 looks good so as long as we test it enough!! deinittely need to tidy up the catalog `associations` and make it less generic, then we can make the code more sensible

PR #881, opened 2017-03-14
Validate all promos against UK delivery country, pending full fix

jacobwinch:
@paulbrown1982 

This should cover us whilst we are working on the correct fix.


jacobwinch:
https://trello.com/c/JHPk7QK7/193-fix-renewal-with-promo

PR #885, opened 2017-03-16
hardwires the earliest fulfilment date for weekly to the 24th

AWare:
hardwires the earliest fulfilment date for weekly to the 24th 
(and renames the method to calculate it in the checkoutservice)

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -45,7 +45,8 @@ object CheckoutService {
     p => Days.daysBetween(now, p.startDate), d => zuora.gracePeriodInDays.plus(zuora.defaultDigitalPackFreeTrialPeriod)
   )
 
-  def determineFirstAvailablePaperDate(now: LocalDate): LocalDate = {
+  def determineFirstAvailableWeeklyDate(now: LocalDate): LocalDate = {
+    val initialFulfilment = new LocalDate("2017-03-24")
     val dayOfWeek = now.getDayOfWeek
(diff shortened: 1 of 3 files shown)


johnduffell:
but why?

AWare:
https://trello.com/c/lA03nR4C/197-guardian-weekly-hard-code-selected-date-to-friday-24th-till-fulfilment-report-has-gone-out

johnduffell:
actually it doesn't hard code it, it puts a limit on how early it can be

johnduffell:
sounds good, although I feel like we should just change the fulfilment delay to be 8 days instead of 7 😆 👍

AWare:
@jacobwinch

jacobwinch:
However, we need some kind of reminder to remove the references to the 24th and to un-comment the tests.

jacobwinch:
(Since we don't want to revert this pull request, as then we'll lose your refactoring)

app/services/CheckoutService.scala
```diff
@@ -44,16 +44,14 @@ object CheckoutService {
   def paymentDelay(in: Either[PaperData, DigipackData], zuora: ZuoraProperties)(implicit now: LocalDate): Days = in.fold(
     p => Days.daysBetween(now, p.startDate), d => zuora.gracePeriodInDays.plus(zuora.defaultDigitalPackFreeTrialPeriod)
   )
+  def nextFriday(d: LocalDate):LocalDate = d match {
```
johnduffell: much clearer logic!!! :1st_place_medal:

PR #888, opened 2017-03-17
write renewal email address to zuora not salesforce

johnduffell:
the description is on https://github.com/guardian/membership-common/pull/453

This is ready to go once we have a version number for the above changes.
@paulbrown1982 @pvighi @jacobwinch @lmath


app/services/CheckoutService.scala
```diff
@@ -361,15 +364,16 @@ class CheckoutService(identityService: IdentityService,
       }
     }
 
-    def ensureEmail(contact: Contact, subscription: Subscription[WeeklyPlan]) =
-      if (subscription.readerType != ReaderType.Direct || contact.email.isDefined) {
+    def ensureEmail(contact: Queries.Contact, subscription: Subscription[WeeklyPlan]) = {
+      if (contact.email.isDefined) {
```
johnduffell: now we are updating it on the bill to in zuora, this also works for giftors renewing their gifted sub

johnduffell:
This doesn't get updated over to the salesforce contact automatically, however for gifted subs I'm not clear whose email address that would be.

app/services/ExactTargetService.scala
```diff
@@ -164,28 +165,34 @@ class ExactTargetService(
...
+      })
 
     buildDataExtensionRow.flatMap { row =>
-      SqsClient.sendDataExtensionToQueue(Config.holidaySuspensionEmailQueue, row).map {
-        case Success(sendMsgResult) => logger.info(s"Successfully enqueued ${subscription.name.get}'s updated billing schedule email.")
-        case Failure(e) => logger.error(s"Failed to enqueue ${subscription.name.get}'s updated billing schedule email. Details were: " + row.toString, e)
-      }
+      EitherT(SqsClient.sendDataExtensionToQueue(Config.holidaySuspensionEmailQueue, row).map {
+        case Success(sendMsgResult) => \/.right(())
+        case Failure(e) => \/.left(s"Details were: ${row.toString}, ${e.toString}")
```
jacobwinch: Does the row contain PII? If so, we should not log it.

PR #893, opened 2017-03-22
Allow six for six promotions to be entered organically (a fix for six)

AWare:
- [x] Allow users to select any plan for the product they are buying.
- [x] Hide promotional only plans by default.
- [x] Show promotional plans when they are the default plan (via URL path) for the checkout page.
- [x] Show promotional plans when the user types in a promotion or it's included in the URL query.

Eventually we'll migrate 6-for-6 promotions to not be Tracking type. See https://github.com/guardian/membership-common/pull/455

cc @jacobwinch @pvighi

### DIFF
--- a/app/controllers/Promotion.scala
+++ b/app/controllers/Promotion.scala
@@ -35,7 +35,7 @@ object Promotion extends Controller with LazyLogging with CatalogProvider {
         .filter(plan => promo.appliesTo.productRatePlanIds contains plan.id)
         .filter(plan => plan.charges.currencies contains currency)
         .map(plan => RatePlanPrice(plan.id, plan.charges)).map { ratePlanPrice =>
-          ratePlanPrice.ratePlanId.get -> ratePlanPrice.chargeList.prettyPricingForDiscountedPeriod(discountPromo, currency) //TODO: make this return adjusted rateplans if we're not changing them
+          ratePlanPrice.ratePlanId.get -> ratePlanPrice.chargeList.prettyPricingForDiscountedPeriod(discountPromo, currency)
       }.toMap
     }
   }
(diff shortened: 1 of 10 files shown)


app/controllers/Promotion.scala
```diff
@@ -35,7 +35,7 @@ object Promotion extends Controller with LazyLogging with CatalogProvider {
         .filter(plan => promo.appliesTo.productRatePlanIds contains plan.id)
         .filter(plan => plan.charges.currencies contains currency)
         .map(plan => RatePlanPrice(plan.id, plan.charges)).map { ratePlanPrice =>
-          ratePlanPrice.ratePlanId.get -> ratePlanPrice.chargeList.prettyPricingForDiscountedPeriod(discountPromo, currency)
+          ratePlanPrice.ratePlanId.get -> ratePlanPrice.chargeList.prettyPricingForDiscountedPeriod(discountPromo, currency) //TODO: make this return adjusted rateplans if we're not changing them
```
paulbrown1982: I would advise against this, there will be no way to know which rate plans you apply to in the tool.

AWare: and appliesTo is already in the response.

AWare:
It's basically ready to go now, but what I'm going to attempt in the next commit:

- [ ] move promotionalPlan metadata from label to input to match the rest of the plan metadata
- [ ] refactor hide/show logic into ratePlanChoice
- [ ] hide non promotional plans when missing from appliesTo
- [ ] choose a plan which is displayed when the selected plan is hidden

AWare:
Ok, I'm pretty happy with this. Probably just needs a little more testing before merge.

PR #900, opened 2017-04-03
Use Play's default execution context

lmath:
Change to use Play's defaultContext as the execution context. 

See: https://stackoverflow.com/questions/30805337/plays-internal-execution-context


jacobwinch:
Looks good to me! 👍

PR #901, opened 2017-04-04
Use TiP to verify first production sale after deployment

jacobwinch:
This PR enables us to report on the first production sale after a deploy, using:
https://github.com/guardian/tip

Note that this requires us to add a GitHub access token to our private .conf file. For the purposes of testing out TiP, I suggest that we use my personal access token for this (temporarily).

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -266,7 +266,6 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
       }
 
       logger.info(s"User successfully became subscriber:\n\tUser: SF=${r.salesforceMember.salesforceContactId}\n\tPlan: ${subscribeRequest.productData.fold(_.plan, _.plan).name}\n\tSubscription: ${r.subscribeResult.subscriptionName}")
-      if (tpBackend.environmentName == "PROD") Tip.verify()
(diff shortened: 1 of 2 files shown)


app/controllers/Checkout.scala
```diff
@@ -309,6 +308,7 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
       val eventualMaybeSubscription = tpBackend.subscriptionService.get[com.gu.memsub.subsv2.SubscriptionPlan.ContentSubscription](Name(subsName))
       eventualMaybeSubscription.map { maybeSub =>
         maybeSub.map { sub =>
+          if (tpBackend.environmentName == "PROD") Tip.verify()
```
jacobwinch: @pvighi suggested that I moved the verification step here, as this means that we will also catch a situation where we have broken the logic which displays the thankyou page

johnduffell: good, that means it's like the server side version of the GA "conversion" target

johnduffell:
so you're adding the configuration as a separate operation in aws directly, I assume!
👍 🎉 🌟 🏇

johnduffell:
no tip notification? 😕

johnduffell:
🎉

PR #902, opened 2017-04-04
Write identityId into Zuora

jacobwinch:
* Upgrade to: https://github.com/guardian/membership-common/pull/458
* If the user has an identity account, write the identityId onto the Zuora account object

### DIFF
--- a/app/services/PaymentService.scala
+++ b/app/services/PaymentService.scala
@@ -18,8 +18,9 @@ class PaymentService(val stripeService: StripeService) {
     def makePaymentMethod: Future[PaymentMethod]
   }
 
-  case class DirectDebitPayment(paymentData: DirectDebitData, firstName: String, lastName: String, memberId: ContactId) extends Payment {
-    override def makeAccount = Account.goCardless(memberId, GBP, autopay = true)
+  case class DirectDebitPayment(paymentData: DirectDebitData, firstName: String, lastName: String, purchaserIds: PurchaserIdentifiers) extends Payment {
+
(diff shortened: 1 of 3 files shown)


app/services/PaymentService.scala
```diff
@@ -48,11 +49,18 @@ class PaymentService(val stripeService: StripeService) {
     }
   }
 
+  def identityIdForAccount(purchaserIds: PurchaserIdentifiers) = {
+    purchaserIds.identityId match {
+      case Some(idUser) => idUser.id
+      case None => ""
```
johnduffell: is "" really an identity id?

johnduffell: I prefer to keep the optionality in the type system if possible, rather than flattening none into an empty string so early

johnduffell: ok paul mentioned that this is only a temporary state until `purchaserIds.identityId` becomes non optional in a few days

johnduffell:
👍 go for it!
I will merge this into my branch now as it's blocking me a bit, so don't go rebasing or squashing or anything!! otherwise I will 😭 cry

jacobwinch:
Thanks @johnduffell , I'm going to merge this and tackle the re-factor to render identityId non-optional in a separate PR: [Trello card](https://trello.com/c/krRcStO5/82-make-identityid-non-optional-in-processsubscription)

johnduffell:
sounds good to me, thanks jacob! 👍

johnduffell:
and thanks for not rebasing 😛

PR #903, opened 2017-04-04
Do not allow managing cancelled subs

pvighi:
Logging in to /manage with a cancelled home delivery sub causes a 500 error.  [trello card](https://trello.com/c/KFnLL0AM/50-improve-handling-of-cancelled-subs-who-try-to-access-manage)

@jacobwinch  mentioned that there are similar problems with other types of subs, for example attempting to renew a cancelled subscription.

This change filters out cancelled subs so the user would get an "Unable to verify your details" error when trying to log in.

@johnduffell @jacobwinch @paulbrown1982

### DIFF
--- a/app/controllers/AccountManagement.scala
+++ b/app/controllers/AccountManagement.scala
@@ -27,7 +27,7 @@ import play.api.mvc._
 import utils.TestUsers.PreSigninTestCookie
 import views.html.account.thankYouRenew
 import views.support.Pricing._
-
+import views.support.Dates._
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import scala.concurrent.Future
 import scalaz.std.scalaFuture._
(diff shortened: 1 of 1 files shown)


johnduffell:
If the details are correct and we actually know it's cancelled, we should be more helpful.. it's not like we need to be cagey about the reason if the details are accurate

johnduffell:
however this is an improvement on a 500 error and a quick fix, so I'll leave it to future improvement or up to you for the message 👍

paulbrown1982:
👍 too

pvighi:
I added a custom error message for cancelled subs in /manage. I'm open to suggestions on the exact wording, date format or anything else
canc

pvighi:
changed date format
canc

johnduffell:
👍 :1st_place_medal:

PR #906, opened 2017-04-10
don't need to hard code weekly date any more

johnduffell:
This is breaking the tests as of today (since friday finished) so it needs fixing.  Since we're not manually fulfilling, it's not needed any more anyway.

@paulbrown1982 @pvighi @jacobwinch


jacobwinch:
We should also remove any client-side references to a hard-coded date, but let's get this merged first as it's blocking master 👍

PR #907, opened 2017-04-11
use the suspension service call to get the unfinished suspensions

johnduffell:
If people have broken holidays in the past, we don't want to deal with them in case they are broken in some way (finish before they start) This filters them before they can cause an issue.

@paulbrown1982 @jacobwinch @pvighi @lmath

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.390-SNAPSHOT",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -43,7 +43,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.389",
+    "com.gu" %% "membership-common" % "0.390-SNAPSHOT",
```
johnduffell: https://github.com/guardian/membership-common/pull/459

jacobwinch:
👍 (pending update to released membership-common)

PR #909, opened 2017-04-13
Manually fulfilling today due to Easter bank holiday

paulbrown1982:
Manually fulfilling on Thursday this week due to Monday being Easter Bank Holiday. I've updated tests accordingly to cover this week, and then the normal algorithm beyond.

Before:
picture 70

After:
picture 69

cc @jacobwinch 

Also (fulfilling is the only word where it has two Ls): http://dictionary.cambridge.org/dictionary/english/fulfilling 😛


jacobwinch:
👍 - bonus points for future-proofing the tests

PR #910, opened 2017-04-13
Fix holidays over term end bug.

AWare:
I've added a bunch of logging to the suspend calls, and am also attempting to renew the sub if the holiday falls over a term boundary. Also bumps common, which changes the rest endpoint.


AWare:
There's a report of someone that had a renewal run early and not getting fulfilled. Blocking this PR until we figure this out.

AWare:
Still waiting on Salesforce :(

AWare:
This works 👌🏻 with fulfilment, unblocking. 
@jacobwinch

jacobwinch:
Awesome 👍

AWare:
Tested and verified in prod 👍 

This was causing fastly to timeout, but we've extended the time to first byte.

jacobwinch:
Nice work - the Fastly timeout is an interesting catch... I wonder if it could be partly responsible for the frequency of:
https://sentry.io/the-guardian/subscriptions/issues/209062388/

PR #911, opened 2017-04-13
Add appnexus pixel

paulbrown1982:
Add the AppNexus tracking pixel with the various IDs representing the stage of the conversion funnel. Note, the Acquisitions team have not set up IDs for the Guardian Weekly flow.

Also fixed the Facebook tracker which was erroring in the console.

picture 71

cc @AWare @jacobwinch @pvighi @lmath

### DIFF
--- a/assets/javascripts/modules/analytics/appnexus.js
+++ b/assets/javascripts/modules/analytics/appnexus.js
@@ -1,0 +1,45 @@
+define([
+    'modules/analytics/analyticsEnabled',
+    'modules/analytics/dntEnabled'
+], function (analyticsEnabled, dntEnabled) {
+    'use strict';
+
+    var segmentedPageCodes = {
+        'Digital': {
+            'Landing': '8326260',
+            'Checkout': '8326264',
+            'Thankyou': '836984'
+        },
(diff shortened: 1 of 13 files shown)


assets/javascripts/modules/analytics/appnexus.js
```diff
@@ -0,0 +1,45 @@
...
+        if (!(guardian && guardian.pageInfo && guardian.pageInfo.productData && guardian.pageInfo.productData.productSegment)) { return; }
+
+        var pageCodes = segmentedPageCodes[guardian.pageInfo.productData.productSegment];
+        if (!pageCodes) { return; }
+
+        var pageType = guardian.pageInfo.pageType;
+        if (!pageType) { return; }
+
+        var oImg = document.createElement('img');
+        oImg.setAttribute('src', 'https://secure.adnxs.com/seg?t=2&add=' + pageCodes[pageType]);
```
paulbrown1982: The t=2 refers to how the pixel was exported from the platform. There are two options, for it to be exported as Javascript or an image. The first option adds a t=1, whereas the second a t=2.

PR #913, opened 2017-04-18
Yaml cloudformation and CODE config

pvighi:
Translated the cloudformation template to YAML and made some changes to support a CODE env
@paulbrown1982


conf/CODE.conf
```diff
@@ -0,0 +1,36 @@
+include "touchpoint.DEV.conf"
+include "touchpoint.UAT.conf"
+include "application"
+
+stage = "DEV"
+
+identity {
+  baseUri = "https://idapi.code.dev-theguardian.com"
+  production.keys=false
+  webapp.url="https://profile-origin.thegulocal.com"
```
pvighi: this basic conf works for some basic tests but some things need to be configured properly still, like all these references to thegulocal.com

PR #914, opened 2017-04-18
Changed config in preparation for switch to systemd and deb packages

pvighi:
[equivalent change in cas proxy](https://github.com/guardian/content-authorisation-proxy/pull/76)
@paulbrown1982 @jacobwinch

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -50,4 +50,4 @@ get-address-io-api {
     key = "q0M6MVJQm0CslqDvMcKUDQ6396"
 }
 
-include file("/etc/gu/subscriptions-frontend.conf")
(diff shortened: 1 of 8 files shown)


conf/application.conf
```diff
@@ -50,4 +50,4 @@ get-address-io-api {
     key = "q0M6MVJQm0CslqDvMcKUDQ6396"
 }
 
-include file("/etc/gu/subscriptions-frontend.conf")
+include file("/etc/gu/subscriptions-frontend.private.conf")
```
paulbrown1982: You should also update the aws command in README.md to reflect the download of the new file location.

PR #919, opened 2017-04-24
Call toUpperCase on postcodes.

AWare:
This should fix our postcode data quality issues, but we might need to keep an eye on this.


jacobwinch:
Looks fine to me (once tests are fixed) 👍

PR #921, opened 2017-04-28
Manually fulfil this week

AWare:
gif by Daniela Sherer


jacobwinch:
I ran through the flow locally and can also see users signing up in PROD - need to investigate what went wrong with the tests, but I'm fairly confident that it's unrelated to this change.

PR #927, opened 2017-05-05
Use px not seg for appnexus conversion code

paulbrown1982:
Fixed mistake where I did not notice that the appnexus pixel should have different path and parameter names when conversion vs segment type.

cc @jacobwinch @pvighi @lmath @johnduffell @AWare


johnduffell:
how did you know it should be done differently? I can't really check this against the docs

Other than that I'm sure it's fine though! 👍

PR #928, opened 2017-05-08
Fixed bug causing appnexus conversion pixel not to show

paulbrown1982:
Fixed bug where I was inspecting the wrong variable to determine whether to show the conversion or segmentation pixel. My testing on Friday picked up a false-positive when it only tested that the segmentation pixel showed correctly.

cc @jacobwinch @johnduffell @AWare @pvighi @lmath


johnduffell:
👍

PR #929, opened 2017-05-09
allow manage of recurring 6 months weekly

johnduffell:
This follows on from allowing manage six months one off, so people can renew.  This one will allow people who are on recurring to get in, this is because some people are on recurring plans in a non renewing sub for some reason.
@jacobwinch @paulbrown1982 @lmath

### DIFF
--- a/app/views/support/BillingPeriod.scala
+++ b/app/views/support/BillingPeriod.scala
@@ -8,6 +8,7 @@ object BillingPeriod {
       case Month => "every month"
       case Quarter => "every 3 months"
       case Year => "every 12 months"
+      case SixMonthsRecurring => "every six months"
       case period: OneOffPeriod => s"for ${period.noun}"
     }
   }
--- a/build.sbt
+++ b/build.sbt
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.403",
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.403",
+    "com.gu" %% "membership-common" % "0.404",
```
johnduffell: https://github.com/guardian/membership-common/pull/475

app/views/support/BillingPeriod.scala
```diff
@@ -8,6 +8,7 @@ object BillingPeriod {
       case Month => "every month"
       case Quarter => "every 3 months"
       case Year => "every 12 months"
+      case SixMonthsRecurring => "every six months"
```
johnduffell: This would look a little odd if auto renew is turned off as it will show they are on "every six months"
image

jacobwinch: Presumably these people do still pay every 6 months (just via some manual process) so I think it makes sense?

jacobwinch:
👍

app/views/support/BillingPeriod.scala
johnduffell: 😄 hope they will keep paying us!

PR #931, opened 2017-05-10
Show the 'Pay in £GBP' checkbox for all currencies on Zone C

paulbrown1982:
Show the 'Pay in £GBP' checkbox for all currencies on Zone C not just when picking a USD country.

The backend validation basically checks that the currency override is only GBP (as all plans have a GPB price) and it is only provided when in Zone C.

Landing on an Australian checkout:

picture 97
picture 98

And then when changing country to an Euros country:

picture 101
picture 99

See: https://trello.com/c/3ynoNDSG/153-allow-gbp-payment-option-for-all-zone-c-purchases-not-just-usd

cc @pvighi

### DIFF
--- a/app/model/PaymentValidation.scala
+++ b/app/model/PaymentValidation.scala
@@ -18,7 +18,7 @@ object PaymentValidation {
   def validateCurrency(currency: Currency, settings: CountryWithCurrency, plan: CatalogPlan.ContentSubscription): Boolean = {
 
     def isValidCurrencyOverride = plan.product match {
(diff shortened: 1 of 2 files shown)


pvighi:
I checked out the branch and if i try to buy something in pounds for spain a server side validation fails :

[Exception: Backend validation failed: identityId=; JavaScriptEnabled=true; currency invalid plan, currency and country combination]
In   subscriptions-frontend/app/controllers/Checkout.scala:178

PR #932, opened 2017-05-10
load the delivery address from zuora via rest

johnduffell:
Previously we called zuora 3 times to get the sub, account, and billto address, then salesforce to get the delivery address for guardian weekly.
Now I changed it to call zuora twice to get the sub, and account via rest, which automatically pulls in the addresses.
This saves us 2 calls on log in to manage, and means we don't depend on salesforce as much.
@jacobwinch @paulbrown1982 @lmath

### DIFF
--- a/app/controllers/AccountManagement.scala
+++ b/app/controllers/AccountManagement.scala
@@ -166,56 +166,59 @@ object ManageWeekly extends ContextLogging {
 
   }
 
-  def apply(billingSchedule: Option[BillingSchedule], weeklySubscription: Subscription[WeeklyPlan], promoCode: Option[PromoCode])(implicit request: Request[AnyContent], resolution: TouchpointBackend.Resolution): Future[Result] = {
+  def apply(
+    billingSchedule: Option[BillingSchedule],
+    weeklySubscription: Subscription[WeeklyPlan],
+    promoCode: Option[PromoCode]
+  )(implicit
+    request: Request[AnyContent],
+    resolution: TouchpointBackend.Resolution
+  ): Future[Result] = {
     implicit val tpBackend = resolution.backend
     implicit val rest = tpBackend.simpleRestClient
-    implicit val zuoraRestService = new ZuoraRestService[Future]
+    implicit val zuoraRest = new ZuoraRestService[Future]
     implicit val flash = request.flash
     implicit val subContext = weeklySubscription
     if (weeklySubscription.readerType != ReaderType.Agent) {
-      // go to SF to get the contact mailing information etc
-      GetSalesforceContactForSub.zuoraAccountFromSub(weeklySubscription)(tpBackend.zuoraService, tpBackend.salesforceService.repo, defaultContext).flatMap { account =>
-        val futureSfContact = GetSalesforceContactForSub.sfContactForZuoraAccount(account)(tpBackend.zuoraService, tpBackend.salesforceService.repo, defaultContext)
-        val futureZuoraBillToContact = tpBackend.zuoraService.getContact(account.billToId)
-        futureSfContact.flatMap { contact =>
-          futureZuoraBillToContact.map { zuoraContact =>
-            zuoraContact.country.map { billToCountry =>
-              val catalog = tpBackend.catalogService.unsafeCatalog
-              val weeklyPlans = weeklySubscription.planToManage.product match {
-                case Product.WeeklyZoneA => catalog.weekly.zoneA.plans
-                case Product.WeeklyZoneB => catalog.weekly.zoneB.plans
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.404",
+    "com.gu" %% "membership-common" % "0.405-SNAPSHOT",
```
johnduffell: https://github.com/guardian/membership-common/pull/476

app/controllers/AccountManagement.scala
```diff
@@ -166,56 +166,59 @@ object ManageWeekly extends ContextLogging {
...
+              Ok(weeklySubscription.asRenewable.map { renewableSub =>
+                info(s"sub is renewable - showing weeklyRenew page")
+                views.html.account.weeklyRenew(renewableSub, account.soldToContact, account.billToContact.email, billToCountry, weeklyPlanInfoList, existingCurrency, promoCode)
+              } getOrElse {
+                info(s"sub is not renewable - showing weeklyDetails page")
+                views.html.account.weeklyDetails(weeklySubscription, billingSchedule, account.soldToContact)
+              })
           }
+        }.getOrElse {
+          info(s"no valid bill to country for account")
```
jacobwinch: This should probably be an error - I think we'd like to know about it, because it's an invalid state?

app/controllers/AccountManagement.scala
```diff
@@ -166,56 +166,59 @@ object ManageWeekly extends ContextLogging {
...
+          Ok(views.html.account.details(None, promoCode))
         }
-      }
+      }.run.map(_.fold({ error =>
+        info(s"problem getting account: $error")
```
jacobwinch: This could be error too?

jacobwinch:
I've made a couple of small comments, but overall this makes me very happy 👍

app/controllers/AccountManagement.scala
johnduffell: yes I was wondering that too (it's existing code) and when I tried it, it just dumps you back on the page with no comment.  We seem to use some kind of "flash" json thing built in to play to show the "unable to ...." message.  But I was thinking we need to revamp the whole thing to produce some kind of meaningful message (or error code) in every situation.  Then when people call in, even if TP don't have a list of codes, we would have additional information.

jacobwinch: Yes agreed, there are a lot of things we could do here to improve error handling. My main thought was that if this is logged at error level (rather than info), it'll at least throw to Sentry, which gives us a better chance of spotting the problem in prod.

johnduffell:
@jacobwinch look at the last commit, although I could have gone a lot further if I'd pulled the refactoring ball of wool!
I decided not to go down the error codes route as we can search the logs based on sub id.
image

app/views/account/fragments/yourDetails.scala.html
```diff
@@ -23,12 +24,14 @@
         <dd class="mma-section__list--content">
             <div>@contact.firstName @contact.lastName</div>
             <div>@{
+                def opt(string: String) = Some(string).filter(_.trim.nonEmpty)
```
paulbrown1982: How about tidying up the display and putting the trim inside the Some? e.g.

`Some(string.trim).filter(_.nonEmpty)`

app/views/account/weeklyRenew.scala.html
```diff
@@ -62,7 +63,7 @@ <h3 class="mma-section__header">Renew your subscription</h3>
                 data-billing-country="@billToCountry.alpha2"
                 data-promo-code="@promoCode.map(_.get)"
                 data-currency="@currency.iso"
-                data-delivery-country="@contact.mailingCountryParsed.map(_.alpha2).getOrElse(billToCountry.alpha2)"
+                data-delivery-country="@contact.country.map(_.alpha2).getOrElse(billToCountry.alpha2)"
```
paulbrown1982: We should consider making the SoldToContact.country non-optional given that it's a required field in our Zuora tennant (because we have ZTax installed): https://knowledgecenter.zuora.com/DC_Developers/G_SOAP_API/E1_SOAP_API_Object_Reference/Contact

johnduffell: it's only optional in the reader in case we don't have it in our lookup/enum.  I think in reality it should be there, but we would end up with a more fail fast situation as the whole sub would be unreadable in that case (as it would either throw an exception or return a Left.)

paulbrown1982: 👌 feel free to create a Trello ticket on the nice-to-have backlog :)

PR #940, opened 2017-05-17
Strip leading zeros from the Subscriber IDs provided by customers in /manage

paulbrown1982:
Strip leading zeros from the Subscriber IDs provided by customers in the /manage page, before looking up in Zuora. This is because a lot of our migrated subscribers have paperwork which contains these leading zeros back when they were managed by a third party, and are getting confused when using this page, and calling up the call centre.

https://trello.com/c/m9kKA4fQ/174-strip-leading-zeros-from-manage-page-sub-id-before-authenticating-with-zuora

cc @johnduffell @AWare @jacobwinch @pvighi @lmath

### DIFF
--- a/app/forms/AccountManagementLoginForm.scala
+++ b/app/forms/AccountManagementLoginForm.scala
@@ -3,7 +3,9 @@ package forms
 import play.api.data.Form
 import play.api.data.Forms._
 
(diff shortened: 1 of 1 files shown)


johnduffell:
thanks! 👍 🚤

PR #944, opened 2017-05-18
Renewal fixes

pvighi:
problems with renewals fixed:
- promotion validation used Zuora Country for the ajax check and salesforce for the server side check
- server side promotion validation didn't cause an error it just didn't apply the promotion and continued with the renewal
- exceptions being thrown in the steps after the zuora amendments were executed resulted in an error telling the customer that the sub was not renewed when it fact it was.

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -346,11 +346,10 @@ class CheckoutService(
       finalRenewCommand.withContextLogging("final renew command")
     }
 
-    def getValidPromotion(contact: Contact): Option[ValidPromotion[com.gu.memsub.promo.Renewal]] = for {
+    def getValidPromotion(addressCountry: Option[Country]): Option[ValidPromotion[com.gu.memsub.promo.Renewal]] = for {
         code <- renewal.promoCode.withContextLogging("promo code from client")
-        deliveryCountryString <- contact.mailingCountry.withContextLogging("salesforce mailing country")
-        deliveryCountry <- Some(Country.UK).withContextLogging("delivery country object")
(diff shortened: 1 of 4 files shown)


app/services/CheckoutService.scala
```diff
@@ -389,32 +390,63 @@ class CheckoutService(
...
 
-    def pricesMatch(displayedPrice: String, subscriptionPrice: String) = {
-      if(displayedPrice != subscriptionPrice){
-        logger.error(s"Client and server side prices divergent, we showed $displayedPrice and wanted to charge $subscriptionPrice from ${subscription.name} (${renewal.promoCode})")
-        false
+    case class ContactInfo(billto: ZuoraContact, soldto: ZuoraContact, salesforceContact: Contact)
+
+    def postRenewalSteps(subscription: Subscription[WeeklyPlanOneOff], renewal: model.Renewal, subscriptionPrice: String, salesforceContact: Contact, newTermStartDate: LocalDate) = {
+      def logException(e: Exception): Unit = {
+        logger.error(s"unexpected error in post renewal steps", e)
```
jacobwinch: Would be nice to use ContextLogging here so we know which subscribers didn't get an email!

jacobwinch:
Ouch, nice catch(es).

Looks like a massive improvement 👍

app/services/CheckoutService.scala
jacobwinch: Could also explicitly state what has failed (i.e. sub has been renewed but email has not been sent) so we know exactly what this means from Sentry without digging into the code.

PR #952, opened 2017-05-26
Make identity check more robust

AWare:
By checking for a 200 response before attempting to parse JSON

### DIFF
--- a/app/services/IdentityService.scala
+++ b/app/services/IdentityService.scala
@@ -26,7 +26,11 @@ class IdentityService(identityApiClient: => IdentityApiClient) extends LazyLoggi
 
   def doesUserExist(email: String): Future[Boolean] =
     identityApiClient.userLookupByEmail(email).map { response =>
-      (response.json \ "user" \ "id").asOpt[String].isDefined
+      response.status match {
+        case status if status == Status.OK =>  (response.json \ "user" \ "id").asOpt[String].isDefined
+        case _ => false
+      }
+
     }
 
   def userLookupByCredentials(accessCredentials: AccessCredentials): Future[Option[IdUser]] = accessCredentials match {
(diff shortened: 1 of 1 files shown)


paulbrown1982:
👍

app/services/IdentityService.scala
```diff
@@ -26,7 +26,11 @@ class IdentityService(identityApiClient: => IdentityApiClient) extends LazyLoggi
 
   def doesUserExist(email: String): Future[Boolean] =
     identityApiClient.userLookupByEmail(email).map { response =>
-      (response.json \ "user" \ "id").asOpt[String].isDefined
+      response.status match {
+        case status if status == Status.OK =>  (response.json \ "user" \ "id").asOpt[String].isDefined
+        case _ => false
```
jacobwinch: Could also add a case which matches on 5xx responses and logs an error before returning false.

app/services/IdentityService.scala
```diff
@@ -26,7 +26,14 @@ class IdentityService(identityApiClient: => IdentityApiClient) extends LazyLoggi
 
   def doesUserExist(email: String): Future[Boolean] =
     identityApiClient.userLookupByEmail(email).map { response =>
-      (response.json \ "user" \ "id").asOpt[String].isDefined
+      response.status match {
+        case status if status == Status.OK =>  (response.json \ "user" \ "id").asOpt[String].isDefined
+        case status => {
+          logger.error(s"ID API failed on email check with HTTP status $status")
```
jacobwinch: Presumably we need to restrict this case to certain codes though (e.g. 404s are presumably not an error, because this just indicates that the user doesn't have Identity yet)

AWare:
I'm not 100% sure that we want to let users register if the email check fails, might want to watch after merge and if this remains a problem make it return an option and show an error.

jacobwinch:
👍

PR #953, opened 2017-05-31
Avoid having to reflect on our ways every time we deploy (use riff-raff.yaml)

pvighi:
before :
ways1
after:
ways2


cloud-formation/subscriptions-app.cf.yaml
```diff
@@ -144,6 +144,9 @@ Resources:
           - Effect: Allow
             Action: s3:GetObject
             Resource: arn:aws:s3:::membership-private/membership_directory_cert.p12
+          - Effect: Allow
+            Action: s3:GetObject
+            Resource: !Sub 'arn:aws:s3:::subscriptions-dist/subscriptions/${Stage}/frontend/*'
```
pvighi: I'm not sure if this is the right way to do this but if you deploy using riff-raff.yaml the file is uploaded to the bucket with read permissions just for aws-membership. Maybe I can change something in the deployment file to change this

jacobwinch:
👍 👍 👍

cloud-formation/subscriptions-app.cf.yaml
paulbrown1982: I think this is the most secure way.

PR #955, opened 2017-06-01
Make identity check more robust 

AWare:
Reverts guardian/subscriptions-frontend#954

Not forgetting that identity returns a 404 for not found.

### DIFF
--- a/app/services/IdentityService.scala
+++ b/app/services/IdentityService.scala
@@ -27,7 +27,7 @@ class IdentityService(identityApiClient: => IdentityApiClient) extends LazyLoggi
   def doesUserExist(email: String): Future[Boolean] =
     identityApiClient.userLookupByEmail(email).map { response =>
       response.status match {
-        case status if status == Status.OK =>  (response.json \ "user" \ "id").asOpt[String].isDefined
+        case status if status == Status.OK =>  true
         case status if status == Status.NOT_FOUND => false
         case status => {
           logger.error(s"ID API failed on email check with HTTP status $status")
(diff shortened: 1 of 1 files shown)


AWare:
We shouldn't even parse the JSON. 
200 -> 1 or more users
404 -> 0 users
other -> errro
https://github.com/guardian/identity/blob/3d607ee78fa92495913d9a63203b9b7be3ad279e/identity-api/src/main/scala/com/gu/identity/api/servlet/UserServlet.scala

app/services/IdentityService.scala
```diff
@@ -26,7 +26,15 @@ class IdentityService(identityApiClient: => IdentityApiClient) extends LazyLoggi
 
   def doesUserExist(email: String): Future[Boolean] =
     identityApiClient.userLookupByEmail(email).map { response =>
-      (response.json \ "user" \ "id").asOpt[String].isDefined
+      response.status match {
+        case status if status == Status.OK =>  true
```
johnduffell: I may be wrong, but I think you can just say `case Status.OK =>` and it'll just check equality.  You only need to name it if you're using it somewhere.

johnduffell:
looks good, only a small comment! 👍 hopefully we didn't miss any response codes - may be worth checking with an identity person just to be sure ( @markjamesbutler or @jfsoul )

jfsoul:
I suppose you might get a 403, but that would be unexpected and the catch-all case seems appropriate for that.

AWare:
@johnduffell thanks :)

johnduffell:
👍 great, looks great, thanks! 🚀

PR #957, opened 2017-06-05
Wrong codes

paulbrown1982:
There was a typo and these codes should have been the other way round. It was decided to switch the codes on the site rather than the products behind the codes in the promo tool.

@AWare @jacobwinch @johnduffell @lmath @pvighi


AWare:
:+1:

PR #961, opened 2017-06-16
Add a new london postcode.

AWare:
This might not be the best way to check whether an address is in London going forward.


jacobwinch:
👍 for the quick fix

However, as you said, it would be good to think about how this list gets generated and kept up to date...

PR #962, opened 2017-06-20
Fix non-display of error messages.

AWare:
Will show error messages to users. Also fixes uknown typo.


AWare:
requires: https://github.com/guardian/membership-common/pull/495

PR #965, opened 2017-06-27
 adds delivery instructions to the account object in zuora instead of the soldtocontact

AWare:
Updates membership common and adds delivery instructions to the account object in zuora instead of the soldtocontact

there's an informal 1:1 rel here so this is okay

https://github.com/guardian/membership-common/pull/500

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -118,8 +118,7 @@ class CheckoutService(
         name = personalData, // TODO once we have gifting change this to the Giftee's name
         address = paperData.deliveryAddress,
         email = personalData.email, // TODO once we have gifting change this to the Giftee's email address
-        phone = personalData.telephoneNumber,
-        deliveryInstructions = paperData.deliveryInstructions
+        phone = personalData.telephoneNumber
       ))
       case _ => None
     }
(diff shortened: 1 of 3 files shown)


app/services/CheckoutService.scala
```diff
@@ -307,17 +306,22 @@ class CheckoutService(
...
           val personalData = subscriptionData.genericData.personalData
           require(personalData.address.country.contains(Country.UK), "Direct Debit payment only works in the UK right now")
-          paymentService.makeDirectDebitPayment(paymentData, personalData.first, personalData.last, purchaserIds)
+          paymentService.makeDirectDebitPayment(paymentData, personalData.first, personalData.last, purchaserIds, deliveryInstructions)
         case paymentData@CreditCardData(_) =>
           val plan = subscriptionData.productData.fold(_.plan, _.plan)
           val desiredCurrency = subscriptionData.genericData.currency
           val currency = if (plan.charges.price.currencies.contains(desiredCurrency)) desiredCurrency else GBP
-          paymentService.makeCreditCardPayment(paymentData, currency, purchaserIds)
+          paymentService.makeCreditCardPayment(paymentData, currency, purchaserIds, deliveryInstructions)
```
johnduffell: if I was starting from scratch, I'd do a makeCreditCardPayment, then call a makeZuoraAccount on the result of that, rather than hiding the creating account in the credit card method which looks strange,
However based on the size of the change at least rename this makeCreditCardPayment to makeZuoraAccountWithCreditCard

app/services/PaymentService.scala
```diff
@@ -18,9 +18,10 @@ class PaymentService(val stripeService: StripeService) {
     def makePaymentMethod: Future[PaymentMethod]
   }
 
-  case class DirectDebitPayment(paymentData: DirectDebitData, firstName: String, lastName: String, purchaserIds: PurchaserIdentifiers) extends Payment {
+  case class DirectDebitPayment(paymentData: DirectDebitData, firstName: String, lastName: String, purchaserIds: PurchaserIdentifiers, deliveryInstructions: Option[String]) extends Payment {
```
johnduffell: I'd rename this to something like ZuoraAccountDirectDebit

app/services/CheckoutService.scala
pvighi: I think the biggest confusion is that Payment is defined here as :
  sealed trait Payment {
    def makeAccount: Account
    def makePaymentMethod: Future[PaymentMethod]
  }

which to me looks more like a payment method/ account builder

AWare: I've renamed it to AccountAndPayment

johnduffell: good idea, although I was meaning the method names primarily

johnduffell:
apart from the rename suggested in the above comment, it's ready! good work on the extra tidyup 👍 💯

PR #966, opened 2017-07-05
Send phone as normalisedphonenumber

AWare:
Previously, the phone number would get stored in Zuora as the string that the user has entered, and in salesforce as 44... and in identity as a case class that's like our NormalisedPhoneNumber. 

This brings salesforce and zuora into line, and they'll now save them as +44...

Depends on https://github.com/guardian/membership-common/pull/502

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -118,7 +118,7 @@ class CheckoutService(
         name = personalData, // TODO once we have gifting change this to the Giftee's name
         address = paperData.deliveryAddress,
         email = personalData.email, // TODO once we have gifting change this to the Giftee's email address
-        phone = NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, personalData.address.country).map(n => s"${n.countryCode}${n.localNumber}")
+        phone = NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, personalData.address.country)
       ))
       case _ => None
     }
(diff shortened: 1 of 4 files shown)


AWare:
Depends on... https://github.com/guardian/membership-common/pull/504

Making the conversion into the right form a job for checkoutservice and not commands.

AWare:
Ready to merge. On leave tomorrow, happy if anyone wants to merge it.

app/services/CheckoutService.scala
```diff
@@ -118,7 +118,7 @@ class CheckoutService(
         name = personalData, // TODO once we have gifting change this to the Giftee's name
         address = paperData.deliveryAddress,
         email = personalData.email, // TODO once we have gifting change this to the Giftee's email address
-        phone = personalData.telephoneNumber
+        phone = NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, personalData.address.country)
```
pvighi: so in the end did we decide to use paperData.deliveryAddress.country or personalData.address.country for the normalised country?

AWare: `paperData.deliveryAddress.country` might be the better choice.

pvighi: I guess one problem could be that some products do not have delivery address

build.sbt
```diff
@@ -44,7 +44,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2,
-    "com.gu" %% "membership-common" % "0.427",
+    "com.gu" %% "membership-common" % "0.429",
```
lmath: Can you please bump this this to 0.431?

AWare: Will do

lmath: Thank you!

PR #972, opened 2017-08-10
First attempt at adding online support for home delivery issues

jacobwinch:
This PR adds a section to /manage, where customers (or agents acting on their behalf) can check on the status of missing newspapers:

image

Submitting the form hits the fulfilment-lookup API (https://github.com/guardian/fulfilment-lookup). 

If the API call succeeds and the sub was included on the file:

image

If the API call succeeds and the sub was not included on the file: 

image

If there is a problem calling the API or if the lambda fails to perform the lookup, we default to directing them towards customer services (maintaining the current process):

image

After discussing with @johnduffell, this first iteration just allows us/CSRs/CEX to know whether we included a sub on the file or not.

In the future, I'd like to have the backend add the details about whether the sub was in the fulfilment file or not to a Salesforce case. This would enable us to clearly track the number of delivery problems which are due to technical fulfilment issues, and the number which are due to issues with the distributor. It should also reduce the time that customers wait to resolve these type of issues (and the number of back and forth negative service contacts which occur as a result of the current process).

Before shipping this, we need to:

- [x] fix the scheduled download of SF fulfilment files
- [x] deploy fulfilment-lookup API to PROD
- [x] add fulfilment-lookup-api-key/fulfilment-lookup-api-url to the config files


app/controllers/AccountManagement.scala
```diff
@@ -132,6 +133,28 @@ object ManageDelivery extends ContextLogging {
     }
   }
 
+  def raiseNonDeliveryCase(deliveryIssue: DeliveryIssue)(implicit request: Request[DeliveryIssue], touchpoint: TouchpointBackend.Resolution): Future[Result] = {
+    implicit val tpBackend = touchpoint.backend
+    logger.info(s"Delivery issue is: $deliveryIssue")
+    val futureResponse = FulfilmentLookupService.lookupSubscription(tpBackend.environmentName, deliveryIssue)
+    futureResponse.map { response =>
+      val responseBody = response.body().string()
+      response.body.close()
```
johnduffell: I'd do all the nuts and bolts returning something nice in the lookupSubscription call, and just return a Future[boolean] or something to say was it successful or not.
Otherwise the fact that it's an http call leaks into the calling code.

johnduffell:
ok I'm really excited that this is on the way 🎉 ,
my only small but important comment is really from a dev process point of view - as discussed I still think we should try to get this out as a minimum viable product and then improve it from there rather than a big bang.

My suggestion would be since we're calling fulfilment lookup in this code, and it should work already (I think?) then we can have a "track delivery" instead of report, and tell people whether it was out for delivery.
That doesn't require any more interactions with sf/zuora/etc, it's just a single function from a high level.

Then I would add the code to let the user raise a case "I expected a paper but it didn't arrive" which would add in the salesforce case raising functionality.  We could remove the track delivery bit at that point if we didn't like it from a UX point of view.

As long as each piece we release doesn't actually harm things (e.g. promising to do something, but it not happening, or implying that the user is at fault, or causing a problem at the call centre) I think it's better to release in smaller self contained pieces and just describing accordingly.

Not sure what other people think, maybe they'll chip in!  And not to disparage the fact that this is excellent work!

jacobwinch:
@johnduffell - thanks, sounds sensible to me. I've re-worked the code and edited the PR description in order to start things off by taking this approach...

johnduffell:
great so quick! I'll take a look tomorrow morning

app/configuration/Config.scala
```diff
@@ -128,4 +128,8 @@ object Config {
   val getAddressIOApiUrl = config.getString("get-address-io-api.url")
   val getAddressIOApiKey = config.getString("get-address-io-api.key")
 
+  val fulfilmentLookupApiKey = config.getString("fulfilment-lookup-api-key")
```
johnduffell: above we used .key and .url presumably so we can add to the config as 
```
fulfilment-lookup-api {
  key=askjdfhgasdfgjk
  url=https://sdjkfhskdjfh
}
```

app/forms/DeliveryIssueForm.scala
```diff
@@ -0,0 +1,19 @@
...
+  val lookup: Form[TrackDelivery] = Form(
+    mapping(
+      "subscriptionName" -> nonEmptyText.transform[Name](Name, _.get),
+      "issueDate" -> jodaLocalDate("dd MMMM yyyy")
+    )(TrackDelivery.apply)(TrackDelivery.unapply)
+  )
+
+}
+
+case class TrackDelivery(subscriptionName: Name, issueDate: LocalDate)
```
johnduffell: `TrackDeliveryRequest` ?

app/services/FulfilmentLookupService.scala
```diff
@@ -0,0 +1,61 @@
...
+            \/-(validLookup.value)
+          case error: JsError =>
+            logger.error("Failed to parse response as FulfilmentLookup")
+            -\/(s"Response was successful but body could not be parsed")
+        }
+      } else {
+        val errorBody = response.body.string()
+        response.body.close()
+        val message = s"Failed to perform delivery tracking for ${trackDelivery.subscriptionName} due to error: ${errorBody}"
+        logger.error(message)
```
johnduffell: you're passing the message back already, I'd recommend not error logging at this stage but pushing that to the outside.  we can log at debug here (not that it'll go anywhere at the moment)

app/controllers/AccountManagement.scala
```diff
@@ -132,6 +133,21 @@ object ManageDelivery extends ContextLogging {
...
 
+  def fulfilmentCheck(trackDelivery: TrackDelivery)(implicit request: Request[TrackDelivery], touchpoint: TouchpointBackend.Resolution): Future[Result] = {
+    implicit val tpBackend = touchpoint.backend
+    logger.info(s"Attempting to perform tracking lookup: $trackDelivery")
+    val futureLookupAttempt = FulfilmentLookupService.lookupSubscription(tpBackend.environmentName, trackDelivery)
+    futureLookupAttempt.map { lookupAttempt => lookupAttempt match {
+        case \/-(lookup) =>
+          logger.info(s"Successful delivery tracking for $trackDelivery; showing deliveryTrackingSuccess")
+          Ok(views.html.account.deliveryTrackingSuccess(lookup, trackDelivery.issueDate))
+        case -\/(_) =>
```
johnduffell: so here, keep hold of the message in the left, and error log at this point (as this is the primary entry point/controller

app/controllers/AccountManagement.scala
```diff
@@ -396,4 +412,11 @@ object AccountManagement extends Controller with ContextLogging with CatalogProv
     implicit val resolution: TouchpointBackend.Resolution = TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)
     ManageWeekly.renewThankYou
   }
+
+  def trackDelivery: Action[TrackDelivery] = accountManagementAction.async(parse.form(TrackDeliveryForm.lookup)) { implicit request =>
+    implicit val resolution: TouchpointBackend.Resolution = TouchpointBackend.forRequest(PreSigninTestCookie, request.cookies)
+    val lookup = request.body
+    ManageDelivery.fulfilmentCheck(lookup)
```
johnduffell: inline this - as the top level def is ideally responsible for turning a web request into some nice scala object, and some nice scala object back into a web response.  as it is we're splitting that in half so the function signature of fulfilmentCheck is kind of crossing the abstraction levels

pvighi:
If I'm not mistaken if we merge this PR  it would be the first time we use the bucket fulfilment-output-test from production code. Maybe we should rename it to something else. 

Also can we trust that the salesforce file will always be in to the S3 folder fulfilment-lookup  checks?

app/services/FulfilmentLookupService.scala
```diff
@@ -0,0 +1,61 @@
...
+        jsonBody.validate[FulfilmentLookup] match {
+          case validLookup: JsSuccess[FulfilmentLookup] =>
+            \/-(validLookup.value)
+          case error: JsError =>
+            logger.error("Failed to parse response as FulfilmentLookup")
+            -\/(s"Response was successful but body could not be parsed")
+        }
+      } else {
+        val errorBody = response.body.string()
+        response.body.close()
```
johnduffell: you're doing a close in the else here, but not in the id part.  if you need the string regardless, do all that low level stuff in the httpClient function and just return a string, then do what you need with the string safely outside.

app/views/account/deliveryTrackingSuccess.scala.html
```diff
@@ -0,0 +1,36 @@
...
+    date: LocalDate
+)
+
+@main("Delivery Status | The Guardian", bodyClasses=List("is-wide")) {
+    <main class="page-container gs-container gs-container--slim">
+        @fragments.page.header("Delivery Status", None, List("l-padded"))
+        <section class="section-slice section-slice--bleed section-slice--limited">
+            @if(fulfilmentLookup.subscriptionInFile) {
+                <p>
+                    Your issue is out for delivery.
```
johnduffell: I'd prefer something like 
'Your issue was allocated for delivery on datedatedate.  If you have still not received your issue after the expected date, please contact....etcetc'

johnduffell:
can't review the JS much, but everything looks great with some suggestions above!

jacobwinch:
@pvighi yes you're right - I initially started using that folder just because I knew that's where I could get hold of the files in S3, but perhaps we should rename it... I've made a note to check that the scheduled download is working reliably before shipping this too.

app/views/account/deliveryTrackingFailure.scala.html
```diff
@@ -0,0 +1,13 @@
+@main("Delivery Status | The Guardian", bodyClasses=List("is-wide")) {
+    <main class="page-container gs-container gs-container--slim">
+        @fragments.page.header("Sorry, we were unable to determine your delivery status", None, List("l-padded"))
+        <section class="section-slice section-slice--bleed section-slice--limited">
+            In order to resolve this issue, please contact customer services on: <a class="u-link" href="tel:+44 330 333 6767">+44 (0) 330 333 6767</a>
+            or email <a class="u-link" href="mailto:subscriptions@@theguardian.com">subscriptions@@theguardian.com</a>.
+        </section>
+        <section class="section-slice--bleed mma-section">
+            <a class="button button--primary button--large" href="@routes.AccountManagement.processLogin">
```
paulbrown1982: does this need ().url on the end?

assets/javascripts/modules/track/trackDeliveryFields.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
paulbrown1982: It's best to be consistent with using semicolons in ES6 derivative modules, either use them on every line or only when inlining.

assets/javascripts/modules/track/trackDeliveryFields.jsx
```diff
@@ -0,0 +1,26 @@
...
+import ReactDOM from 'react-dom';
+import moment from 'moment'
+import formElements from './formElements'
+import CustomDatePicker from '../react/customDatePicker'
+
+require('react-datepicker/dist/react-datepicker.css');
```
AWare: can this use an import statement instead?

assets/javascripts/modules/trackDelivery.js
```diff
@@ -0,0 +1,21 @@
...
+define([
+    'modules/track/formElements'
+], function (
+    formElements
+) {
+    'use strict';
+    function init() {
+        if (formElements.$TRACK_DELIVERY_FORM.length) {
+            require(['modules/track/trackDeliveryFields'], function(trackDeliveryFields) {
+                trackDeliveryFields.default.init();
```
AWare: might be able to avoid a `.default.` if the whole file was es6

PR #975, opened 2017-08-14
Move the delivery instructions back to the SoldTo contact

paulbrown1982:
Moved the delivery instructions back to the SoldTo contact now that we can definitely sync down from Salesforce into that object.

See: https://github.com/guardian/membership-common/pull/510

cc @AWare @johnduffell @pvighi @lmath @jacobwinch


johnduffell:
thanks for upgrading the other membership-common stuff at the same time

PR #976, opened 2017-08-14
When suspending delivery, ignore broken (backwards) holidays

johnduffell:
Just bump membership common- it seems to work with a test sub in uat A-S00070317

https://github.com/guardian/membership-common/pull/513

@jacobwinch


johnduffell:
this is working ok in prod now

PR #983, opened 2017-09-07
Strip quotation marks from delivery instructions

jacobwinch:
Every now and then somebody includes a " character in their Delivery Instructions. This breaks the csv parser I'm using in fulfilment lookup. Presumably it could also cause problems for the fulfilment-lambdas project.

We could spend time adding client-side validation, and preventing users from doing this, but I think stripping these characters out before we write them into Zuora/Salesforce is sufficient (and also probably less frustrating for the user completing the form).

We need to do something similar in Salesforce to prevent CEX from including these characters when manually updating address details.

### DIFF
--- a/app/model/SubscriptionData.scala
+++ b/app/model/SubscriptionData.scala
@@ -66,7 +66,9 @@ case class PaperData(
   deliveryAddress: Address,
   deliveryInstructions: Option[String],
   plan: CatalogPlan.Paper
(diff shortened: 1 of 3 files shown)


paulbrown1982:
I agree, stripping them is easier. I can't imagine any delivery instructions which absolutely need quotation marks to make sense.

PR #984, opened 2017-09-13
Convert 'Delivery Tracking' to 'Delivery Problem Reporting'

jacobwinch:
In https://github.com/guardian/subscriptions-frontend/pull/972, we added an MVP version of 'delivery tracking'. Now that we've observed the basics working in [PROD](https://membership-logs.gutools.co.uk/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(_source),index:'logstash-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'message:%20%22Successful%20delivery%20tracking%22')),sort:!('@timestamp',desc))), it feels like it's worth converting this into self-service functionality for delivery problems.

The backend work required - which adds logic to raise cases in Salesforce - is in progress [here](https://github.com/guardian/fulfilment-lookup/pull/6) (a few minor tweaks to the case parameters are still required).

Whilst working on this, I noticed that the manage page for home delivery is still very 'holiday suspension' focused, when this page actually offers some other options now. Consequently, I've taken a first stab at making the copy more generic here too. 

**Current manage page:**
image

**New manage page:**
image

**Report a delivery problem:**
image

**Problem reported successfully:**
image

**Failure to raise problem:**
image

@paulbrown1982 - how do we normally review this type of copy change (and the new copy that I've added) for subs? Suggestions are welcome!

cc @johnduffell @pvighi @AWare @lmath


app/controllers/AccountManagement.scala
```diff
@@ -353,12 +354,13 @@ object AccountManagement extends Controller with ContextLogging with CatalogProv
 
     val futureMaybeFutureManagePage = for {
       subscription <- OptionT(eventualMaybeSubscription).filter(!_.isCancelled)
+      account <- OptionT(tpBackend.zuoraService.getAccount(subscription.accountId).map{Some(_)}.recover{case t: Throwable => None})
```
jacobwinch: This call used to happen inside [billingSchedule](https://github.com/guardian/membership-common/blob/e9e2118695c7c5d52463b8a7300a339795bdf959/src/main/scala/com/gu/memsub/services/PaymentService.scala#L91)

app/controllers/AccountManagement.scala
```diff
@@ -353,12 +354,13 @@ object AccountManagement extends Controller with ContextLogging with CatalogProv
...
       pendingHolidays <- OptionT(tpBackend.suspensionService.getUnfinishedHolidays(subscription.name, now).map(_.toOption))
-      billingSchedule <- OptionT(tpBackend.commonPaymentService.billingSchedule(subscription.id, subscription.accountId, numberOfBills = 13).map(Some(_):Option[Option[BillingSchedule]]))
+      billingSchedule <- OptionT(tpBackend.commonPaymentService.billingSchedule(subscription.id, account, numberOfBills = 13).map(Some(_):Option[Option[BillingSchedule]]))
```
jacobwinch: So we aren't actually making an extra call here, we just use [this version](https://github.com/guardian/membership-common/blob/e9e2118695c7c5d52463b8a7300a339795bdf959/src/main/scala/com/gu/memsub/services/PaymentService.scala#L93) of billingSchedule instead of the old one.

app/controllers/AccountManagement.scala
```diff
@@ -353,12 +354,13 @@ object AccountManagement extends Controller with ContextLogging with CatalogProv
...
     } yield {
       val maybeFutureManagePage = subscription.planToManage.product match {
         case Product.Delivery => subscription.asDelivery.map { deliverySubscription =>
-          Future.successful(ManageDelivery(errorCodes, pendingHolidays, billingSchedule, deliverySubscription))
+          Future.successful(ManageDelivery(errorCodes, pendingHolidays, billingSchedule, deliverySubscription, account))
```
jacobwinch: And that means we can now pass the account through, which gives us the sfContactId (required for case raising)

app/services/FulfilmentLookupService.scala
```diff
@@ -16,25 +16,30 @@ object FulfilmentLookupService extends StrictLogging {
...
 
-  def buildRequest(environment: String, trackDelivery: TrackDeliveryRequest): okhttp3.Request = {
+  def buildRequest(environment: String, deliveryIssue: ReportDeliveryIssue): okhttp3.Request = {
     val apiDateFormatter = DateTimeFormat.forPattern("yyyy-MM-dd")
-    val formattedDate = apiDateFormatter.print(trackDelivery.issueDate)
-    val body = RequestBody.create(MediaType.parse("application/json"), Json.obj("subscriptionName" -> trackDelivery.subscriptionName.get, "issueDate" -> formattedDate).toString())
+    val formattedDate = apiDateFormatter.print(deliveryIssue.issueDate)
+    val json = Json.obj(
+      "subscriptionName" -> deliveryIssue.subscriptionName.get,
+      "sfContactId" -> deliveryIssue.sfContactId,
```
jacobwinch: This is the change that the fulfilment-lookup API actually needs

conf/routes
```diff
@@ -56,7 +56,7 @@ POST        /manage/suspend                     controllers.AccountManagement.pr
 GET         /manage/signout                     controllers.AccountManagement.logout
 POST        /manage/renew                       controllers.AccountManagement.processRenewal
 GET         /manage/renew/thank-you             controllers.AccountManagement.renewThankYou
-POST        /manage/track-delivery              controllers.AccountManagement.trackDelivery
+POST        /manage/report-issue                controllers.AccountManagement.reportDeliveryIssue
```
paulbrown1982: Maybe change to /report-problem

paulbrown1982:
👍 but I think we need to be aware of the words 'issue' and 'problem' and use them judiciously. 

1. Customers have a problem. To solve that problem they raise an Issue.
2. Also note, Issue has an overloaded meaning in magazine publishing (e.g. Issue 30), so I'd use 'problem' wherever possible in customer messaging.

jacobwinch:
Thanks @paulbrown1982 - both good points. Because of the overloading (which I actually ran into once or twice whilst coding this), I've converted issue to problem everywhere to keep the code easy to follow.

paulbrown1982:
👍

jacobwinch:
Looking at the test failure now...

jacobwinch:
Looks like we timed out after 30 seconds (after pressing submit payment) on the first test:
https://saucelabs.com/beta/tests/fe79a6f0b37a455b8c59f54f248af37f/commands#82
https://saucelabs.com/beta/tests/fe79a6f0b37a455b8c59f54f248af37f/commands#132

The second test ran fine and I haven't touched this area of the code, so I'm going to re-run the tests.

jacobwinch:
Okay, I'm going to put that test failure down to flaky sandbox backend services and move on...

PR #985, opened 2017-09-14
Use CODE fulfilment API for test users running against PROD app

jacobwinch:
After shipping https://github.com/guardian/subscriptions-frontend/pull/984, I wanted to perform an end-to-end test in Production. But I'd forgotten to do this...

Note that this requires an update for the private PROD config file.

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -128,8 +128,7 @@ object Config {
   val getAddressIOApiUrl = config.getString("get-address-io-api.url")
   val getAddressIOApiKey = config.getString("get-address-io-api.key")
 
-  val fulfilmentLookupApiKey = config.getString("fulfilment-lookup-api.key")
-
-  val fulfilmentLookupApiUrl = config.getString("fulfilment-lookup-api.url")
+  def fulfilmentLookupApiKey(env: String) = config.getString(s"fulfilment-lookup-api.$env.key")
+  def fulfilmentLookupApiUrl(env: String) = config.getString(s"fulfilment-lookup-api.$env.url")
 
 }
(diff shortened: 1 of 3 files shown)


app/configuration/Config.scala
```diff
@@ -128,8 +128,7 @@ object Config {
   val getAddressIOApiUrl = config.getString("get-address-io-api.url")
   val getAddressIOApiKey = config.getString("get-address-io-api.key")
 
-  val fulfilmentLookupApiKey = config.getString("fulfilment-lookup-api.key")
-
-  val fulfilmentLookupApiUrl = config.getString("fulfilment-lookup-api.url")
+  def fulfilmentLookupApiKey(env: String) = config.getString(s"fulfilment-lookup-api.$env.key")
+  def fulfilmentLookupApiUrl(env: String) = config.getString(s"fulfilment-lookup-api.$env.url")
```
johnduffell: there's already a stageConfig for zuora and salesforce -
 https://github.com/guardian/subscriptions-frontend/pull/985/files#diff-f2117ad9c5563c42557d76930e24b9adL69
https://github.com/guardian/subscriptions-frontend/pull/985/files#diff-f2117ad9c5563c42557d76930e24b9adL85
does this fit in with the same structure in the config?

johnduffell:
great! my only q is about whether we want to follow the existing convention for splitting normal/default backend config rather than a new one?
👍

app/configuration/Config.scala
jacobwinch: Yeah, I considered this for a while too - in the end I decided not to bloat TouchPoint backend any further. I think this whole area needs to be tidied up / renamed, but I'll save that one for a rainy day...

jacobwinch:
^^^ There's something strange going on with the tests in Saucelabs right now - the label swapping is due to me messing around with an old Travis build...

PR #990, opened 2017-11-03
Send acquisition events

joelochlann:
Send acquisition events to Ophan so we can track subscriptions in our acquisitions table along with other reader revenue channels

I need to write some tests for this but thought I'd get some eyes on it in the meantime

### DIFF
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -47,7 +47,10 @@ object CommonActions {
 
   val StoreAcquisitionDataAction = new ActionBuilder[Request] {
     def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]) = block(request).map(result => {
-      request.getQueryString("acquisitionData").fold(result)(a => result.withSession(request.session.data.toSeq ++ Seq("acquisitionData" -> a) :_*))
+      request.getQueryString("acquisitionData").fold(result)(a => {
+        val sessionWithAcquisitionData = request.session.data.toSeq ++ Seq("acquisitionData" -> a)
+        result.withSession(sessionWithAcquisitionData: _*)
+      })
     })
   }
 
--- a/app/actions/CommonActions.scala
+++ b/app/actions/CommonActions.scala
@@ -45,6 +45,12 @@ object CommonActions {
 
   val CachedAction = resultModifier(Cached(_))
 
+  val StoreAcquisitionDataAction = new ActionBuilder[Request] {
+    def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]) = block(request).map(result => {
+      request.getQueryString("acquisitionData").fold(result)(a => result.withSession(request.session.data.toSeq ++ Seq("acquisitionData" -> a) :_*))
+    })
+  }
+
   val CSRFCachedAsyncAction = (block: Request[_] => Future[Result]) => CSRFCheck(action = CachedAction.async(block), config = CSRFConfig.global.copy(checkContentType = (x: Option[String]) => true))
 
   def noCache(result: Result): Result = NoCache(result)
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -30,7 +30,6 @@ import utils.RequestCountry._
 import utils.TestUsers.{NameEnteredInForm, PreSigninTestCookie}
 import views.html.{checkout => view}
 import views.support.{PlanList, BillingPeriod => _, _}
-import com.gu.acquisition.services.{MockOphanService, OphanService}
 
 import scala.Function.const
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
@@ -277,9 +276,13 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
       }
 
       val promotion = subscribeRequest.genericData.promoCode.map(_.get).flatMap(code => tpBackend.promoService.findPromotion(NormalisedPromoCode.safeFromString(code)))
-      val s = MockOphanService.submit(SubscriptionAcquisitionComponents(subscribeRequest, promotion, acquisitionData))
 
-      s.fold(println, println)
+      OphanService(isTestService = tpBackend.environmentName != "PROD")
+        .submit(SubscriptionAcquisitionComponents(subscribeRequest, promotion, acquisitionData))
+        .bimap(
+          err => logger.warn(s"Error submitting acquisition data to Ophan. ${err}"),
+          submission => logger.info(s"Submitted acquisition data to Ophan. ${submission}")
+        )
 
       logger.info(s"User successfully became subscriber:\n\tUser: SF=${r.salesforceMember.salesforceContactId}\n\tPlan: ${subscribeRequest.productData.fold(_.plan, _.plan).name}\n\tSubscription: ${r.subscribeResult.subscriptionName}")
       Ok(Json.obj("redirect" -> routes.Checkout.thankYou().url)).withSession(session)
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -39,6 +39,8 @@ import scalaz.std.option._
 import scalaz.std.scalaFuture._
 import scalaz.syntax.applicative._
 import scalaz.{NonEmptyList, OptionT}
+import cats.instances.future._
+
 object Checkout extends Controller with LazyLogging with CatalogProvider {
 
   import SessionKeys.{Currency => _, UserId => _, _}
@@ -273,7 +275,10 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
       if (acquisitionData.isEmpty) {
         logger.warn(s"No acquisitionData in session")
       }
-      MockOphanService.submit(SubscriptionAcquisitionComponents(subscribeRequest, acquisitionData))
+
+      val s = MockOphanService.submit(SubscriptionAcquisitionComponents(subscribeRequest, acquisitionData))
+
+      s.fold(println, println)
 
       logger.info(s"User successfully became subscriber:\n\tUser: SF=${r.salesforceMember.salesforceContactId}\n\tPlan: ${subscribeRequest.productData.fold(_.plan, _.plan).name}\n\tSubscription: ${r.subscribeResult.subscriptionName}")
       Ok(Json.obj("redirect" -> routes.Checkout.thankYou().url)).withSession(session)
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -269,7 +269,11 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
         _ + _
       }
 
-      MockOphanService.submit(SubscriptionAcquisitionComponents(subscribeRequest))
+      val acquisitionData = request.session.get("acquisitionData")
+      if (acquisitionData.isEmpty) {
+        logger.warn(s"No acquisitionData in session")
+      }
+      MockOphanService.submit(SubscriptionAcquisitionComponents(subscribeRequest, acquisitionData))
 
       logger.info(s"User successfully became subscriber:\n\tUser: SF=${r.salesforceMember.salesforceContactId}\n\tPlan: ${subscribeRequest.productData.fold(_.plan, _.plan).name}\n\tSubscription: ${r.subscribeResult.subscriptionName}")
       Ok(Json.obj("redirect" -> routes.Checkout.thankYou().url)).withSession(session)
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -30,6 +30,7 @@ import utils.RequestCountry._
 import utils.TestUsers.{NameEnteredInForm, PreSigninTestCookie}
 import views.html.{checkout => view}
 import views.support.{PlanList, BillingPeriod => _, _}
+import com.gu.acquisition.services.{MockOphanService, OphanService}
 
 import scala.Function.const
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
@@ -267,6 +268,8 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
       val session = (productData ++ userSessionFields ++ appliedCodeSession ++ subscriptionDetails).foldLeft(request.session - AppliedPromoCode - PromotionTrackingCode) {
         _ + _
       }
+
+      MockOphanService.submit(SubscriptionAcquisitionComponents(subscribeRequest))
 
       logger.info(s"User successfully became subscriber:\n\tUser: SF=${r.salesforceMember.salesforceContactId}\n\tPlan: ${subscribeRequest.productData.fold(_.plan, _.plan).name}\n\tSubscription: ${r.subscribeResult.subscriptionName}")
       Ok(Json.obj("redirect" -> routes.Checkout.thankYou().url)).withSession(session)
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -276,7 +276,8 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
         logger.warn(s"No acquisitionData in session")
       }
 
-      val s = MockOphanService.submit(SubscriptionAcquisitionComponents(subscribeRequest, acquisitionData))
+      val promotion = subscribeRequest.genericData.promoCode.map(_.get).flatMap(code => tpBackend.promoService.findPromotion(NormalisedPromoCode.safeFromString(code)))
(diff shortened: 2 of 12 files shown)


app/actions/CommonActions.scala
```diff
@@ -45,6 +45,12 @@ object CommonActions {
 
   val CachedAction = resultModifier(Cached(_))
 
+  val StoreAcquisitionDataAction = new ActionBuilder[Request] {
+    def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]) = block(request).map(result => {
+      request.getQueryString("acquisitionData").fold(result)(a => result.withSession(request.session.data.toSeq ++ Seq("acquisitionData" -> a) :_*))
```
johnduffell: a should be  `acquisitionDataFromQueryString` I guess?
this line is quite long and may be good to split into a val or def or something just to get names on some of these things

johnduffell: so here it's only added into the session once the request has completed, so it won't be in the session during the current request.  is that correct?

app/controllers/Checkout.scala
```diff
@@ -268,6 +270,20 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
+      if (acquisitionData.isEmpty) {
+        logger.warn(s"No acquisitionData in session")
+      }
+
+      val promotion = subscribeRequest.genericData.promoCode.map(_.get).flatMap(code => tpBackend.promoService.findPromotion(NormalisedPromoCode.safeFromString(code)))
+
+      OphanService(isTestService = tpBackend.environmentName != "PROD")
+        .submit(SubscriptionAcquisitionComponents(subscribeRequest, promotion, acquisitionData))
+        .bimap(
+          err => logger.warn(s"Error submitting acquisition data to Ophan. ${err}"),
```
Mullefa: `{}` unnecessary when interpolating vals

app/controllers/Checkout.scala
```diff
@@ -268,6 +270,20 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
+          submission => logger.info(s"Submitted acquisition data to Ophan. ${submission}")
```
Mullefa: Just checking, are we ok logging the submission considering it contains e.g. the browser id ?

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+        )
+      )
+    }
+
+    private def parseJsonSafely(json: String): Option[JsValue] = {
+      try {
+        Some(Json.parse(json))
+      } catch {
+        case scala.util.control.NonFatal(t) =>
+          logger.warn(s"""Unable to parse "$json" as JSON. ${t.getMessage()}""")
```
Mullefa: Think it's idiomatic in Scala to call `getMessage` without empty parameters

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    private def referrerAcquisitionDataFromJSON(json: String): Option[ReferrerAcquisitionData] = {
+      val parsedJson = parseJsonSafely(json)
+
+      parsedJson
+        .map(json => Json.fromJson[ReferrerAcquisitionData](json))
+        .flatMap({
+          case JsSuccess(referrerAcquisitionData, _) => Some(referrerAcquisitionData)
+          case e: JsError => {
```
Mullefa: Enclosing `{}` not needed

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+          None
+      }
+    }
+
+    private def referrerAcquisitionDataFromJSON(json: String): Option[ReferrerAcquisitionData] = {
+      val parsedJson = parseJsonSafely(json)
+
+      parsedJson
+        .map(json => Json.fromJson[ReferrerAcquisitionData](json))
+        .flatMap({
```
Mullefa: Enclosing `()` not needed

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+          case JsSuccess(referrerAcquisitionData, _) => Some(referrerAcquisitionData)
+          case e: JsError => {
+            logger.warn(s"Unable to decode JSON $parsedJson to an instance of ReferrerAcquisitionData. ${JsError.toJson(e).toString()}")
```
Mullefa: Don't need to call `toString()` explicitly

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+)
+
+
+object SubscriptionAcquisitionComponents {
+
+  implicit object subscriptionAcquisitionSubmissionBuilder
+    extends AcquisitionSubmissionBuilder[SubscriptionAcquisitionComponents] with LazyLogging {
+
+    def buildOphanIds(components: SubscriptionAcquisitionComponents): Either[String, OphanIds] = {
+      import components._
```
Mullefa: If you're doing an import, you might as well `import  components.subscribeRequest.genericData._`

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+    private def printOptionsFromPaperData(p: PaperData): Option[PrintOptions] = {
+      // Explicit imports because Product and Benefit have name collisions
+      import com.gu.memsub.Product.{Delivery, Voucher, Weekly}
+      import com.gu.memsub.Benefit.{MondayPaper, TuesdayPaper, WednesdayPaper, ThursdayPaper, FridayPaper, SaturdayPaper, SundayPaper, Digipack}
+
+      val paperDays = p.plan.charges.benefits.list.collect {
+        case p: PaperDay => p
+      }
+
+      sealed trait When
```
Mullefa: Don't think this trait hierarchy should be created per function call.

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+      val when = Some(paperDays.sortBy(_.dayOfTheWeekIndex)) collect {
+        case List(SundayPaper) => Sunday
+        case List(SaturdayPaper, SundayPaper) => Weekend
+        case List(MondayPaper, TuesdayPaper, WednesdayPaper, ThursdayPaper, FridayPaper, SaturdayPaper) => Sixday
+        case List(MondayPaper, TuesdayPaper, WednesdayPaper, ThursdayPaper, FridayPaper, SaturdayPaper, SundayPaper) => Everyday
+      }
+
+      val hasDigipack = p.plan.charges.benefits.list.contains(Digipack)
+
+      val printProduct = Some(when, p.plan.product, hasDigipack) collect {
```
Mullefa: Champion effort!

app/services/OphanService.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import com.gu.okhttp.RequestRunners
+
+object OphanService {
+
+  def apply(isTestService: Boolean): com.gu.acquisition.services.OphanService =
+    if (isTestService) {
+      com.gu.acquisition.services.MockOphanService
+    } else {
+      com.gu.acquisition.services.OphanService.prod(RequestRunners.client)
```
Mullefa: I don't think we want to create a new Ophan service per non-test service request.

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+      try {
+        Some(Json.parse(json))
+      } catch {
+        case scala.util.control.NonFatal(t) =>
+          logger.warn(s"""Unable to parse "$json" as JSON. ${t.getMessage()}""")
+          None
+      }
+    }
+
+    private def referrerAcquisitionDataFromJSON(json: String): Option[ReferrerAcquisitionData] = {
```
Mullefa: Had a play around - here is what a Scalaz implementation could look like...

```scala
private def referrerAcquisitionDataFromJSON(json: String): Option[ReferrerAcquisitionData] = {
  import \/._
  fromTryCatchNonFatal(Json.parse(json))
    .leftMap(err => logger.warn(s"""Unable to parse "$json" as JSON. ${err.getMessage}"""))
    .flatMap { jsValue =>
      Json.fromJson[ReferrerAcquisitionData](jsValue).fold(
        err => left(logger.warn(s"Unable to decode JSON $jsValue to an instance of ReferrerAcquisitionData. ${JsError.toJson(err)}")),
        data => right(data)
      )
    }
    .toOption
}
```
Not sure if it's preferable?

app/controllers/Checkout.scala
```diff
@@ -268,6 +270,20 @@ object Checkout extends Controller with LazyLogging with CatalogProvider {
...
 
+      val acquisitionData = request.session.get("acquisitionData")
+      if (acquisitionData.isEmpty) {
+        logger.warn(s"No acquisitionData in session")
+      }
+
+      val promotion = subscribeRequest.genericData.promoCode.map(_.get).flatMap(code => tpBackend.promoService.findPromotion(NormalisedPromoCode.safeFromString(code)))
+
+      OphanService(isTestService = tpBackend.environmentName != "PROD")
+        .submit(SubscriptionAcquisitionComponents(subscribeRequest, promotion, acquisitionData))
```
johnduffell: is this blocking? or does it return a future? we should probably Future.sequence it into the response if so.

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+import com.gu.stripe.Stripe.Charge
+import com.typesafe.scalalogging.LazyLogging
+import forms.SubscriptionsForm
+import ophan.thrift.event.PaymentProvider.{Gocardless, Stripe}
+import ophan.thrift.event._
+import play.api.libs.json._
+
+import scala.collection.Set
+
+//import com.gu.acquisition.model.ReferrerAcquisitionData.referrerAcquisitionDataReads
```
johnduffell: unused import

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+          browserId = subscribeRequest.genericData.ophanBrowserId
+        )
+      )
+    }
+
+    private def parseJsonSafely(json: String): Option[JsValue] = {
+      try {
+        Some(Json.parse(json))
+      } catch {
+        case scala.util.control.NonFatal(t) =>
```
johnduffell: why not just wrap this in the Try {...} as it catches the same thing.  you can do an onFailure to log and then a toOption

app/controllers/Checkout.scala
joelochlann: it returns an EitherT of a Future, but it's fire & forget so I don't think we need to wait for it

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+          case JsSuccess(referrerAcquisitionData, _) => Some(referrerAcquisitionData)
+          case e: JsError => {
+            logger.warn(s"Unable to decode JSON $parsedJson to an instance of ReferrerAcquisitionData. ${JsError.toJson(e).toString()}")
+            None
+          }
+        })
+    }
+
+    private def printOptionsFromPaperData(p: PaperData): Option[PrintOptions] = {
+      // Explicit imports because Product and Benefit have name collisions
```
johnduffell: :joy:

app/controllers/Checkout.scala
joelochlann: Good point, this is more for debug purposes - will remove

app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -0,0 +1,181 @@
...
+        case (Some(Sunday), Voucher, true) => PrintProduct.VoucherSundayPlus
+        case (Some(Weekend), Voucher, false) => PrintProduct.VoucerWeekend
+        case (Some(Weekend), Voucher, true) => PrintProduct.VoucerWeekendPlus
+        case (Some(Sixday), Voucher, false) => PrintProduct.VoucherSixday
+        case (Some(Sixday), Voucher, true) => PrintProduct.VoucherSixdayPlus
+        case (Some(Everyday), Voucher, false) => PrintProduct.VoucherEveryday
+        case (Some(Everyday), Voucher, true) => PrintProduct.VoucherEverydayPlus
+
+        case (_, _: Weekly, false) => PrintProduct.GuardianWeekly
+        case (_, _: Weekly, true) => PrintProduct.GuardianWeeklyPlus
```
johnduffell: the thing that scares me about this is this information should be in the Catalog, but it's been explicitly not represented there (for better or worse) and now it's hard coded somewhere non-resuable which may not be updated if the actual catalog changes.
ideally the catalog would be updated to actually know about all the different products and you could just read it from there, but clearly that's a more scary refactor.

Also, does guardian weekly plus even exist?

assets/javascripts/modules/checkout/submit.js
```diff
@@ -78,18 +84,20 @@ define([
...
-    function send() {
+    function send(ophanPageViewId) {
         var form = formEls.$CHECKOUT_FORM[0];
         var data = serializer([].slice.call(form.elements));
         var submitEl = formEls.$CHECKOUT_SUBMIT[0];
         var errorElement = $('.js-error');
 
         errorElement.text('');
 
+        data.ophanBrowserId = cookie.getCookie('bwid');
```
paulbrown1982: Have you checked that the subscribe.theguardian.com site sets these cookies if visited not via any other theguardian.com domain? Right now subs includes the membership.js Ophan tag, but it might get inserted after this .send?

app/model/SubscriptionData.scala
```diff
@@ -54,7 +54,10 @@ case class SubscriptionData(
    personalData: PersonalData,
    paymentData: PaymentData,
    promoCode: Option[PromoCode],
-   currency: Currency
+   currency: Currency,
+   ophanPageViewId: Option[String],
+   ophanVisitId: Option[String],
+   ophanBrowserId: Option[String]
```
johnduffell: I notice all the person/payment data is grouped into case classes, maybe it's worth having an OphanData case class to keep this neat?

app/services/OphanService.scala
johnduffell: agreed, we did have some issues with leaking thread pools when creating services more than once, so be very careful with that.
I think we have a cloudwatch metric for thread-count, so please check that for a few days after deploying this

johnduffell:
great work, can't really comment on the JS, maybe @AWare or @paulbrown1982 might give it a quick scan.

app/actions/CommonActions.scala
joelochlann: yep that's correct!

app/model/SubscriptionAcquisitionComponents.scala
joelochlann: yeah, I reckon it is! 💎

assets/javascripts/modules/checkout/submit.js
```diff
@@ -25,16 +27,20 @@ define([
...
+             formElements,
+             ophan,
+             cookie) {
     'use strict';
 
     function submitHandler() {
         if (formEls.$CHECKOUT_SUBMIT.length) {
-            formEls.$CHECKOUT_SUBMIT[0].addEventListener('click', submit, false);
+            formEls.$CHECKOUT_SUBMIT[0].addEventListener('click', function(e) {
+                ophan.loaded.then(function(o) { submit(e, o.viewId) });
```
joelochlann: Thinking about this line, I'm wondering what happens if Ophan doesn't load - will that prevent submission of the form?

app/model/SubscriptionAcquisitionComponents.scala
joelochlann: It's a very good point. But most changes to the catalog will also require us to update our Thrift definitions anyway, so even if the catalog code could automatically handle a change in the mapping of delivery days to print products, our tracking would still probably break if we didn't know about these changes. I think a better approach is to ensure that the Acquisitions team know about any significant changes to the subscription products on offer

joelochlann: Also, I've removed Guardian Weekly Plus - you're right, it doesn't exist!

assets/javascripts/modules/checkout/submit.js
joelochlann: OK I've addressed this now!

Mullefa:
🙌 🙌 🙌 🙌 🙌

PR #994, opened 2017-11-09
Make warnings fatal and fix warnings

joelochlann:
#990 caused 500 errors in production because of a non-exhaustive pattern match. If warnings were actually treated as errors, this issue would have failed the build and been fixed before making it to production.

Would you be happy to enable this setting on subscriptions-frontend?


johnduffell:
I'd be surprised if there's not an option to only make this particular case (or a  subset) fatal (re membership common)

PR #997, opened 2017-11-13
Bump referrer acquisition data version and add support for abTests 

jranks123:
The new version of `acquisition-event-model` allows multiple AB tests to be passed through. This PR 
allows subs frontend to be able to deal with that, whilst also allowing backwards compatibility with the old, now deprecated, `abTest` field.

### DIFF
--- a/app/model/SubscriptionAcquisitionComponents.scala
+++ b/app/model/SubscriptionAcquisitionComponents.scala
@@ -106,6 +106,14 @@ object SubscriptionAcquisitionComponents {
       printOptions
     }
 
+    def getAbTests(referrerAcquisitionData: Option[ReferrerAcquisitionData]): Option[AbTestInfo] =
+     referrerAcquisitionData.flatMap { r =>
+       r.abTests match {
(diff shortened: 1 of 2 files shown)


app/model/SubscriptionAcquisitionComponents.scala
```diff
@@ -106,6 +106,15 @@ object SubscriptionAcquisitionComponents {
       printOptions
     }
 
+    def getAbTests(referrerAcquisitionData: Option[ReferrerAcquisitionData]): Option[AbTestInfo] = {
+       val abTest = referrerAcquisitionData.flatMap(_.abTest)
```
joelochlann: might be simpler to map `abTest` to `AbTestInfo(Set(ab))` and then do `getOrElse`. and I think you can do this all inside the `.flatMap`

```scala
referrerAcquisitionData.flatMap { r =>
  r.abTest
    .map(ab => AbTestInfo(Set(ab)))
    .getOrElse(r.abTests.map(AbTestInfo(_))
}

```

PR #1000, opened 2017-11-14
Fixed a bug where customers in Australia could not renew their one-off Guardian Weekly subscriptions

paulbrown1982:
Fixed a bug where customers in Australia could not renew their one-off Guardian Weekly subscriptions. 

It was caused by Stripe Checkout not using the same Stripe account as the serverside interactions. There is no country change on the /manage page so the page renders with the correct country chosen.

Changes in javascriptFirstSteps fix the issue, the stuff in TouchpointBackend and CheckoutService is just some small refactoring to reduce the number of items sitting on TouchpointBackend and required within the CheckoutService.

Tested on code, sub id in dev Zuora: A-S00043952

cc @jacobwinch @AWare

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -68,8 +68,7 @@ class CheckoutService(
   zuoraProperties: ZuoraProperties,
   promoService: PromoService,
   promoPlans: DiscountRatePlanIds,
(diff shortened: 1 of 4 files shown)


AWare:
I thought sending the wrong key when creating the handler, and the correct one when opening it worked. Do we need to propagate this fix?

PR #1008, opened 2017-11-22
Black friday

rupertbates:
We have been asked to add some new copy and promo codes to the site for a 'Black Friday' promotion.

I have also made a few other changes:
* Remove an old endpoint related to an S&C test
* Extracted quite a bit of Scala template code out into fragments to reduce code duplication
* Removed the 'Save up to 75% on the app store price' bullet point as apparently this has been in the backlog for a while.


app/services/BlackFriday.scala
```diff
@@ -0,0 +1,55 @@
+package services
+
+import configuration.Config
+import model.DigitalEdition
+import model.promoCodes.{GuardianWeekly, _}
+import org.joda.time.DateTime
+
+//TODO: Remove after 3rd December 2017
```
paulbrown1982: So all that's needed to tidy things up after it self-deactivates is to delete this object and just remove any usage of it throughout the templates? i.e. there's no need to replace it with anything?

If so, if this is made an implicit class of something, and is forcibly the only overriding/implicit class, can we make it so that just deleting the class will work in future, and no need to modify the templates?

paulbrown1982:
👍 Nice work. I just left a comment for interest/posterity sake, please don't let it block the release.

app/services/BlackFriday.scala
rupertbates: It will require a bit more refactoring than that because this object now provides the promocodes that are passed through the templates. I could have moved that responsibility into a separate class and had this just deal with whether the promotion was running or not, but even then that other object would have had to have had some awareness of the promotion so it seemed easier just to put it in here for now.

PR #1015, opened 2017-12-04
Fixes some renewal/react bugs

AWare:
Had a this.props.{...} hanging around a functional component, and did not check a card was actually returned by members-data-api.


johnduffell:
thanks for fixing that! 👍 🚀

PR #1016, opened 2017-12-04
lint our jsx files

AWare:
This turns on jsxlint, and fixes most of the errors. 

The properties not checked error is reduced to a warning because it will take a lot of work to fix.


AWare:
image

`react/prop-types` is all 51 warnings. we should consider adding Prop-Types or whether this linter rule should be disabled.

paulbrown1982:
Looks great, just a couple of little bits I'd like to understand more about. Let's chat tomorrow.

app/views/fragments/checkout/promoCode.scala.html
```diff
@@ -19,7 +19,7 @@
         <div><strong>✓&nbsp;&nbsp;Promotion applied</strong></div>
         <div class="js-promo-code-snippet"></div>
         <div class="u-note">
-             <a class="u-link js-promo-code-tsandcs" href="#" target="_blank">See full terms and conditions</a>
+             <a class="u-link js-promo-code-tsandcs" href="#" target="_blank" rel="noopener noreferrer">See full terms and conditions</a>
```
paulbrown1982: What does this mean/do?

assets/javascripts/modules/react/customDateRangePicker.jsx
```diff
@@ -2,79 +2,88 @@ import React from 'react'
...
+
+     displayName = 'CustomDateRangePicker';
+
+     defaultProps = (() => {
         const maxWeeks = 6;
         return {
             remainingDays: (maxWeeks * 7),
             firstSelectableDate: moment(),
-            lastStartDate: moment().add(1, 'year')
+            laststartDate: moment().add(1, 'year')
```
paulbrown1982: Why was this case changed?

app/views/fragments/checkout/promoCode.scala.html
AWare: without it the window object is available to the calling page.https://mathiasbynens.github.io/rel-noopener/

AWare:
This also fixes people being told payment details updated correctly when they did not.

paulbrown1982:
Tested on CODE before merging.

PR #1019, opened 2017-12-06
Swap customer.experience for customer.operations in CAS staff access list

jacobwinch:
I've spoken to a few people in CEX and none of them are able to access this tool. I think this might be down to the Google group that we're currently using, hence this PR.

cc @paulbrown1982


paulbrown1982:
👍 the access does not work canonically, so if you're part of a group which is part of one of these groups, it does not work for you.

jacobwinch:
Thanks @paulbrown1982 - I'll ship this and see how we get on, but that'll be useful info for future problems with this!

PR #1021, opened 2017-12-07
Switch renew to use fetch instead of reqwest.

AWare:
Renew wasn't handling fail states particularly gracefully, and we were seeing successful renews shown the failure screen, this should show them a better error- but gives us more granularity if it continues.

### DIFF
--- a/assets/javascripts/modules/renew/renew.es6
+++ b/assets/javascripts/modules/renew/renew.es6
@@ -103,7 +103,6 @@ export function send(state, errorHandler) {
                     //We got a 200, and then couldn't parse the JSON.
                     //The renewal probably worked, and we should not really be in this branch.
                     errorHandler(renewNetworkErrorMessage)
-
                 })
                 return;
             }
--- a/assets/javascripts/modules/renew/renew.es6
+++ b/assets/javascripts/modules/renew/renew.es6
@@ -19,10 +19,9 @@ export class SortCode {
 
 }
 
-let renewErrorMessage = 'Sorry, your subscription could not be renewed. Please contact us at gwsubs@theguardian.com or call +44 (0) 330 333 6767. Lines open weekdays 8am-8pm, weekend 8pm-6pm';
+let renewErrorMessage = {title:'Sorry', body:'Your subscription could not be renewed. Please contact us at gwsubs@theguardian.com or call +44 (0) 330 333 6767. Lines open weekdays 8am-8pm, weekend 8pm-6pm'};
 
-let renewNetworkErrorMessage = `A response was not received when attempting to renew your subscription. Please refresh this page and try again. If you have seen this error before: please contact us at gwsubs@theguardian.com or call +44 (0) 330 333 6767. Lines open weekdays 8am-8pm, weekend 8pm-6pm
-`
+let renewNetworkErrorMessage = {title:'Connection Error', body:'Please refresh the page to see the current status of your subscription. Alternatively, please contact us at gwsubs@theguardian.com or call +44 (0) 330 333 6767. Lines open weekdays 8am-8pm, weekend 8pm-6pm'};
 
 export function validAccount(accountNumber) {
     return /^\d{6,10}$/.test(accountNumber);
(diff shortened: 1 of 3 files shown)


assets/javascripts/modules/renew/renew.es6
```diff
@@ -19,8 +18,12 @@ export class SortCode {
 
 
 }
+
 let renewErrorMessage = 'Sorry, your subscription could not be renewed. Please contact us at gwsubs@theguardian.com or call +44 (0) 330 333 6767. Lines open weekdays 8am-8pm, weekend 8pm-6pm';
 
+let renewNetworkErrorMessage = `A response was not received when attempting to renew your subscription. Please refresh this page and try again. If you have seen this error before: please contact us at gwsubs@theguardian.com or call +44 (0) 330 333 6767. Lines open weekdays 8am-8pm, weekend 8pm-6pm
```
jacobwinch: Do we need the first sentence here?

assets/javascripts/modules/renew/renew.es6
```diff
@@ -81,20 +84,39 @@ export function send(state, errorHandler) {
...
+            body: JSON.stringify(payload),
+            headers: new Headers({
+                'Csrf-Token': document.querySelector('input[name="csrfToken"]').getAttribute('value'),
+                'Content-Type':  'application/json'
+            }),
+            credentials: 'include'
+        })
+        fetch(request).then((response) => {
+            if (response.status == 200) {
+                console.log(response)
```
jacobwinch: Remove this log statement

assets/javascripts/modules/renew/renew.es6
```diff
@@ -81,20 +84,39 @@ export function send(state, errorHandler) {
...
+                response.json().then((json) => {
+                    window.location.assign(json.redirect);
+                }).catch((e) => {
+                    Raven.captureException(e)
+                    //We got a 200, and then couldn't parse the JSON.
+                    //The renewal probably worked, and we should not really be in this branch.
+                    errorHandler(renewNetworkErrorMessage)
```
jacobwinch: This error needs to be extremely carefully worded if we're going to return it when we receive a 200 from the API

assets/javascripts/modules/renew/renew.es6
```diff
@@ -81,20 +84,39 @@ export function send(state, errorHandler) {
...
+                return;
+            }
+            Raven.captureException(`Renewal failed with status code ${response.status}`);
+            if (response.status == 500 || response.status == 400) {
+                errorHandler(renewErrorMessage);
+                return;
+            }
+            errorHandler(renewNetworkErrorMessage)
+            return;
         }).catch((r) => {
```
jacobwinch: So this case will capture network timeouts etc.? Out of interest, what timeout do we use for this request?

jacobwinch: It's quite possible that the cases we were seeing where users saw the failure screen but the renew actually succeeded were due to us giving up on waiting for a response, whilst members-data-api was still processing the request.

assets/javascripts/modules/renew/renew.es6
AWare: I'm not 100% on the wording.

assets/javascripts/modules/renew/renew.es6
AWare: I think it's appropriate to show the network message. 
If we see this:
- we've changed the response
- somehow the user can't parse valid json
- the message has been changed in flight

The user should refresh then call, I think.

assets/javascripts/modules/renew/renew.es6
AWare: I think it's likely that is the case. 

I think the timeout we're likely to be hitting is the fastly one. IIRC we increased it to match the Zuora one ages ago, but this could be an expensive enough call to be a problem.

AWare:
image
image

jacobwinch:
👍 once the log statement is removed

PR #1022, opened 2017-12-08
correctly send public key to members-data-api

AWare:
Still trying to debug identity subscriptions not being identified correctly by /manage 

But this fixes a nasty bug with the renewal flow on /manage

I'd like to see this as an [ERROR] instead of a [WARN]:
https://github.com/guardian/members-data-api/blob/64f4342e570d25da8bdb4e0d4f09a4ea7ace28e3/membership-attribute-service/app/controllers/AccountController.scala#L123-L123

### DIFF
--- a/app/controllers/AccountManagement.scala
+++ b/app/controllers/AccountManagement.scala
@@ -68,7 +68,7 @@ object SessionSubscription extends StrictLogging {
         d.leftMap(e => logger.warn(s"Error looking up SF Contact for logged in user with Identity ID ${identityUser.user.id}: $e")).toOption.flatten
       })
       zuoraSubscription <- OptionT(tpBackend.subscriptionService.current[ContentSubscription](salesForceUser).map{subs =>
-        if (subs.length > 1) logger.warn("User with multiple subscriptions, only serving first.")
+        if (subs.length > 1) logger.warn(s"Logged in user with Identity ID ${identityUser.user.id}: with ${subs.length} subscriptions, only serving first.")
         subs.headOption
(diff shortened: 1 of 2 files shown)


app/controllers/AccountManagement.scala
```diff
@@ -67,7 +67,10 @@ object SessionSubscription extends StrictLogging {
       salesForceUser <- OptionT(tpBackend.salesforceService.repo.get(identityUser.user.id).map { d =>
         d.leftMap(e => logger.warn(s"Error looking up SF Contact for logged in user with Identity ID ${identityUser.user.id}: $e")).toOption.flatten
       })
-      zuoraSubscription <- OptionT(tpBackend.subscriptionService.current[ContentSubscription](salesForceUser).map(_.headOption/*FIXME if they have more than one they can only manage the first*/))
+      zuoraSubscription <- OptionT(tpBackend.subscriptionService.current[ContentSubscription](salesForceUser).map{subs =>
+        if (subs.length > 1) logger.warn(s"Logged in user with Identity ID ${identityUser.user.id}: with ${subs.length} subscriptions, only serving first.")
```
jacobwinch: This for comprehension is getting pretty hard to read - can this be split out into a separate function?

jacobwinch:
👍 - what was the impact of this bug?

app/controllers/AccountManagement.scala
AWare: Yes, the whole block needs rewriting to deal with multiple subs.

PR #1024, opened 2017-12-12
Add new promo codes and copy for Christmas flash sale

rupertbates:
This PR replaces https://github.com/guardian/subscriptions-frontend/pull/1023/files which I'm going to close.

Marketing have decided to run another 'Flash sale' similar to Black Friday except that it only applies to digipack. 

To do this I've reused the Black Friday code but renamed it to the more generic Flash Sale. Apparently these are going to be an ongoing feature of our marketing efforts from now on so we might as well leave this code in long term.

### DIFF
--- a/app/views/fragments/productLists/productListUk.scala.html
+++ b/app/views/fragments/productLists/productListUk.scala.html
@@ -9,12 +9,7 @@
         <h2 class="block__title">Paper + digital</h2>
         <div class="block__info">
             <ul class="block__list">
-                @if(inOfferPeriod) {
-                    <li class="block__list-item"><strong>Subscribe today and save an extra 50% for the first 3 months</strong></li>
-                    <li class="block__list-item">Year-round savings on the retail price</li>
-                } else {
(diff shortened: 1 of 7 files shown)


app/views/fragments/productLists/productListUk.scala.html
```diff
@@ -1,5 +1,5 @@
 @import model.DigitalEdition.UK
-@import services.BlackFriday._
+@import services.FlashSale._
 @import model.promoCodes._
 
 @(promoCodes: Map[PromoCodeKey, String], includeEditionInUrl: Boolean = false)
```
jacobwinch: Some of this copy was being removed here: https://github.com/guardian/subscriptions-frontend/pull/1023/files#diff-fbdc75d5de16bcd9a56b38da9aa67504

If this sale only applies to Digital Pack, then I think this still needs to go?

rupertbates: Yes I just realised that doh! Good spot Jacob.

PR #1025, opened 2017-12-12
Updating copy

svillafe:
We need to delete Kindle Fire and Android tablet as a benefit. We don't support them anymore.

### DIFF
--- a/app/views/offers/offers_uk.scala.html
+++ b/app/views/offers/offers_uk.scala.html
@@ -16,9 +16,8 @@
                     <ul class="block__list">
                         <li class="block__list-item">14-day free trial of our digital pack</li>
                         @UK.digitalPackSaving.map { saving => <li class="block__list-item"><strong>Save up to @saving% on the app store price</strong></li> }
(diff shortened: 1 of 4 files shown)


app/views/offers/offers_uk.scala.html
```diff
@@ -16,9 +16,8 @@ <h2 class="block__title">Digital</h2>
                     <ul class="block__list">
                         <li class="block__list-item">14-day free trial of our digital pack</li>
                         @UK.digitalPackSaving.map { saving => <li class="block__list-item"><strong>Save up to @saving% on the app store price</strong></li> }
-                        <li class="block__list-item">Access to the Daily Edition and the advert-free Guardian App Premium Tier</li>
+                        <li class="block__list-item">Access to the Daily Edition and the advert-free Guardian App Premium Tier available on Apple and Android</li>
```
davidfurey: I think implies that the daily edition is available on android

svillafe: Yes, I will fix it in a new PR.

PR #1029, opened 2017-12-13
Update Identity ID for guest checkout.

AWare:
Requires https://github.com/guardian/membership-common/pull/555

Saves the identity id for guests in the post successful subscription call

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -203,12 +203,15 @@ class CheckoutService(
         EitherT(identityService.updateUserDetails(personalData, deliveryAddress)(authenticatedIdUser)).leftMap(addErrContext("Authenticated"))
       case None =>
         logger.info(s"User does not have an Identity account. Creating a guest account")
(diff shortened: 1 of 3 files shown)


app/services/CheckoutService.scala
```diff
@@ -185,7 +186,7 @@ class CheckoutService(
   private def storeIdentityDetails(
       subscribeRequest: SubscribeRequest,
       authenticatedUserOpt: Option[AuthenticatedIdUser],
-      memberId: ContactId): EitherT[Future, NonEmptyList[SubsError], IdentitySuccess] = {
+      memberId: ContactId,result: SubscribeResult): EitherT[Future, NonEmptyList[SubsError], IdentitySuccess] = {
```
jacobwinch: Need a space here

app/services/CheckoutService.scala
```diff
@@ -206,6 +207,7 @@ class CheckoutService(
           EitherT(salesforceService.repo.updateIdentityId(memberId, succ.userData.id.id)).swap.map(err =>
```
jacobwinch: Could pull succ.userData.id.id out into a val here (and perhaps rename succ whilst we're at it)

PR #1031, opened 2017-12-14
Remove kindle and android from digital pack promo image

davidfurey:


davidfurey:
verbal approval from @svillafe

PR #1036, opened 2017-12-22
Use the latest membership-common to write the new HolidayStart__c field

paulbrown1982:
Use the latest membership-common to ensure we write the HolidayStart__c field when creating a new holiday.

picture 284

cc @pvighi

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2 % "test",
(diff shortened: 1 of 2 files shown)


pvighi:
looks like a test doesn't compile

paulbrown1982:
All works fine in CODE. Was able to create a sub and set 3 holidays, including soon, over the renewal gap and back between the 2 other holidays again.

paulbrown1982:
Here's an example of a production future holiday suspension now showing the correct date:

picture 288

PR #1040, opened 2018-01-04
Marketing consent

rupertbates:
## Why is this PR necessary?
As part of our GDPR work, we need to change the way we collect marketing consent.
Trello card [here](https://trello.com/c/dtlSHP3x/1156-marketing-consent-on-subscribe)
cc @jfsoul

## What needs to be done?
1. Change the wording of the consent form - TODO: this is undergoing some UX analysis but will be completed before launch.
2. Change the way that consent is persisted to Identity:

## How?
Previously we have stored marketing consent by posting to the user/me endpoint with a boolean representing whether the user checked the consent checkbox. Although we will continue to do this for now, Identity will soon start ignoring this value in favour of a [new API](https://docs.google.com/document/d/1JDEpehzToi9aAMg4Fk7n_mnvQ_GOOf_kgOPSyWLBweA/)

### The new method
If user gives consent:
* We call Identity on the new consent-email endpoint to instruct them that the user has given consent to 'supporter' related emails. 
* Identity emails the user asking them to click a link to confirm their choice.

If user does not give consent then no call to the consent emailer is made.

### Guest users
For users who are not signed in during the checkout we create a guest user with Identity. The user then has the option to add a password on the thank you page. 

For users who do this and who have also given us marketing consent we now need to suppress the standard validation email which would otherwise be sent out when we set their password in Identity. 

This is described in more detail in [the doc mentioned above](https://docs.google.com/document/d/1JDEpehzToi9aAMg4Fk7n_mnvQ_GOOf_kgOPSyWLBweA/).

  
 **New copy:**
screen shot 2018-01-10 at 17 19 45

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -171,17 +171,27 @@ class CheckoutService(
                          promotion: Option[ValidPromotion[NewUsers]]): SubNel[PostSubscribeResult] = {
 
     val purchaserIds = PurchaserIdentifiers(contactId, user.map(_.user))
-    val res = (
-      storeIdentityDetails(subscriptionData, user, contactId, result).run |@|
-      sendETDataExtensionRow(result, subscriptionData, gracePeriod(promotion), purchaserIds, promotion)
-    ) { case(id, email) =>
-      (id.toOption.map(_.userData), id.swap.toOption.toSeq.flatMap(_.list) ++ email.swap.toOption.toSeq.flatMap(_.list))
-    }
+
+    val res = for {
+      id <- storeIdentityDetails(subscriptionData, user, contactId, result).run
+      _ <- sendConsentEmail(subscriptionData)
+      email <- sendETDataExtensionRow(result, subscriptionData, gracePeriod(promotion), purchaserIds, promotion)
+    } yield (id.toOption.map(_.userData), id.swap.toOption.toSeq.flatMap(_.list) ++ email.swap.toOption.toSeq.flatMap(_.list))
+
     EitherT(res.map(\/.right[FatalErrors, PostSubscribeResult]))
   }
 
   private def gracePeriod(promo: Option[ValidPromotion[NewUsers]]) =
     promo.flatMap(_.promotion.asDiscount).fold(zuoraProperties.gracePeriodInDays)(_ => Days.ZERO)
+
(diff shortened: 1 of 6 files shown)


app/services/IdentityService.scala
```diff
@@ -258,12 +281,21 @@ object IdentityApiClient extends IdentityApiClient with LazyLogging {
...
       .withCloudwatchMonitoringOfPost
   }
+
+  override val consentEmail: Email => Future[WSResponse] = { email =>
+    val endpoint = authoriseCall(WS.url(s"$identityEndpoint/consent-email"))
+
+    val json = Json.obj("email" -> email, "set-consents" -> List("supporter"))
+    endpoint.post(json)
+      .withWSFailureLogging(endpoint)
+      .withCloudwatchMonitoringOfPost
```
jacobwinch: I know you're just following the convention here, but are any of these Cloudwatch metrics providing any value?

rupertbates: Good question I would defer to you or @paulbrown1982 on that.

jacobwinch: Personally I'd be inclined to delete them all, rather than adding more. 

I don't think anyone looks at them, and I've had a quick skim over our AWS alarms and there don't seem to be any related to these metrics.

app/services/CheckoutService.scala
```diff
@@ -171,18 +171,28 @@ class CheckoutService(
...
 
   private def gracePeriod(promo: Option[ValidPromotion[NewUsers]]) =
     promo.flatMap(_.promotion.asDiscount).fold(zuoraProperties.gracePeriodInDays)(_ => Days.ZERO)
 
+  private def sendConsentEmail(subscribeRequest: SubscribeRequest): Future[\/[NonEmptyList[IdentityFailure], Unit]] = {
+    val personalData = subscribeRequest.genericData.personalData
+    if (personalData.receiveGnmMarketing)
+      identityService.consentEmail(personalData.email)
+    else
+      Future.successful(\/.right(Unit))
```
jacobwinch: Might be worth logging the case where we skip sending an email (at info level)?

jacobwinch:
A couple of minor comments, but overall this looks good to me 👍

app/services/IdentityService.scala
paulbrown1982: People only ever look at them when clutching at straws when not being able to resolve an issue via log lines.

rupertbates: It sounds like we should probably remove them then but I think that should be a separate PR

PR #1041, opened 2018-01-08
Awin affiliate tracking

rupertbates:
We are launching an affiliate scheme to pay commission to referrers when a we sell a product.

We are working with a company called Awin to do this.

[Trello card here](https://trello.com/c/pbQRuFKA/1117-affiliate-tags)

[Implementation docs here](https://drive.google.com/drive/folders/19AOGPkPbbFjqPJTdCcCba3EwpLiSHBwW)

*NB. I am in the process of checking the data we are sending to Awin with the GDPR team*

### DIFF
--- a/app/model/SubscriptionOps.scala
+++ b/app/model/SubscriptionOps.scala
@@ -54,6 +54,7 @@ object SubscriptionOps extends LazyLogging {
       }
     }
 
+    def firstPrice = firstPlan.charges.price.getPrice(currency).map(_.amount).getOrElse(0F)
     def firstPlan = sortedPlans.head
     def firstProduct = firstPlan.product
     def currentPlans = subscription.plans.list.filter(p => (p.start == now || p.start.isBefore(now)) && p.end.isAfter(now))
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -49,5 +49,7 @@
 
         @fragments.global.footer(product, edition)
         @fragments.footerJavaScript()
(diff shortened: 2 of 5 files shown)


app/model/SubscriptionOps.scala
```diff
@@ -54,6 +54,7 @@ object SubscriptionOps extends LazyLogging {
       }
     }
 
+    def firstPrice = firstPlan.charges.price.getPrice(currency).map(_.amount).getOrElse(0F)
```
paulbrown1982: As you're adding this to a string in the JSON in thankyou.scala.html, I think you can do .map(_.pretty).mkString

app/views/main.scala.html
```diff
@@ -49,5 +49,7 @@
 
         @fragments.global.footer(product, edition)
         @fragments.footerJavaScript()
+        <!--AWIN affiliate tracking tag - see here: https://drive.google.com/drive/folders/19AOGPkPbbFjqPJTdCcCba3EwpLiSHBwW -->
+         <script src="https://www.dwin1.com/11152.js" type="text/javascript" defer="defer"></script>
```
paulbrown1982: You could add the script to the DOM inside the init of awin.js, then if the user has opted out via DNT, or we're in test mode and so analyticsEnabled() returns false,  we won't load the JS from dwin1.com

johnduffell:
is this PR still needed?

johnduffell:
>28 days old, reopen/open another if needed

PR #1043, opened 2018-01-10
Updating stripe image

svillafe:
Updating Stripe image to use S3

cc: @paulbrown1982


svillafe:
It is working on PROD ✅

PR #1047, opened 2018-01-15
Post-Garnett Tidying

Ap0c:
## Why are you doing this?

Couple tweaks to make stuff look less broken.

## Changes

- Applied the new black to the identity popup links.
- Updated the primary button active and focus states to be more legible, and to match the hover state.

## Screenshots

**Before:**

popup-before

button-before

**After:**

popup-after

button-after


Ap0c:
Works in PROD.

PR #1050, opened 2018-01-18
Fix polyfill and widths of renew radio buttons in ie11.

AWare:
The final polyfill in the list was not loading as the get parameter separator wasn't separating as it still thought it was in a list. This also adds Array.prototype.findIndex


AWare:
If we look at the polyfills without the .min it will show us how they have parsed the query params.

Old polyfill: https://polyfill.guim.co.uk/v2/polyfill.js?features=default-3.6,Element.prototype.dataset,Array.prototype.includes,Array.find,fetch%26unknown=polyfill
New polyfill: https://polyfill.guim.co.uk/v2/polyfill.js?features=default-3.6,Element.prototype.dataset,Array.prototype.includes,Array.prototype.find,fetch,Array.prototype.findIndex,&unknown=polyfill

PR #1051, opened 2018-01-22
Changes to the GoCardless service

paulbrown1982:
Changes to the GoCardless service:

1. Made the service and its config extraction part of the TouchpointBackend so that it uses the sandbox/live environment accordingly.
2. Created specific API tokens for this application (GDPR Cyber Essentials) and added them to the new config structure. I've also updated S3 accordingly.
3. Updated the Review Details part of the UI so that 'Sort code' is listed before 'Account number', like it is in the editable section.
4. Updated the AJAX endpoint so that caching is turned off - this might have caused issues in the past?

Sandbox console:
picture 322

Live (Prod) console:
picture 321

cc @svillafe @johnduffell @jacobwinch @davidfurey

### DIFF
--- a/app/services/GocardlessService.scala
+++ b/app/services/GocardlessService.scala
@@ -1,23 +1,21 @@
 package services
 
+import com.gocardless.GoCardlessClient
+import com.gocardless.GoCardlessClient.Environment.{LIVE, SANDBOX}
 import com.gocardless.errors.GoCardlessApiException
 import com.gocardless.resources.BankDetailsLookup.AvailableDebitScheme
 import com.typesafe.scalalogging.LazyLogging
-import configuration.Config
 import model.DirectDebitData
-
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
+import touchpoint.GoCardlessToken
+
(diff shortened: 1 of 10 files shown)


app/services/GocardlessService.scala
```diff
@@ -1,23 +1,21 @@
...
 
-trait GoCardlessService {
-  def mandatePDFUrl(paymentData: DirectDebitData): Future[String]
-  def checkBankDetails(paymentData: DirectDebitData): Future[Boolean]
-}
+class GoCardlessService(goCardlessToken: GoCardlessToken) extends LazyLogging {
 
-object GoCardlessService extends GoCardlessService with LazyLogging {
-  lazy val client = Config.GoCardless.client
+  lazy val client: GoCardlessClient = GoCardlessClient.create(goCardlessToken.token, if (goCardlessToken.isProdToken) LIVE else SANDBOX)
```
paulbrown1982: LIVE (ie Prod) = https://api.gocardless.com
SANDBOX = https://api-sandbox.gocardless.com

PR #1052, opened 2018-01-22
Add message to updatePayment to show feature is unavailable.

AWare:
(Rather than attempt failed.)

### DIFF
--- a/assets/javascripts/modules/mma/updatePayment.jsx
+++ b/assets/javascripts/modules/mma/updatePayment.jsx
@@ -31,7 +31,6 @@ const handleStripeResponse = (t, url, key) => {
             if (resp.status == 200) {
                 return resp.json()
             } else {
-                console.log('wow this is bad')
                 Raven.captureException(`Renewal failed with status code ${resp.status}`)
                 return Promise.reject(`Renewal failed with status code ${resp.status}`)
             }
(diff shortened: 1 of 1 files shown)


AWare:
image
Could do with better phrasing.

assets/javascripts/modules/mma/updatePayment.jsx
```diff
@@ -54,11 +54,12 @@ const WAITING = 'waiting'
 const SUCCESS = 'success'
 const FAILURE = 'failure'
 const FORM = 'form'
+const UNAVAILABLE = 'unavailable'
 
 const CardUpdate = ({ card, handler }) => (<div><button className="button button--primary button--large" onClick={handler}>•••• •••• •••• {card && card.last4 || '••••'} Update Card</button></div>)
 const Success = () => (<p>Thank you, we have successfully updated your payment details.</p>)
 const Failure = ({ phone }) => <p>Unfortunately, we are unable to update your payment details at this time, please contact the call centre. {phone}</p>
```
jacobwinch: Perhaps this should ask the users to try again, since we failed whilst actually processing the update?

assets/javascripts/modules/mma/updatePayment.jsx
```diff
@@ -54,11 +54,12 @@ const WAITING = 'waiting'
...
-
+const Unavailable = ({ phone }) => (<p>Unfortunately, it is not possible to update your payment details online. To update your payment details, please contact the call centre. {phone}</p>)
```
jacobwinch: I like the idea of splitting this into a separate message 👍 

But if I understand it correctly, we are still trying to use this one message to cover a lot of cases:

1. We timed out waiting for members-data-api to retrieve the details (problem temporary)
2. members-data-api returned an error when attempting to retrieve the details (problem may be temporary)
3. members-data-api returned a 200, but the response body is empty ([because they don't have an identity account?](https://github.com/guardian/members-data-api/blob/master/membership-attribute-service/app/controllers/AccountController.scala#L197)) (problem is permanent)

Do you think we should refine the error handling further so that we aren't driving unnecessary contacts to the call centre?

jacobwinch:
I agree that finding the correct phrasing is tricky - in isolation your new phrasing looks good, but for somebody who has just renewed, is it clear that they don't need to call / do anything else here? Perhaps hiding the section if someone has just renewed is preferable?

AWare:
image

jacobwinch:
👍

PR #1056, opened 2018-01-25
Jan flash sale

rupertbates:
Add dates, copy and codes for a new flash sale. I've refactored things a bit to better accommodate sales where not all categories of product are included.


jacobwinch:
💯 🥇 🏆

PR #1057, opened 2018-01-26
Update react

AWare:
This updates react to 16, and gives us a nicer/newer datepicker.

This should probably land after we fix the address finder api.


package.json
```diff
@@ -80,6 +77,9 @@
   },
   "dependencies": {
     "babel-runtime": "^6.26.0",
+    "react": "^16.2.0",
```
paulbrown1982: Why would these move from devDependencies to full dependencies?

paulbrown1982:
👍 once the package conflict is resolved.

PR #1058, opened 2018-01-29
add compile time dependency injection to all controllers

johnduffell:
This is part of the steps to upgrade to play 2.6.  The deprecated globals are no longer present, so I am going through fixing all the deprecation warning in existing 2.5 code.

The first one was we were using GlobalSettings rather than a Module or ApplicationLoader.  I have added an ApplicationLoader to enable compile time DI, and construct the routes object with a set of created controllers.
Unfortunately the controllers are all constructed before the deprecated global variables are available for use.  This meant that anything that used touchpoint backend would try to initialise the actorsystem.  In the end I just made the whole object into a class and now the touchpoint backend is injected into the controllers.

In a later PR I will start to do the other deprecated/dependency injected fields.

This also relies on the changes to future handling I made in https://github.com/guardian/membership-common/pull/559

@pvighi @paulbrown1982 @jacobwinch @davidfurey any comments or +1s please!


app/Filters.scala
```diff
@@ -1,15 +0,0 @@
-import javax.inject.Inject
```
johnduffell: this has moved to MyApplicationLoader `MyComponents.httpFilters` just below https://github.com/guardian/subscriptions-frontend/pull/1058/files#diff-751cafce06c7a60432eacdc822ece49dR74

app/Global.scala
```diff
@@ -1,16 +0,0 @@
-import configuration.Config
```
johnduffell: this has moved to `MyApplicationLoader.load` just below https://github.com/guardian/subscriptions-frontend/pull/1058/files#diff-751cafce06c7a60432eacdc822ece49dR39

app/MyApplicationLoader.scala
```diff
@@ -0,0 +1,83 @@
...
+import play.api.http.DefaultHttpFilters
+import play.api.libs.concurrent.AkkaComponents
+import play.api.libs.ws.ahc.AhcWSComponents
+import play.api.mvc.EssentialFilter
+import play.filters.csrf.{CSRFComponents, CSRFFilter}
+import play.filters.headers.{SecurityHeadersConfig, SecurityHeadersFilter}
+import router.Routes
+import services.{IdentityService, TouchpointBackends}
+
+class MyApplicationLoader extends ApplicationLoader {
```
johnduffell: https://www.playframework.com/documentation/2.6.x/ScalaCompileTimeDependencyInjection
"The application loader is constructed and invoked every time the application is reloaded, to load the application.

This trait’s load method takes as an argument the application loader Context, which contains all the components required by a Play application that outlive the application itself and cannot be constructed by the application itself."

app/MyApplicationLoader.scala
```diff
@@ -0,0 +1,62 @@
...
+import play.api.ApplicationLoader.Context
+import play.api._
+import play.api.libs.ws.ahc.AhcWSComponents
+import play.api.mvc.EssentialFilter
+import play.filters.csrf.CSRFComponents
+import play.filters.headers.{SecurityHeadersConfig, SecurityHeadersFilter}
+import router.Routes
+import services.TouchpointBackends
+
+class MyApplicationLoader extends ApplicationLoader {
```
johnduffell: https://www.playframework.com/documentation/2.6.x/ScalaCompileTimeDependencyInjection
"The application loader is constructed and invoked every time the application is reloaded, to load the application.

This trait’s load method takes as an argument the application loader Context, which contains all the components required by a Play application that outlive the application itself and cannot be constructed by the application itself."

app/controllers/AccountManagement.scala
```diff
@@ -167,16 +163,18 @@ object ManageWeekly extends ContextLogging {
   import play.api.mvc.Results._
 
   // this sequencing concatenates errors if any, otherwise aggregates rights
-  def sequence[A](list: List[\/[String, A]]): \/[String, List[A]] = {
-    val errors = list collect {
-      case -\/(x) => x
+  def sequence[A](list: List[\/[NonEmptyList[String], A]]): \/[NonEmptyList[String], List[A]] = {
```
johnduffell: the change here is to make it concatenate the errors into a list instead of a string, which seems neater and helped the types

app/controllers/AccountManagement.scala
```diff
@@ -202,39 +200,42 @@ object ManageWeekly extends ContextLogging {
...
     implicit val tpBackend = resolution.backend
     implicit val rest = tpBackend.simpleRestClient
     implicit val zuoraRest = new ZuoraRestService[Future]
     implicit val flash = request.flash
     implicit val subContext = weeklySubscription
 
-    def getRenewalPlans(account: ZuoraRestService.AccountSummary, currency: Currency) = {
-      val catalog = tpBackend.catalogService.unsafeCatalog
+    def getRenewalPlans(account: ZuoraRestService.AccountSummary, currency: Currency): Future[\/[NonEmptyList[String], List[WeeklyPlanInfo]]] = {
+      EitherT(tpBackend.catalogService.catalog).flatMap { catalog =>
```
johnduffell: this function returns a future anyway so we might as well flatmap the future catalog rather than awaiting the catalog itself and then using it directly.  Then it down to the caller to wait.

This was needed because if the future times out, it throws an error.  This was not a problem before as with it being a lazy static val it seemed to retry.  When I changed touchpoint backend to be a class, it seemed to try once and then fail instantly, so I thought it's a good time as any to fix properly.

app/controllers/AccountManagement.scala
```diff
@@ -296,18 +298,18 @@ object ManageWeekly extends ContextLogging {
             returnError(e.getMessage, sub)
         }
       }
-      response.valueOr(error => Future(BadRequest(jsonError(error))))
+      response.run.map(_.valueOr(error => Future(BadRequest(jsonError(error))))).flatMap(identity)
```
johnduffell: there's no flatten on a future and I ended up with 2 futures in different places in the stack

app/controllers/Management.scala
```diff
@@ -22,22 +23,22 @@ import play.api.libs.ws.WSClient
 import scalaz.{Semigroup, Validation, ValidationNel}
 import scalaz.syntax.std.option._
 
-class Management @Inject()(override val wsClient: WSClient)  extends Controller with LazyLogging with OAuthActions {
+class Management(override val wsClient: WSClient, actorSystem: ActorSystem, fBackendFactory: TouchpointBackends)  extends Controller with LazyLogging with OAuthActions {
```
johnduffell: the @inject annotation is for runt ime DI which we aren't using in general.  So I have removed this as the controller is explicitly created now rather than relying on the framework to find out how to create one

app/controllers/PromoLandingPage.scala
```diff
@@ -23,102 +24,111 @@ import utils.Tracking.internalCampaignCode
...
-  }
+import scala.concurrent.Future
+import scalaz.OptionT
+import scalaz.std.scalaFuture._
+
+class PromoLandingPage (override val wsClient: WSClient, tpBackend: TouchpointBackend) extends Controller with OAuthActions {
+
+  private val catalogPromoLandingPage: Future[CatalogPromoLandingPage] = tpBackend.catalogService.catalog.map(_.map(c => new CatalogPromoLandingPage(c)).valueOr(e => throw new IllegalStateException(s"$e while getting catalog")))
+
+  class CatalogPromoLandingPage(val catalog: Catalog) {
```
johnduffell: this hasn't changed as much as it looks, I just moved all the things that relied on a catalog into this nested class so we could avoid changing the methods at all, but other methods that don't rely on a catalog would be unchanged so wouldn't end up returning a future.
Unfrotauntely due to all the code being "text" it thinks everything has changed anyway, slightly negating my effort...

In reality these classes should be separate, a lot of the classes I worked on were not  very cohesive, but that's not a job for this PR!

app/controllers/Promotion.scala
```diff
@@ -20,68 +20,81 @@ import views.support.Pricing._
...
+          val result = promo.validate(country)
 
+          def body = Json.obj(
+            "promotion" -> Json.toJson(promo),
+            "adjustedRatePlans" -> Json.toJson(ratePlans),
+            "isValid" -> result.isRight,
+            "errorMessage" -> result.swap.toOption.map(_.msg)
+          )
+
+          result.fold(_ => NotAcceptable /*!?! fix this*/ (body), _ => Ok(body))
```
johnduffell: github claims I added this (and I probably did) but not in this PR

app/services/IdentityService.scala
```diff
@@ -178,59 +180,12 @@ trait IdentityApiClient {
...
-  implicit class FutureWSLike(eventualResponse: Future[WSResponse]) {
-    /**
-     * Example of ID API Response: `{"status":"error","errors":[{"message":"Forbidden","description":"Field access denied","context":"privateFields.billingAddress1"}]}`
-     *
-     * @return
-     */
-    private def applyOnSpecificErrors(reasons: List[String])(block: WSResponse => Unit): Unit = {
-      def errorMessageFilter(error: JsValue): Boolean =
-        (error \ "message").asOpt[JsString].exists(e => reasons.contains(e.value))
+class IdentityApiClientImpl(wsClient: WSClient) extends IdentityApiClient with LazyLogging {
```
johnduffell: here I created a class so I can inject the ws client, but I only put the stuff that needs a ws client into it, and left the other stuff in the object (which is now below)

conf/routes
```diff
@@ -2,86 +2,86 @@
 # This file defines all application routes (Higher priority routes first)
 # ~~~~
 
-GET         /robots.txt                      @controllers.CachedAssets.at(path="/public", file="robots.txt")
-GET         /favicon.ico                     @controllers.CachedAssets.busted(path="images/favicons/32x32.ico")
+GET         /robots.txt                      controllers.CachedAssets.at(path="/public", file="robots.txt")
```
johnduffell: the @ sign means it uses the traditional way to create the controller, but now we switch to the compile time DI so I removed it

app/controllers/AccountManagement.scala
jacobwinch: 💯

app/controllers/AccountManagement.scala
```diff
@@ -325,12 +327,12 @@ object ManageWeekly extends ContextLogging {
 }
 
 
-class AccountManagement extends Controller with ContextLogging with CatalogProvider with CommonActions {
+class AccountManagement(fBackendFactory: TouchpointBackends) extends Controller with ContextLogging with CommonActions {
```
jacobwinch: What about 'backendFactory' or 'touchpointBackends' or 'thirdPartyEnvironment'? (I don't like fBackendFactory!)

app/controllers/Checkout.scala
```diff
@@ -42,160 +42,167 @@ import scala.util.Try
...
-        case Some(planList) => renderCheckoutFor(planList)
-        case None => Redirect(routes.Homepage.index())
+        matchingPlanList match {
+          case Some(planList) => renderCheckoutFor(planList)
+          case None => Redirect(routes.Homepage.index())
+        }
       }
+    }.valueOr { err =>
+      logger.error(s"failed landingPage: ${err.list.mkString(", ")}")
+      Future.successful(InternalServerError("oops"))
```
jacobwinch: There's still an "oops" here!

app/controllers/Checkout.scala
```diff
@@ -42,160 +42,167 @@ import scala.util.Try
...
-      matchingPlanList match {
-        case Some(planList) => renderCheckoutFor(planList)
-        case None => Redirect(routes.Homepage.index())
+        matchingPlanList match {
+          case Some(planList) => renderCheckoutFor(planList)
+          case None => Redirect(routes.Homepage.index())
+        }
       }
+    }.valueOr { err =>
+      logger.error(s"failed landingPage: ${err.list.mkString(", ")}")
```
jacobwinch: Would "Failed to load checkout for..." be easier to follow here?

app/controllers/DigitalPack.scala
```diff
@@ -26,11 +29,16 @@ class DigitalPack extends Controller with CommonActions {
...
     val digitalEdition = getById(code).getOrElse(INT)
-    val plan = TouchpointBackend.Normal.catalogService.unsafeCatalog.digipack.month
-    val price = plan.charges.price.getPrice(digitalEdition.countryGroup.currency).getOrElse(plan.charges.gbpPrice)
-    Ok(views.html.digitalpack.info(digitalEdition, price))
+    touchpointBackend.catalogService.catalog.map(_.map { catalog =>
+      val plan = catalog.digipack.month
+      val price = plan.charges.price.getPrice(digitalEdition.countryGroup.currency).getOrElse(plan.charges.gbpPrice)
+      Ok(views.html.digitalpack.info(digitalEdition, price))
+    }.valueOr { err =>
+      logger.error(s"failed landingPage: ${err.list.mkString(", ")}")
```
jacobwinch: "Failed to load the Digital Pack landing page..." ?

app/controllers/PromoLandingPage.scala
```diff
@@ -138,21 +150,27 @@ class PromoLandingPage @Inject()(override val wsClient: WSClient) extends Contro
       }
       val hreflang = Hreflangs(Config.subscriptionsUrl + routes.PromoLandingPage.preview(Some(country)).url, hreflangs)
 
-      getLandingPage(promotion, country, hreflang)(promoCode)
+      OptionT(catalogPromoLandingPage.map(_.getLandingPage(promotion, country, hreflang)(promoCode)))
     }
+    } yield aaa
```
jacobwinch: aaa

app/controllers/Promotion.scala
jacobwinch: 😆

conf/routes
jacobwinch: 👍

jacobwinch:
I have a few small comments about logging improvements / naming, but overall this looks like a great effort 👍 

Thanks for adding the comments about how/why you did things too - it made this much easier to review.

app/controllers/AccountManagement.scala
johnduffell: sorry yes I temporarily renamed TouchpointBackends everywhere to that on purpose, so I could use search and replace on "TouchpointBackend" without catching it.  I renamed the class to what I really wanted, but some variable names slipped through.  Updating.

app/controllers/Checkout.scala
johnduffell: 😆

app/controllers/DigitalPack.scala
johnduffell: why not

johnduffell:
we don't seem to have ant extra warnings appearing, and no 5xx errors, so I'm going to start on the next stage

johnduffell:
apparently this is the non-prefill problem that @AWare was talking about earlier, unfortunately the tests become a bit noisy

PR #1059, opened 2018-02-01
updated common version

pvighi:
get fixed country names

@jacobwinch @paulbrown1982

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2 % "test",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -46,7 +46,7 @@ libraryDependencies ++= Seq(
     ws,
     filters,
     PlayImport.specs2 % "test",
-    "com.gu" %% "membership-common" % "0.494",
+    "com.gu" %% "membership-common" % "0.496",
```
pvighi: looks like this brings in @johnduffell 's change as well

PR #1060, opened 2018-02-02
Fix error handler to use dependency injection rather than use the default

johnduffell:
Following updating things to use Dependency injection, unfortunately I missed the fact that the error handler used to be loaded from an old style class from reflection.

This meant that 404 errors were served the default way which didn't include caching, at which point our cacheing header checking code turned it into a 500.

I worked out what the problem was and wired in the correct functinality.

Ideally I can get this out before the end of the day, to avoid having to decide whether to roll back the main change over the weekend (to avoid messing up our alerting)

@jacobwinch @pvighi @paulbrown1982


pvighi:
looks good 👍

PR #916, opened 2017-04-18
use deb packages, systemd and auto update ami on deploy

pvighi:
similar to changes in [PR#77 in cas proxy in](https://github.com/guardian/content-authorisation-proxy/pull/77)

@paulbrown1982 @jacobwinch

PR #1063, opened 2018-02-05
Tighten up logging policy and delete unused configuration

jacobwinch:
### Why do we need this? 
GDPR / new best practices for log retention time.

### The changes 
* New log group name
* Add the new log group to Cloudformation
* Restrict allowed actions (full list is [here](http://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/permissions-reference-cwl.html))
* Restrict creation of log streams and put log events to the new, Cloudformed log group, so that we don't accidentally log events to a group without a retention policy.
* Delete upstart.conf, as it looks like this has been unused since https://github.com/guardian/subscriptions-frontend/pull/916 was merged.

## To-do
- [x] Test Cloudforming and deployment with CODE stack
- [x] Cloudform PROD stack (after +1)
- [ ] Merge this to master
- [ ] Delete old log group after 2 weeks


jacobwinch:
^^^ This looks like the known issue with failure to pre-populate fields on the checkout form.

PR #1065, opened 2018-02-07
Update play version to 2.6

johnduffell:
This PR actually updates the play version to 2.6.  It follows on from getting rid of the deprecation warnings in the previous PRs
https://github.com/guardian/subscriptions-frontend/pull/1058 - controllers
https://github.com/guardian/subscriptions-frontend/pull/1060 - http error handling
https://github.com/guardian/subscriptions-frontend/pull/1061 - play.isdev
https://github.com/guardian/subscriptions-frontend/pull/1062 - CSRF checker

It required a few knock on changes - this is the minimum to make it work together with minimum library changes, but there are a large number of warnings mainly about the execution context.  This will be addressed immediately after as a means to make this PR as small as it could be.  Unfortunately despite my efforts, this PR is much bigger than I prefer.

The only other issue is that this depends on the snapshot membership common with the play and scalaz update. https://github.com/guardian/membership-common/pull/558 This means it will need merging before this, thus blocking anyone else wanting to update other apps without updating play.  Maybe this isn't bad since Patricio is updating members data api and David is updating Membership frontend but it's a consideration.

I have added a load of my own comments to explain why I did various things, but feel free to challenge as appropriate.  @davidfurey @paulbrown1982 @pvighi @jacobwinch


build.sbt
```diff
@@ -45,9 +45,12 @@ libraryDependencies ++= Seq(
     cache,
     ws,
     filters,
+    jodaForms,
     PlayImport.specs2 % "test",
-    "com.gu" %% "membership-common" % "0.496",
-    "com.gu" %% "memsub-common-play-auth" % "1.2",
+    "com.gu" %% "membership-common" % "0.1-SNAPSHOT",
```
johnduffell: https://github.com/guardian/membership-common/pull/558

app/MyApplicationLoader.scala
```diff
@@ -27,30 +32,37 @@ class MyComponents(context: Context)
   extends BuiltInComponentsFromContext(context)
     with AhcWSComponents
     with CSRFComponents
+    with HttpConfigurationComponents
+    with AssetsComponents
```
johnduffell: Theoretically it could have been left as deprecated but working, but for some reason the assets controller running in backwards compatibility mode always 404ed locally.   rather than trying to diagnose it, it was quicker to actually update to DI.

app/MyApplicationLoader.scala
```diff
@@ -27,30 +32,37 @@ class MyComponents(context: Context)
   extends BuiltInComponentsFromContext(context)
     with AhcWSComponents
     with CSRFComponents
+    with HttpConfigurationComponents
```
johnduffell: I updated play google auth to the latest that uses 2.6, and now it generates the anti forgery token from the session id rather than putting a random token in there just for the purpose (to allow concurrent auth requests).   https://github.com/guardian/play-googleauth/pull/52
It needs the httpConfiguration to generate and check the token.

app/actions/CommonActions.scala
```diff
@@ -3,17 +3,21 @@ package actions
...
  import play.api.mvc._
+ import play.filters.csrf.CSRFCheck
 
- import scala.concurrent.Future
-trait CommonActions {
+ import scala.concurrent.{ExecutionContext, Future}
+final class CommonActions(executionContext: ExecutionContext, cSRFCheck: CSRFCheck, parser: BodyParser[AnyContent]) {
 
-  val StoreAcquisitionDataAction = new ActionBuilder[Request] {
+  val StoreAcquisitionDataAction = new ActionBuilder[Request, AnyContent] {
```
johnduffell: in play 2.6 the ActionBuilder also contains a parser for the content.  Usually we just want the default parser which looks at the content type if any, and deserialises with that if applicable.   This is passed in with an overridden field below.

app/com/gu/memsub/auth/common/KeystoreCredentials.scala
```diff
@@ -0,0 +1,7 @@
+package com.gu.memsub.auth.common
```
johnduffell: I've just copied these from this library to avoid horrible dependency annoyance, please let me know your opinion on keeping it separate or not.
https://github.com/guardian/memsub-common-play-auth

app/configuration/Config.scala
```diff
@@ -27,8 +27,6 @@ object Config {
 
   val playSecret = config.getString("play.crypto.secret")
 
-  lazy val googleAuthConfig = googleAuthConfigFor(config)
```
johnduffell: this has moved to be dependency injected.  to be honest if we're using DI I'd prefer to get rid of config object and just have separate case classes which each area of config only injected where relevant

app/controllers/AccountManagement.scala
```diff
@@ -97,7 +97,7 @@ object ManageDelivery extends ContextLogging {
       _ <- EitherT(tpBackend.suspensionService.renewIfNeeded(sub,newHoliday))
       // 26 because one year from now could be end of second years sub + 2 extra months needed in calculation to cover setting a 6-week suspension on the day before your 12th billing day!
       oldBS <- EitherT(tpBackend.commonPaymentService.flatMap(_.billingSchedule(sub.id, account, numberOfBills = 26).map(_ \/> "Error getting billing schedule")))
-      result <- EitherT(tpBackend.suspensionService.addHoliday(newHoliday, oldBS, account.billCycleDay, sub.termEndDate)).leftMap(getAndLogRefundError(_).map(_.code).list.mkString(","))
+      result <- EitherT(tpBackend.suspensionService.addHoliday(newHoliday, oldBS, account.billCycleDay, sub.termEndDate)).leftMap(getAndLogRefundError(_).map(_.code).list.toList.mkString(","))
```
johnduffell: scalaz kindly returns an IList (invariant list) instead of a real list, so you have to toList if you want a normal list.

app/controllers/OAuth.scala
```diff
@@ -1,23 +1,22 @@
...
+import play.api.mvc._
 
 import scala.concurrent.Future
 
-class OAuth @Inject()(val wsClient: WSClient) extends Controller with OAuthActions{
-  val ANTI_FORGERY_KEY = "antiForgeryToken"
-  implicit val iWsClient = wsClient
+class OAuth (val wsClient: WSClient, commonActions: CommonActions, oAuthActions: OAuthActions) extends Controller {
+
+  import commonActions._
```
johnduffell: this stuff is removed due to the play anti forgery changes in https://github.com/guardian/play-googleauth/pull/52 - I tested resubmitting with the same key over a minute after the origial request and it kicked it out with an error.  The error was a little more revealing than I'd like, but it was an out of date token, so it was working in some way.

app/views/account/details.scala.html
```diff
@@ -52,7 +52,7 @@ <h1 class="suspend-header__title">Manage your Guardian subscription</h1>
                             </legend>
                             <div class="field-panel__intro">
                                 <div class="field-note field-note--offset">
-                                @fragments.forms.securityNote()
+                                @_root_.views.html.fragments.forms.securityNote()
```
johnduffell: the imports in play 2.6 seems to work differently, so it finds the fragments package in the current directory, rather than the root one.  I've explicitly specified it in the affected places and it seems to get around it.

app/views/support/WeeklyPromotion.scala
```diff
@@ -137,13 +137,13 @@ object WeeklyPromotion {
       plan <- plans
       sixOrDiscount = isSixWeek(plan) || maybePromotion.exists(_.appliesTo.productRatePlanIds.contains(plan.id))
       pretty = maybePromotion.filter(_.appliesTo.productRatePlanIds.contains(plan.id)).flatMap(_.asDiscount)
-        .map { discountPromo => plan.charges.prettyPricingForDiscountedPeriod[scalaz.Id.Id, WeeklyLandingPage](discountPromo, currency) }
+        .map { discountPromo => plan.charges.prettyPricingForDiscountedPeriod[scalaz.Id.Identity, WeeklyLandingPage](discountPromo, currency) }
```
johnduffell: I'm not sure we should be using "Identity" as it's not supposed to be a covariant alternative to Id, it seems to be something else that doesn't quite make sense, and wasn't really confirmed by the scalaz people as even a good idea (but has been sitting in the code a while).  We could alternaticely implement our own in the style of David's CovariantNonEmptyList (a CovariantId) with the same implementation as the invariant one.
See my comments on https://github.com/guardian/membership-common/pull/558/files#r166326856

app/controllers/OAuth.scala
jacobwinch: 👍

app/com/gu/memsub/auth/common/KeystoreCredentials.scala
jacobwinch: We could pull this into membership-common instead, but I don't have a particularly strong view either way here...

jacobwinch: Perhaps @davidfurey has a view, as presumably we'll run into similar issues when we do the membership-frontend upgrade?

jacobwinch:
Looks good to me (👍 once the build is fixed)

app/views/promotion/newspaperLandingPage.scala.html
```diff
@@ -7,20 +7,20 @@
...
 @products = @{
     val adapter = planToOptions(promoCode, anyPromotion) _
     val voucherProduct = CollectionSubscriptionProduct(
         title = "",
         description = "",
         altPackagePath = "#delivery",
         options = {
-            voucherPlans.filter(_.charges.digipack.isEmpty).map(adapter).sortBy(_.weeklyPrice).reverse ++
-            voucherPlans.filter(_.charges.digipack.isDefined).map(adapter).sortBy(_.weeklyPrice).reverse
+            voucherPlans.filter(_.charges.digipack.isEmpty).map(adapter).sortBy(_.weeklyPrice).reverse.toList ++
```
paulbrown1982: I think the .toList is not necessary here as it's already been converted in line 10

app/com/gu/memsub/auth/common/KeystoreCredentials.scala
johnduffell: decided not to pursue this, as it will add more dependencies to membership common.  More code I mind less than more dependencies.

app/views/promotion/newspaperLandingPage.scala.html
johnduffell: even intellij told me that too! well spotted
image

PR #1069, opened 2018-02-13
Hide AppNexus Pixel

Ap0c:
# Why?

If the AppNexus pixel fails to load in, the alt text "AppNexus pixel" is shown to the user instead. This is probably not the best UX, as this isn't an image that they are aware of in the normal flow of the page. This PR removes that alt text.

**Note:** I can't figure out how to get the pixel to appear in DEV or CODE.

# Screenshots

subscribe-appnexus

### DIFF
--- a/assets/javascripts/modules/analytics/appnexus.js
+++ b/assets/javascripts/modules/analytics/appnexus.js
@@ -44,7 +44,6 @@ define([
         var parameterValue = pageCodes[pageType];
 
         var oImg = document.createElement('img');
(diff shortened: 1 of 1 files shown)


Ap0c:
Yeah, if the image doesn't load in the alt text [gets displayed](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-alt). I'm assuming that sometimes the network request for the pixel src fails, and we get that alt text instead.

Ap0c:
Looks like the pixel is being correctly inserted into the DOM and the network request is firing OK.

PR #1071, opened 2018-02-15
Don't use fulfilment-lookup response body

jacobwinch:
We don't use the body anymore, so there's no need to parse it. 

This also allows us to change the response of the API, which is taken care of in this PR:
https://github.com/guardian/fulfilment-lookup/pull/15


johnduffell:
great 👍

jacobwinch:
The test caught a timeout from UAT Salesforce:
WARN: okhttp request failure: POST <salesforceurl>/services/apexrest/RegisterCustomer/v1/

image

PR #1072, opened 2018-02-16
don't use deprecated execution context

johnduffell:
This uses the proper dependency injected execution context.

This is part of the post play 2.6 version update.

Please give me a +1!  Also ignore the branch name, I pushed in error to that branch and it created it again.  I could google how to fix it but it's an unnecessary danger!

@jacobwinch @paulbrown1982 @pvighi @davidfurey


app/configuration/Config.scala
```diff
@@ -104,12 +99,6 @@ object Config {
     }
   }
 
-  lazy val casService = {
```
johnduffell: this has moved to somewhere that deserves an execution context.  Cas"Service" isnt exactly configuration

app/services/IdentityService.scala
```diff
@@ -2,32 +2,29 @@ package services
...
-import model.error.IdentityService._
-import services.IdentityApiClient.{authoriseCall, identityEndpoint}
 import services.PersonalDataJsonSerialiser.convertToUser
 
-class IdentityService(identityApiClient: => IdentityApiClient) extends LazyLogging {
+import scala.concurrent.{ExecutionContext, Future}
+import scala.language.{higherKinds, implicitConversions}
+import scalaz.{Monad, NonEmptyList, \/}
+
+class IdentityService[M[_]](identityApiClient: => IdentityApiClient[M])(implicit monad: Monad[M]) extends LazyLogging {
```
johnduffell: this is a M instead of a Future because the tests need to get the result straight away to check, but the real one needs a Future as it will happen later.  This was the neatest way to do it that I could think of off hand.

test/services/IdentityServiceTest.scala
```diff
@@ -1,57 +1,54 @@
...
-      }
+      val res = makeIdentityService(client).doesUserExist(existingEmail)
+      res should be(true)
     }
 
     "returns false if looking up a user by email is unsuccessful" in {
-      makeIdentityService(client).doesUserExist("unknown@example.com").map { res =>
-        assert(!res)
-      }
+      val res = makeIdentityService(client).doesUserExist("unknown@example.com")
```
johnduffell: turn out this was mapping the future with an assertion, which was running on the default execution context.  This was a problem because the future was being discarded and nothing was waiting for the result.
As it happened, the test was checking for a 404, and the mock client on line 41 was just returning 200, so it was failing silently.  I think the code was actually working, but the test was wrong since it was written.
Now that it's all Id we should get the results reported and fail the test when there's an issue.

app/configuration/Config.scala
jacobwinch: 👍

test/services/IdentityServiceTest.scala
jacobwinch: Nice work!

johnduffell:
testing in prod passed at 14:11, however the old instances were not terminated until 14:17.
image
Looks like prout's method of working out when the deploy happened is not reliable - only one instance needs to be available for it to assume deploy finished.

PR #1073, opened 2018-02-19
Bump membership-common to 0.502

lmath:
I made a breaking change in membership-common here https://github.com/guardian/membership-common/pull/564 and so I need to bump the membership-common version to the latest.

cc @paulbrown1982 @jacobwinch @johnduffell @AWare @pvighi


johnduffell:
thanks, good tidiness! 👍 🥇

PR #1075, opened 2018-02-19
fix deprecation of the controller

johnduffell:
This gets rid of the deprecated Controller super class and uses the BaseController which can get its conponents from the injected controllercomponeets

https://www.playframework.com/documentation/2.6.x/Migration26#Scala-Controller-changes

@paulbrown1982 @pvighi @jacobwinch @davidfurey @lmath

### DIFF
--- a/app/MyApplicationLoader.scala
+++ b/app/MyApplicationLoader.scala
@@ -48,23 +48,25 @@ class MyComponents(context: Context)
 
   val httpClient = com.gu.okhttp.RequestRunners.futureRunner
 
+  implicit val cc = controllerComponents
+
   lazy val router: Routes = new Routes(
     httpErrorHandler,
-    new CachedAssets(assets),
-    new Homepage(commonActions),
+    new CachedAssets(assets, controllerComponents),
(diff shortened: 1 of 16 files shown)


app/MyApplicationLoader.scala
```diff
@@ -48,23 +48,25 @@ class MyComponents(context: Context)
...
 
+  implicit val cc = controllerComponents
+
   lazy val router: Routes = new Routes(
     httpErrorHandler,
-    new CachedAssets(assets),
-    new Homepage(commonActions),
+    new CachedAssets(assets, controllerComponents),
+    new Homepage(commonActions, controllerComponents),
     new Management(actorSystem = actorSystem, touchpointBackends, oAuthActions),
```
johnduffell: I wasn't sure whether it should be implicit or passed in explicitly, it happened this way because I used search and replace to refactor the class definition.  I'm happy to go one way or the other, or sit on the fence!

johnduffell:
as per a chat with David I've added all the implicit parameters explicitly in the app loader.

PR #1077, opened 2018-02-20
Adding the missing UB postcodes

paulbrown1982:
Adding the missing UB postcodes (UB10 and UB11 are covered by the substring of UB1)

cc @johnduffell


johnduffell:
👍

PR #1079, opened 2018-02-23
use encrypted AMI

pvighi:
tested in CODE


pvighi:
there was a network error or something which caused the teamcity build to fail. Re run the build and it worked the second time

PR #1080, opened 2018-02-26
Stop logging server errors on bad requests

jacobwinch:
Play's default for [logServerError](https://github.com/playframework/playframework/blob/8cea51a3360a500cc7c424d2b8b633d42d7c2f07/framework/src/play/src/main/java/play/http/DefaultHttpErrorHandler.java#L180-L195) explicitly uses the phrase 'Internal server error' - so it isn't appropriate to reuse this method when a bad request is received.


johnduffell:
great PR! thanks 👍 🎉

PR #1084, opened 2018-03-05
Remove bower for JS *and zxcvbn*

AWare:
This just overrides our js deps with differently specified versions. 

Except for: promise-polyfill and domready

We don't use domready, and we use polyfill.io for polyfilling promise now. 

We should also examine our CSS bower deps, but with less urgency.


AWare:
todo: remove bower'd zxcvbn

AWare:
removed bower'd zxcvbn

PR #1087, opened 2018-03-08
Remove Date Input Field

Ap0c:
# Why?

@CPKING discovered that the current date input field for vouchers and paper delivery allows the user to enter any value they want, including incorrect dates and invalid values. The fix is to remove the manual input field altogether, replacing it with an embedded date picker. Does anyone have thoughts on whether this could cause any problems?

cc @JustinPinner 

# Screenshots

**Before**

broken-input

**After**

embedded-input

### DIFF
--- a/assets/javascripts/modules/react/customDatePicker.jsx
+++ b/assets/javascripts/modules/react/customDatePicker.jsx
@@ -15,6 +15,7 @@ export default class CustomDatePick extends React.Component {
         return <DatePicker
             selected={this.state.startDate}
             onChange={this.handleChange}
+            inline
             {...this.props} />
     }
     


michaelwmcnamara:
Problem with this.
The current live version of this date-picker is accessible for keyboard only users - you can tab to it, open the field and pick a date.
The version in this branch is not accessible for keyboard-only users. You cannot select dates unless you use the mouse.

Currently this fix trades an edge case that has never come up outside of the building for a real accessibility issue which will definitely impact a portion of our traffic.

michaelwmcnamara:
Is it possible to catch this through validation of the date - either in the frontend or during processing ?

johnduffell:
great spot @michaelwmcnamara ! 💥

AWare:
This bug is also present on https://reactdatepicker.com/#example-12
I can't find anything in their issue tracker, but https://github.com/Hacker0x01/react-datepicker/issues/720 is similar. 

http://airbnb.io/react-dates/ has similar behaviour.

Ap0c:
We've decided not to go with this approach due to the accessibility issues that @michaelwmcnamara raised.

PR #1091, opened 2018-04-10
Lib vulnerabilities

jbguardian:
**1. logback-classic**
Vulnerable module: ch.qos.logback:logback-classic [CVE-2017-5929] - Logback before 1.2.0 has a serialization vulnerability affecting the SocketServer and ServerSocketReceiver components.
Added by:
- com.gu:kinesis-logback-appender@1.4.2 - 1.2.3 (***)
- com.getsentry.raven:raven-logback@8.0.3 - 1.2.1
- com.gu:tip_2.11@0.1.1 - 1.1.7

**2. jackson-databind**
https://github.com/guardian/zuora-auto-cancel/pull/65

**3. scala-compiler**
Vulnerable module: org.scala-lang:scala-compiler [CVE-2017-15288] - Upgrade to Scala 2.10.7 2.11.12, 2.12.4 or higher which restricts the sensitive file to be readable only by the owner.

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -340,7 +340,7 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
       val eventualMaybeSubscription = tpBackend.subscriptionService.get[com.gu.memsub.subsv2.SubscriptionPlan.ContentSubscription](Name(subsName))
       eventualMaybeSubscription.map { maybeSub =>
         maybeSub.map { sub =>
-          if (tpBackend.environmentName == "PROD") Tip.verify("Subscription purchased")
+          if (tpBackend.environmentName == "PROD") Tip.verify()
           Ok(view.thankyou(sub, passwordForm, resolution, promotion, startDate, edition, billingCountry))
         }.getOrElse {
           Redirect(routes.Homepage.index()).withNewSession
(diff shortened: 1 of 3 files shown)


paulbrown1982:
Adding @mario-galic to let him know that tip might have a Snyk vulnerability, once that's fixed downstream apps won't need to update its transitive dependencies.

mario-galic:
Latest [tip 0.3.3](https://github.com/guardian/tip/pull/37) addresses logback vulnarablity.

mario-galic:
I think we should be bumping direct dependencies only. Bump transitive dependencies can cause insidious bugs as direct dependencies were not compiled against those latest versions and during conflict SBT by default selects the latest.

jbguardian:
I agree but technically we are not upgrading anything as the logback version used by the project before was: 1.2.3, we are just making explicit its use.
Also tip version 0.3.3 is not working, compile error:
```subscriptions-frontend/app/controllers/Checkout.scala:343: not enough arguments for method verify: (pathName: String)scala.concurrent.Future[com.gu.tip.TipResponse].
Unspecified value parameter pathName.
if (tpBackend.environmentName == "PROD") Tip.verify()```

johnduffell:
@mario-galic 
> I think we should be bumping direct dependencies only. Bumping transitive dependencies can cause insidious bugs as direct dependencies were not compiled against those latest versions and during conflict resolution SBT by default selects the latest.

Agreed, it's very dangerous, dependencies with other dependencies are very painful to deal with and we have been hit several times in prod by binary incompatibilities.

However where we don't have the ability to update the library we depend on quickly enough, we have to make a risk assessment and decide if it's likely to explode at runtime.  Is that risk too high in all cases? Or just some?

The other options I think are to stop depending on whatever library (write our own functionality or do without) or to "vendor" the source of the library and update it ourselves which could be more expensive.

johnduffell:
my worry is that we would add all these deps, but who is ever going to remove them?

PR #1093, opened 2018-04-13
Awin affiliate tracking

rupertbates:
This is a placeholder until we get the go ahead to continue with this.

### DIFF
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -47,6 +47,6 @@
         @fragments.global.footer(product, edition, contactUsCountry)
         @fragments.footerJavaScript()
         <!--AWIN affiliate tracking tag - see here: https://drive.google.com/drive/folders/19AOGPkPbbFjqPJTdCcCba3EwpLiSHBwW -->
-         <script src="https://www.dwin1.com/{{advertiserId}}.js" type="text/javascript" defer="defer"></script>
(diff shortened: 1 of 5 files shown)


app/views/main.scala.html
```diff
@@ -46,5 +46,7 @@
 
         @fragments.global.footer(product, edition, contactUsCountry)
         @fragments.footerJavaScript()
+        <!--AWIN affiliate tracking tag - see here: https://drive.google.com/drive/folders/19AOGPkPbbFjqPJTdCcCba3EwpLiSHBwW -->
+         <script src="https://www.dwin1.com/11152.js" type="text/javascript" defer="defer"></script>
```
johnduffell: should this only be referenced on the thank you page when the variables are set?

johnduffell:
looks good, the only issue is that the JS lib appears to be loaded and parsed on every page rather than just on checkout, I wonder if that would be bad for the page weight. 👍

rupertbates:
Just for posterity and in answer to John's question about the lib being loaded on every page, this is according to the specification given by Awin. Docs are [here](https://drive.google.com/drive/folders/19AOGPkPbbFjqPJTdCcCba3EwpLiSHBwW)

johnduffell:
glad this one's finally out of limbo!! 🥇 well done
image

PR #1101, opened 2018-05-23
Some fixes for the Awin implementation

rupertbates:
We have finally got the go ahead to continue with the affiliate tracking work that was started with https://github.com/guardian/subscriptions-frontend/pull/1093

This PR adds a tracking pixel and sets the `Awin` javascript object on `window` so that it can be found by their tracking script (it was getting lost before).

Related docs are [here](https://drive.google.com/drive/folders/19AOGPkPbbFjqPJTdCcCba3EwpLiSHBwW)

### DIFF
--- a/app/views/fragments/checkout/awinPixel.scala.html
+++ b/app/views/fragments/checkout/awinPixel.scala.html
@@ -1,0 +1,16 @@
+@import com.gu.memsub.subsv2.Subscription
+@import com.gu.memsub.subsv2.SubscriptionPlan.ContentSubscription
+@import model.SubscriptionOps._
+@import views.support.PlanOps._
+@import com.gu.memsub.promo.Promotion.AnyPromotion
+
+@(subscription: Subscription[ContentSubscription], promotion: Option[AnyPromotion] = None)
+<!-- Image Pixel Tracking - Mandatory -->
+<img src="https://www.awin1.com/sread.img?tt=ns&tv=2
+&merchant=11152
+&amount=@subscription.firstPrice
+&cr=@subscription.currency
+&ref=@subscription.id.get
+&parts=@{subscription.firstPlan.segment}:@subscription.firstPrice
+&vc=@promotion.map(_.description).getOrElse("")
+&ch=aw&testmode=0" border="0" width="0" height="0">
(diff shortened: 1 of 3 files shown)


assets/javascripts/modules/analytics/awin.js
```diff
@@ -11,14 +11,15 @@ define([
         AWIN.Tracking.Sale = {};
 
         /*** Set your transaction parameters ***/
-        const productData = guardian.pageInfo.productData
+        var productData = guardian.pageInfo.productData
```
davidfurey: Why did you need to change this to a `var`?

app/views/fragments/checkout/awinPixel.scala.html
```diff
@@ -0,0 +1,16 @@
...
+@import views.support.PlanOps._
+@import com.gu.memsub.promo.Promotion.AnyPromotion
+
+@(subscription: Subscription[ContentSubscription], promotion: Option[AnyPromotion] = None)
+<!-- Image Pixel Tracking - Mandatory -->
+<img src="https://www.awin1.com/sread.img?tt=ns&tv=2
+&merchant=11152
+&amount=@subscription.firstPrice
+&cr=@subscription.currency
+&ref=@subscription.id.get
```
davidfurey: Is `id` an option, should this have a getOrElse?

assets/javascripts/modules/analytics/awin.js
svillafe: Why we can't keep `const` here?

rupertbates: I wasn't sure which js version we were targeting and Intellij was complaining. Digging a bit deeper though, it's fine so I've reinstated it.

rupertbates: I wasn't sure which js version we were targeting and Intellij was complaining. Digging a bit deeper though, it's fine so I've reinstated it.

svillafe: Ok great 👍

app/views/fragments/checkout/awinPixel.scala.html
rupertbates: It is not an option actually it just has a val called get

davidfurey: Fair enough, slightly unexpected!

PR #1102, opened 2018-05-23
Turn off logstash while we decommission the RR ELK stack

JustinPinner:
## Why are you doing this?
We're going to remove the ELK stack that we're shipping logs because of GDPR changes. This PR stops the application from attempting to communicate with the Kinesis stream that will disappear when the stack is deleted.

Later on, we'll have created a new Kinesis stream that sends fully GDPR-compliant data to a new backend. Then we can re-enable (or revert this PR) when the new stream is available and configured for this application.

## Changes
* Change `logstash.enabled` to `false` in PROD and CODE conf files 

## Screenshots
N/A

### DIFF
--- a/conf/CODE.public.conf
+++ b/conf/CODE.public.conf
@@ -34,6 +34,6 @@ google.analytics.tracking.id="UA-33592456-4"
 
 stripe.checkout.flag = true
 
-logstash.enabled=true
+logstash.enabled=false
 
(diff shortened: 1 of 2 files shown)


JustinPinner:
Tested on CODE. Nothing blew up. Subscription to digipack worked and was logged to cloudwatch. No sign of anything in Kibana. Claiming this as a ✅situation!

PR #1103, opened 2018-05-23
Use more restrictive role for ssh access

jacobwinch:
Also removes a call to an old ssh-related script from userdata


jacobwinch:
Tested on CODE.

paulbrown1982:
👍 👍 👍

jacobwinch:
It looks like there is a problem with SauceLabs (causing the tests to fail).

PR #1105, opened 2018-05-25
More awin work

rupertbates:
After more discussion with Awin it has turned out that there are a few more requirements we need to put in place, namely:

* Commission group must be one of 'Newspaper', 'Weekly' or 'Digital'
* Channel parameter must only be set to 'aw' if Awin is the referrer otherwise 'na'

### DIFF
--- a/assets/javascripts/modules/analytics/awin.js
+++ b/assets/javascripts/modules/analytics/awin.js
@@ -1,25 +1,52 @@
 //Implementation docs here: https://drive.google.com/drive/folders/19AOGPkPbbFjqPJTdCcCba3EwpLiSHBwW
 define([
     'modules/analytics/analyticsEnabled',
-    'modules/analytics/dntEnabled'
-], function (analyticsEnabled, dntEnabled) {
+    'modules/analytics/dntEnabled',
+    'utils/cookie'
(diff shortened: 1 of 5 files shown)


assets/javascripts/modules/analytics/awin.js
```diff
@@ -1,27 +1,54 @@
...
+        if (values && values[0] === 'afl' && values[1] === 'awin'){
+            return 'aw';
+        }
+        return 'na';
+    }
+
+    function storeChannel() {
+        const parsedUrl = new URL(window.location.href);
+        const utmSource = parsedUrl.searchParams.get('utm_source');
+        const utmMedium = parsedUrl.searchParams.get('utm_medium');
```
Ap0c: Does we have polyfills for [this](https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility) on IE, Edge, old Safari in this codebase?

rupertbates: I tested on IE11 & Safari 6.2 and it works

Ap0c: I guess the answer is yes then, 👍.

PR #1106, opened 2018-05-25
Update Guava to keep snyk happy

davidfurey:


build.sbt
```diff
@@ -72,7 +72,7 @@ libraryDependencies ++= Seq(
     "net.logstash.logback" % "logstash-logback-encoder" % "4.11",
     "com.fasterxml.jackson.dataformat" % "jackson-dataformat-cbor" % "2.9.2",
     "com.gu" %% "acquisition-event-producer-play26" % "3.0.0",
-    "com.google.guava" % "guava" % "23.4-jre" % "test",
+    "com.google.guava" % "guava" % "25.0-jre", //-- added explicitly - snyk report avoid logback vulnerability
```
jacobwinch: Did you mean to drop the "test" and add this comment?

davidfurey: Yes, Guava is included in the project through the play-googleauth dependency.  To include an up to date version of the library in the application, I needed to bump the version and drop the test label.

jacobwinch: Thanks for the explanation 👍

PR #1112, opened 2018-06-14
Move the AWIN javascript into the thank you page

rupertbates:
Something is preventing the AWIN tracking code from firing properly when it is included via main.js.
It feels like a race condition because apparently it has worked on occasion. 

To try to work round this I'm including the code directly in the thank you page, if this works we can work out what the best route forward is from here...

### DIFF
--- a/app/views/checkout/thankyou.scala.html
+++ b/app/views/checkout/thankyou.scala.html
@@ -36,6 +36,47 @@
         currency: '@subscription.currency',
         subscriptionId: '@subscription.id.get'
     };
+
+    function getCookie(name) {
+        var nameEQ = name + '=';
+        var ca = document.cookie.split(';');
+        for (var i = 0; i < ca.length; i++) {
+            var c = ca[i];
+            while (c.charAt(0) === ' ') { c = c.substring(1, c.length); }
+            if (c.indexOf(nameEQ) === 0) { return c.substring(nameEQ.length, c.length); }
+        }
+        return null;
+    }
+
+    function getAwinChannel() {
+        const channel = getCookie('gu_referrer_channel');
+        if (channel === null) {
+            return 'direct'; //No referrer the user came direct to the site
+        }
+
+        const values = channel ? channel.split('&') : null;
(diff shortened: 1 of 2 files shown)


app/views/checkout/thankyou.scala.html
```diff
@@ -36,6 +36,47 @@
...
+        return 'na'; //A referrer other than Awin
+    }
+
+    //<![CDATA[ /*** Do not change ***/
+    var AWIN = {};
+    AWIN.Tracking = {};
+    AWIN.Tracking.Sale = {};
+
+    /*** Set your transaction parameters ***/
+    const productData = guardian.pageInfo.productData;
```
Ap0c: Does subs have a list of supported browsers? Using `const` here will cause issues with, e.g. IE10. Is it worth using `var` here as with the `AWIN` variable just in case? I think the `const` is globally scoped anyway because it's sitting at the top level of a script tag.

app/views/checkout/thankyou.scala.html
```diff
@@ -36,6 +36,47 @@
...
+    AWIN.Tracking.Sale.amount = productData.amount;
+    AWIN.Tracking.Sale.orderRef = productData.subscriptionId;
+    AWIN.Tracking.Sale.parts = productData.productSegment + ':' + productData.amount;
+    AWIN.Tracking.Sale.voucher = productData.promoCode;
+    AWIN.Tracking.Sale.currency = productData.currency;
+    AWIN.Tracking.Sale.test = '0';
+    AWIN.Tracking.Sale.channel = getAwinChannel();
+    //]]>
```
Ap0c: Is it worth wrapping this in the classic `(function() {})()` and explicitly putting `AWIN` on the `window` as before? That way `getCookie`, `getAwinChannel` and `productData` won't bleed out into the global scope.

rupertbates: I want to keep this code exactly as provided for the purposes of this test. I'm planning to refactor after we get it working.

PR #1114, opened 2018-06-15
Remove the 'Exclusive Benefits' section from the promo landing page

rupertbates:
Apparently the Exclusive benefits section of the [Digital Pack promotion landing page](https://subscribe.theguardian.com/p/DPS82X) is no longer accurate so we are removing it completely.

#### The section that has been removed:
screen shot 2018-06-15 at 11 05 54


AWare:
:+1:

PR #1120, opened 2018-07-09
Redirect Weekly Landing To Annual Promo

Ap0c:
# Why?

We want to run an annual promotion by default for Guardian Weekly, which is handled by the promo page. This redirects requests to the Weekly landing page to the promo page with the Annual promotion applied. Other promotions for Guardian Weekly running during this time will still work using their own promo code.

When this promotion ends, we can revert this PR to restore the Weekly Landing page as it currently is.

[**Trello Card**](https://trello.com/c/id118pmD/1746-changes-to-the-gw-subscription-options)

cc @JustinPinner 

# Changes

- Added temporary redirect from the Weekly landing page.


johnduffell:
yep looks good, presumably you copied the general pattern from elsewhere so it's pretty standard 👍

PR #1124, opened 2018-07-11
Hide Quarterly Weekly Pricing

Ap0c:
## Why?

Having quarterly pricing on the region selector for Guardian Weekly makes it harder to tell that there's an annual option. This removes it, so that you see the price when you click through, and get quarterly and annual side-by-side.

cc @JustinPinner 

## Changes

- Removed template code showing pricing on the region buttons.

### Before

weekly-before-region
weekly-before-price

### After

weekly-after-region
weekly-after-price


johnduffell:
that makes a lot more sense, given the quesiton is "where do you want it sent to" I doubt people would move country just to get a cheaper rate! 👍

PR #1127, opened 2018-07-24
Add install tracking param to app store links

rupertbates:
Add install referrer params to the app store links to enable us to attribute installs to this site.


JustinPinner:
Late to the party, but 👍

PR #1129, opened 2018-08-07
Prevent search engines indexing parameterised pages

JustinPinner:
Sometimes we use INTCMP parameters on our subscription index pages. If these are crawled and indexed by search engines, that can cause problems for us and for people searching for subscription-related pages.

This PR extends the use of the existing `Hreflangs` case classes so that we can include a `<link rel="canonical" href="the_canonical_link_here">` in the `<head>` section of those pages.

It also adds a few extra entries to `robots.txt` to prevent indexing of URLs that don't have a valid context as a direct hit from search results.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,6 +1,11 @@
 User-agent: *
 
-Disallow: /patterns
+Disallow: /checkout
+Disallow: /healthcheck
 Disallow: /manage
+Disallow: /management/manifest
+Disallow: /p/*
+Disallow: /patterns
+Disallow: /test-users
 
 Allow: /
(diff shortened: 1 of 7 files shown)


public/robots.txt
```diff
@@ -1,6 +1,11 @@
 User-agent: *
 
-Disallow: /patterns
+Disallow: /checkout
+Disallow: /healthcheck
 Disallow: /manage
+Disallow: /management/manifest
+Disallow: /p/*
```
JustinPinner: I don't _think_ it makes sense to index `/p/*` pages, as they're generally short-lived if they're only used for time-limited promotions. Unless anyone disagrees?

paulbrown1982: 👍 but never-say-never :)

PR #1130, opened 2018-08-09
Change the weekly landing page redirects to retain any query params

rupertbates:
Currently if you go to `https://subscribe.theguardian.com/weekly?foo=bar` you get redirected to `https://subscribe.theguardian.com/p/10ANNUAL?country=GB` and the `foo` query param is lost.

This PR changes the behaviour so that you now end up on `https://subscribe.theguardian.com/p/10ANNUAL?country=GB&foo=bar`

### DIFF
--- a/app/controllers/WeeklyLandingPage.scala
+++ b/app/controllers/WeeklyLandingPage.scala
@@ -28,7 +28,7 @@ class WeeklyLandingPage(tpBackend: TouchpointBackend, commonActions: CommonActio
 
   def index(country: Option[Country]) = NoCacheAction { implicit request =>
     val maybeCountry = country orElse request.getFastlyCountry
-    Redirect(routes.WeeklyLandingPage.withCountry(maybeCountry.map(_.alpha2).getOrElse(international)))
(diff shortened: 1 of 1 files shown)


app/controllers/WeeklyLandingPage.scala
```diff
@@ -28,15 +28,15 @@ class WeeklyLandingPage(tpBackend: TouchpointBackend, commonActions: CommonActio
 
   def index(country: Option[Country]) = NoCacheAction { implicit request =>
     val maybeCountry = country orElse request.getFastlyCountry
-    Redirect(routes.WeeklyLandingPage.withCountry(maybeCountry.map(_.alpha2).getOrElse(international)))
+    Redirect(routes.WeeklyLandingPage.withCountry(maybeCountry.map(_.alpha2).getOrElse(international)).url, request.queryString, TEMPORARY_REDIRECT)
```
Ap0c: Should this be a "See Other"?

rupertbates: After discussion we are going to leave this as 'TEMPORARY_REDIRECT'

PR #1133, opened 2018-08-17
Custom intcmp values that are valid during a flash sale

lmath:
During a flash sale, it is useful to be able to set a custom value for the internal campaign code to more easily track the promotion in GA. 

This adds that capability. 

This does mean that there is a map of intcmp values that are valid during the flash sale that needs to be manually updated whenever we set up another flash sale.

### DIFF
--- a/app/services/FlashSale.scala
+++ b/app/services/FlashSale.scala
@@ -10,12 +10,12 @@ object FlashSale extends FlashSale {
   val startTime = new DateTime(2018, 8, 20, 0, 0)
   val endTime = new DateTime(2018, 9, 3, 0, 0)
 
-  val flashSaleIntcmp: Map[String, String] =
+  val flashSaleIntcmp: Map[String, Intcmp] =
     Map(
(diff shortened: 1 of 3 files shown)


app/services/FlashSale.scala
```diff
@@ -2,26 +2,46 @@ package services
...
+}
+
+trait FlashSale {
 
-    val startTime = new DateTime(2018, 8, 20, 0, 0)
-    val endTime = new DateTime(2018, 9, 3, 0, 0)
+  val startTime: DateTime
+  val endTime: DateTime
+
+  val flashSaleIntcmp: Map[String, Intcmp]
```
jacobwinch: Minor: it might be easier to understand this if it was a Map[[PromoCode](https://github.com/guardian/membership-common/blob/91523b8ce30603a3b1072e65456a6d237b78b5bf/src/main/scala/com/gu/memsub/promo/Promotion.scala#L18-L20), Intcmp]. To be fair, the example and test code already make this quite simple to understand...

PR #1134, opened 2018-08-20
Make it easier to debug failures to load the product catalog

jacobwinch:
https://trello.com/c/7m3LobBn/26-retro-action-improve-logging-when-we-fail-to-load-the-catalog


johnduffell:
thanks for doing that! 👍

PR #1137, opened 2018-08-29
[DO NOT MERGE] Reader Revenue Hackday 28-29 August 2018

paulbrown1982:
This hack hijacks most of the hyperlinks on the checkout page and displays a popup giving a promo code if the user clicks on a link having gone as far as giving their first name.

The hijacked links are highlighted in yellow for the purpose of the demo, they wouldn't be if such a feature was put live to real customers.

The promo code HACKDAY10 will not work on any other product than Digital Pack for the moment. I'll need to create a different code for the other products/pages.

picture 494

### DIFF
--- a/app/views/checkout/payment.scala.html
+++ b/app/views/checkout/payment.scala.html
@@ -244,6 +244,86 @@
         </form>
 
     </section>
-
 </main>
+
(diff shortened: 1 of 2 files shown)


jacobwinch:
Just tidying up old PRs, so I'm going to close this.

PR #1138, opened 2018-09-06
send Identity id and salesforce account id on enqueued emails

QuarpT:
Required for Braze integration.
If the identity Id is not found, the userId defaults to the sfContactId
sfContactId is set as a custom attribute in Braze
UserId is used when upserting a user to Braze before sending the email

### DIFF
--- a/app/services/ExactTargetService.scala
+++ b/app/services/ExactTargetService.scala
@@ -151,7 +151,7 @@ class ExactTargetService(
 
     for {
       row <- buildWelcomeEmailDataExtensionRow(subscribeResult, subscriptionData, gracePeriod, validPromotion, purchaserIds)
-      response <- SqsClient.sendDataExtensionToQueue(Config.welcomeEmailQueue, row)
+      response <- SqsClient.sendDataExtensionToQueue(Config.welcomeEmailQueue, row, purchaserIds.identityIdWithFallback, purchaserIds.contactId.salesforceContactId)
     } yield {
       response match {
         case Success(sendMsgResult) => logger.info(s"Successfully enqueued ${subscribeResult.subscriptionName} welcome email for user ${purchaserIds.identityId}.")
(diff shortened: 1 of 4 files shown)


app/services/ExactTargetService.scala
```diff
@@ -168,32 +168,39 @@ class ExactTargetService(
...
+        numberOfSuspensionsLinedUp = numberOfSuspensionsLinedUp,
+        daysUsed = daysUsed,
+        daysAllowed = daysAllowed
+      )
+    }
 
-    buildDataExtensionRow.flatMap { row =>
-      EitherT(SqsClient.sendDataExtensionToQueue(Config.holidaySuspensionEmailQueue, row).map {
+    def sendToQueue(salesforceContact: Contact, row: HolidaySuspensionBillingScheduleDataExtensionRow) = {
+      SqsClient.sendDataExtensionToQueue(Config.holidaySuspensionEmailQueue, row, salesforceContact.identityId, salesforceContact.salesforceContactId).map {
```
paulbrown1982: We might need to double check that the Identity ID has actually been added to the Salesforce contact at this point.

app/services/ExactTargetService.scala
```diff
@@ -261,8 +270,10 @@ object SqsClient extends LazyLogging {
             "SubscriberAttributes" ->  Json.toJsFieldJsValueWrapper(row.fields.toMap)
           )
         ),
-        "DataExtensionName" -> row.forExtension.name
-      ).toString()
+        "DataExtensionName" -> row.forExtension.name,
+        "UserId" -> userId,
+        "SfContactId" -> sfContactId
```
paulbrown1982: Having the SfContactId in the data might actually help to speed up the Contact lookup in the workflow service. I think it was searching for one by Identity ID. I'll set a reminder to have a look into that.

paulbrown1982: Actually, you could pick up the SF Contact ID in the workflow service, and not need to change subs frontend?

app/services/ExactTargetService.scala
```diff
@@ -168,32 +168,39 @@ class ExactTargetService(
...
+      SqsClient.sendDataExtensionToQueue(Config.holidaySuspensionEmailQueue, row, salesforceContact.identityId, salesforceContact.salesforceContactId).map {
         case Success(sendMsgResult) => \/.right(())
         case Failure(e) => \/.left(s"Details were: ${row.subscriptionName}, ${e.toString}")
-      })
-    }.run.map {
+      }
+    }
+
+    val enqueueResult = for {
+      salesforceContact <- EitherT.right(GetSalesforceContactForSub(subscription)(zuoraService, salesforceService.repo, executionContext))
```
paulbrown1982: We need to check that this doesn't require an Identity ID anywhere in the lookup, as holiday suspension emails can go to anyone.

QuarpT: You mean for the GetSalesforceContactForSub call? This functionality shouldn't have changed from before

QuarpT:
@paulbrown1982 We were making the `GetSalesforceContactForSub` call in the old code

PR #1142, opened 2018-09-11
Stop publishing identity id for email send

QuarpT:
Stop publishing identity id for email send. It is now retrieved in membership-workflow.
Identity Id is used when setting the external_id in Braze. see https://github.com/guardian/membership-workflow/pull/143


johnduffell:
thanks, looks good to me! 👍

johnduffell:
althouh I should mention for posterity that the welcome email will only be guaranteed to work correctly because we fill in the identity id in SF before sending the email: https://github.com/guardian/subscriptions-frontend/blob/e057a5b8dfd8ec94c48ebae4455f0e1fcb6aed5c/app/services/CheckoutService.scala#L178-L180

QuarpT:
thanks @johnduffell 
I'll need a tick when you're ready for me to merge 😎

johnduffell:
sorry didn't think on that some of the projects block merging before a green tick!

PR #1147, opened 2018-09-24
[WIP] Get consistency for Customer Acceptance Date

paulbrown1982:
When creating a Six for Six offer, instead of triggering the Six Issues charge on the first paper date and delaying the Customer Acceptance Date to 6 weeks later in order to trigger the quarterly plan as per its catalog definition; I have changed the logic so that we set the Customer Acceptance date as the first paper date (as is normal for all other offer types), as well as triggering the Six Issues plan on that date, and delay the quarterly plan to being triggered on a Specific Date 6 weeks after that.

It is the TODO in this doc: https://docs.google.com/spreadsheets/d/1q3k8JTUtbEhho-tgwdkQHFQDYxqBK5j8WerY55u_EQA/

The main benefit is that we no longer have to modify the Customer Acceptance Date on line 101 - it means what it means for all other offers: the date of first paper and the date of first payment.

I'll double check the fulfilment system is compatible with this change, and will merge if all my manual billing tests also pass.

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -88,16 +88,19 @@ class CheckoutService(
     allSixWeekAssociations.map { allSixWeekAssociations =>
       val maybeSixWeekAssociation = allSixWeekAssociations.find(_._1.id.get == originalCommand.ratePlans.head.productRatePlanId)
       val updatedCommand = maybeSixWeekAssociation.map { case (sixWeekPlan, recurringPlan) =>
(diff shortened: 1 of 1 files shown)


jacobwinch:
> It would then also select the 6-for-6 in the introductoryPeriodQuery, and so the customer would get 2 papers for 6 weeks :(

That's a shame :( - should we close this until we have had chance to re-work the fulfilment script?

PR #1148, opened 2018-09-25
Fix slow startup when running locally

jacobwinch:
This http request will not succeed locally, since it's EC2 specific.

See also: https://github.com/guardian/membership-frontend/pull/1426


jacobwinch:
> Testing in PROD failed!

This is broken because Travis is still configured to run these tests in SauceLabs, this should be fixed by https://github.com/guardian/subscriptions-frontend/pull/1149 (+ some config changes that I've made in Travis).

PR #1151, opened 2018-09-27
Use the Ophan browser id as the GA user id if available

rupertbates:
If we have an ophanBrowserId then set it as the [GA User Id](https://support.google.com/analytics/answer/3123663), this will enable us to join up client side and server side tracking as well as cross domain sessions.


assets/javascripts/modules/analytics/ga.js
```diff
@@ -47,6 +47,9 @@ define(['modules/analytics/analyticsEnabled',
         ga('membershipPropertyTracker.set', 'dimension2', identitySignedOut.toString());  // deprecated - is logically equivalent to: dimension6 != "" and dimension7 === "false"
         (guardian.ophan) && ga('membershipPropertyTracker.set', 'dimension3', guardian.ophan.pageViewId); // ophanPageview Id
         (ophanBrowserId) && ga('membershipPropertyTracker.set', 'dimension4', ophanBrowserId); // ophanBrowserId
+        // If we have an ophanBrowserId then set it as the GA User Id, this will enable us to join up client side and server side tracking.
+        // See: https://support.google.com/analytics/answer/3123663
+        (ophanBrowserId) && ga('set', 'userId', ophanBrowserId);
```
svillafe: I personally find more difficult to read this kind of expression (we are using the falsely and truly values of javascript and is less explicit), but I understand you are using the style of the current file.

PR #1152, opened 2018-09-27
Set the correct Scala version in .travis.yml

rupertbates:
Acceptance tests are failing at the moment due to Travis using the wrong version of Scala


johnduffell:
thanks for doing this! 👍

PR #1154, opened 2018-09-28
Be able to sell new GW plans 

lmath:
## Why are you doing this?

We are adding new plans for Guardian weekly. These new plans were added to the catalog here: https://github.com/guardian/membership-common/pull/583#event-1873624786

There is a boolean, [showUpdatedPrices](https://github.com/guardian/subscriptions-frontend/compare/lm-gw-price-rise?expand=1#diff-f5ccd2106078cafb58e6ec7887268e56R167), that will remain set to false during the merge. This PR adds the functionality to be able to sell the new plans if the switch were on (the customer would see the new prices and go to a checkout for a new plan). 

This change also concerns a move from Zone A (UK/US) and Zone C (everything else) to Domestic Zone (UK/US/CA/AU/NZ/EU) and Rest of World Zone (everything not in the domestic zone). 

After this merge even with showUpdatedPrice set to false, we would be able to see one of the new plans if someone happened to have the checkout url.

[**Trello Card**](https://trello.com/c/30NatCvc/1915-wire-in-new-rate-plans-into-subs-site)

## Changes

* Bumps membership common to get the new plans from the catalog
* Adds definition of domestic/restOfWorld plans 
* Adds a PlanPicker which switches between the zoneA/C world and the domestic/rest of world world


app/views/support/WeeklyPromotion.scala
```diff
@@ -7,63 +7,57 @@ import com.gu.memsub.promo.{PromoCode, WeeklyLandingPage}
...
 
-  case class DiscountedRegion(title: String, description: String, countries: Set[Country], discountedPlans: List[DiscountedPlan])
+  case class DiscountedRegion(title: String, description: String, discountedPlans: List[DiscountedPlan])
 
   def validRegionsForPromotion(promotion: Option[PromoWithWeeklyLandingPage], promoCode: Option[PromoCode], requestCountry: Country)(implicit catalog: SubsCatalog): Seq[DiscountedRegion] = {
     val promotionCountries = promotion.map(_.appliesTo.countries).getOrElse(allCountries)
-    val regionForZoneCCountry: Seq[DiscountedRegion] = {
-      if (ZONEC.contains(requestCountry)) {
+
+    val separateRegionIfRequestCountryIsRestOfWorld: Seq[DiscountedRegion] = {
```
jacobwinch: I would call this 'promotedRegion' and add a comment with something like:

// If a user does not qualify for domestic delivery (e.g. if the user is based in South Africa), we want to explicitly call out their (likely) delivery country on the landing page

Or preferably a simpler way to phrase that ^^^ 🙈

app/views/support/WeeklyPromotion.scala
```diff
@@ -7,63 +7,57 @@ import com.gu.memsub.promo.{PromoCode, WeeklyLandingPage}
...
     }
-    val rowWithoutCountry: Seq[DiscountedRegion] = Seq(DiscountedRegion(
-        title = "Rest of the world",
-        description = "Posted to you by air mail",
-        countries = ZONEC - requestCountry,
-        discountedPlans = plansForPromotion(promotion, promoCode, ZONEC - requestCountry, catalog.weekly.zoneC.plans,Currency.USD)
-      ))
+
+    val restOfWorldRegion = Seq(DiscountedRegion(
+      title = "Rest of the world",
```
jacobwinch: One for a follow up PR, but I wonder if 'Rest of the world' is a user-friendly label for this?

app/views/support/WeeklyPromotion.scala
lmath: Yeah, good shout. Names are hard. 😬

PR #1156, opened 2018-10-01
Log errors on start up

jacobwinch:
Copied from: https://github.com/guardian/support-frontend/pull/945


jacobwinch:
^^^ Prout can't see the new build, because we've just set up a re-directs for https://subscribe.theguardian.com/

PR #1158, opened 2018-10-02
Fix acceptance tests

jacobwinch:
The test user cookie is no longer being set correctly due to the redirect referenced here: https://github.com/guardian/subscriptions-frontend/pull/1157.

This redirect means that we are attempting to set the cookies whilst on support.theguardian.com instead of subscribe.theguardian.com, which doesn't work:

> you need to be on the domain that the cookie will be valid for.

https://docs.seleniumhq.org/docs/03_webdriver.jsp#cookies

### DIFF
--- a/test/acceptance/util/Config.scala
+++ b/test/acceptance/util/Config.scala
@@ -25,7 +25,6 @@ object Config {
     logger.info(s"Stage: ${conf.getString("stage")}")
     logger.info(s"Subscriptions Frontend: ${baseUrl}")
     logger.info(s"Identity Frontend: ${identityFrontendUrl}")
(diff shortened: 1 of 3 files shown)


test/acceptance/util/Config.scala
```diff
@@ -25,7 +25,6 @@ object Config {
     logger.info(s"Stage: ${conf.getString("stage")}")
     logger.info(s"Subscriptions Frontend: ${baseUrl}")
     logger.info(s"Identity Frontend: ${identityFrontendUrl}")
-    logger.info(s"Screencast = https://saucelabs.com/tests/${Driver.sessionId}")
```
jacobwinch: This is just some tidy up that I missed when moving from SauceLabs to BrowserStack.

PR #1160, opened 2018-10-02
Fix 6 for 6 for new Guardian Weekly plans

jacobwinch:
https://trello.com/c/PtuBfAFZ/1964-fix-6-for-6-checkout-for-new-plans


jacobwinch:
^^^ There are problems with TeamCity - I'm having a look.

PR #1162, opened 2018-10-02
Add switch and override for GW new prices

JustinPinner:
Add a timed switch for GW launch, and a temporary parameter for testing in the meantime. The parameter can be placed anywhere in the query string.

We'll be pulling out the parameter and generally tidying up soon after the update takes effect.

Also, the canonical and alternate URLs ignore the new parameter, so this shouldn't have any effect on search indexing as long as crawlers all honour the canonical entry equally.

### DIFF
--- a/app/views/support/WeeklyPromotion.scala
+++ b/app/views/support/WeeklyPromotion.scala
@@ -174,7 +174,7 @@ object PlanPicker {
 
   def showUpdatedPrices = {
     val now = DateTime.now().withZone(DateTimeZone.UTC)
-    val threshold = DateTime.parse("2018-10-10T08:00:00").withZone(DateTimeZone.UTC)
+    val threshold = DateTime.parse("2018-10-10T09:45:00").withZone(DateTimeZone.UTC)
     now.isAfter(threshold)
   }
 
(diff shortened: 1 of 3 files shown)


app/views/support/WeeklyPromotion.scala
```diff
@@ -167,20 +171,25 @@ object WeeklyPromotion {
 object PlanPicker {
 
   import model.GuardianWeeklyZones._
-  val showUpdatedPrices = false
 
-  def isInRestOfWorldOrZoneC(country: Country, showUpdatedPrices: Boolean = showUpdatedPrices): Boolean = {
-    if(showUpdatedPrices) GuardianWeeklyZones.restOfWorldZoneCountries.contains(country)
+  def showUpdatedPrices = {
+    val now = DateTime.now().withZone(DateTimeZone.UTC)
+    val threshold = DateTime.parse("2018-10-10T08:00:00").withZone(DateTimeZone.UTC)
```
rupertbates: Is this time right? I thought it was 10.30 now?

JustinPinner: Ah, last I heard it was 09:00. I'll double-check, and I suspect it'll change more than once anyway :)

jacobwinch: Currently it's scheduled for: Wednesday 10th October, 10:45 BST

It changed at least twice whilst you were on holiday so don't be too surprised if it changes again 🙃

PR #1163, opened 2018-10-02
Show GBP currency override when shipping to ROW

lmath:
Currently, the checkout for a **zone C** plan (any countries other than UK/US) provides a currency override to pay in pounds like so:

image

This change will make it so that after we switch to the new plans, that this currency override is available when in the checkout for a **rest of world** plan.


JustinPinner:
Nice!

PR #1164, opened 2018-10-04
Improve unit tests for GW plan selection

jacobwinch:
This PR refactors the GW product selection logic to make it easier to unit test (without creating tons of mocks).


app/views/support/WeeklyPromotion.scala
```diff
@@ -168,45 +181,46 @@ object WeeklyPromotion {
...
 
-object PlanPicker {
+object WeeklyPicker {
 
   import model.GuardianWeeklyZones._
 
-  def showUpdatedPrices = {
+  val updateTime = DateTime.parse("2018-10-10T09:45:00").withZone(DateTimeZone.UTC)
+
+  def showUpdatedPrices(forceShowUpdatedPrices: Boolean, timeToUpdate: DateTime = updateTime): Boolean = {
```
jacobwinch: All of the logic for determining whether to display the new prices now happens in showUpdatedPrices, which makes it easier to test each function in isolation.

app/model/GuardianWeeklyZones.scala
```diff
@@ -25,3 +26,20 @@ object GuardianWeeklyZones {
   val domesticZoneCountries = domesticZoneCountryGroups.flatMap(_.countries)
   val restOfWorldZoneCountries = allCountries.diff(domesticZoneCountries)
 }
+
+object PurchasableWeeklyProducts {
+
+  sealed trait PurchasableWeeklyProduct
```
jacobwinch: This allows us to test that the right product is returned before/after switchover for various regions, without worrying about all of the plans in the catalog

jacobwinch:
^^^ We're having more problems with TeamCity at the moment.

jacobwinch:
I'm going to revert this because it's not going to make it into production today.

PR #1165, opened 2018-10-04
Update the default promo landing page (behind switch)

jacobwinch:
Currently hitting https://subscribe.theguardian.com/weekly redirects to a promo landing page which offers a 10% discount (on the annual plan).

This PR allows us to automatically start redirecting traffic to a new 6 for 6 promo landing page (once the new products are live). This will happen based on the showUpdatedPrices switch status, which we also use to determine whether to show new products or not.

### DIFF
--- a/app/services/WeeklyPicker.scala
+++ b/app/services/WeeklyPicker.scala
@@ -1,0 +1,53 @@
+package services
+
+import com.gu.i18n.{Country, CountryGroup}
+import model.GuardianWeeklyZones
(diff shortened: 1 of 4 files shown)


app/services/WeeklyPicker.scala
```diff
@@ -0,0 +1,53 @@
+package services
```
jacobwinch: This has all just been copied and pasted into a new file.

PR #1172, opened 2018-10-10
Fix display of regions in AU, NZ and CA

jacobwinch:
The current logic means that we hide Australia from Australians, New Zealand from New Zealanders and Canada from Canadians 😱


lmath:
👍

PR #1177, opened 2018-10-12
Set the awin cookie on .theguardian.com not subscribe.theguardian.com

rupertbates:
Now that we are sending affiliate traffic via support.theguardian.com and setting the `gu_referrer_channel` cookie there we need to set it on `.theguardian.com` rather than `subscribe.theguardian.com` so that we share the cookie between the two sites.

### DIFF
--- a/assets/javascripts/utils/cookie.js
+++ b/assets/javascripts/utils/cookie.js
@@ -1,7 +1,7 @@
 define(['utils/base64'], function (base64) {
     'use strict';
 
-    function setCookie(name, value, days, isUnSecure) {
+    function setCookie(name, value, days, isUnSecure, trimSubdomain) {
         var date;
         var expires;
         // used for testing purposes, cookies are secure by default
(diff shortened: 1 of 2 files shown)


assets/javascripts/utils/cookie.js
```diff
@@ -15,7 +15,15 @@ define(['utils/base64'], function (base64) {
             expires = '';
         }
 
-        document.cookie = [name, '=', value, expires, '; path=/', secureCookieString ].join('');
+        var domain = trimSubdomain ? '; domain=' + getShortDomain() : '';
```
akash1810: Can this be a `const`?` Can also use template strings for bonus points!

rupertbates: it will be the only const in the whole codebase so I just tend to match the existing style. Same with template strings. 
It's tricky, I guess we could just start converting as we go but then PRs become a mish mash of actual functional changes and stylistic changes.

akash1810: meep https://github.com/guardian/subscriptions-frontend/pull/1177/files#diff-38cbb81a25ab7304ca9c2b69638a79d0R25

PR #1179, opened 2018-10-15
Tidy up after Guardian Weekly relaunch

jacobwinch:
Now that we sell new Guardian Weekly plans by default, we can remove lots of code which was built to deal with switching between old and new plans / images.

I'd recommend reviewing on a commit by commit basis, as this will make it easier to understand the motivation/links between each change.

There's still more that can be done to simplify this code, but this seemed like a sensible amount of code to rip out in one PR!


app/model/GuardianWeeklyZones.scala
```diff
@@ -7,7 +7,6 @@ object GuardianWeeklyZones {
   private val allCountries = CountryGroup.allGroups.flatMap(_.countries).toSet
 
   val zoneACountryGroups = Set(CountryGroup.UK, CountryGroup.US)
```
jacobwinch: This is still being used in a couple of places, and I didn't want to rip out too much code at once!

app/forms/SubscriptionsForm.scala
```diff
@@ -31,9 +31,6 @@ class SubscriptionsForm(catalog: Catalog) {
     val validPlans =
       catalog.delivery.list.toList ++
       catalog.voucher.list.toList ++
-      catalog.weekly.zoneA.plans.filter(_.availableForCheckout) ++
```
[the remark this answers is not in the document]
lmath: nice 👌

app/views/support/PlanOps.scala
```diff
@@ -59,7 +58,11 @@ object PlanOps {
           availableImages = Seq(ResponsiveImage(controllers.CachedAssets.hashedPathFor("images/digital-pack-garnett.png"), 300)),
             altText = Some("Guardian apps demoed on Apple and Android devices")
         )
-      case Weekly :: Nil => ImagePicker.defaultPackshotImage(rawQueryString)
+      case Weekly :: Nil =>
```
lmath: We no longer need the rawQueryString now right?

app/services/WeeklyPicker.scala
```diff
@@ -10,83 +10,16 @@ object WeeklyPicker {
```
lmath: Nitpicking, but this tidying also makes it possible to tidy up imports here e.g. `import com.gu.memsub.images.{ResponsiveImageGenerator, ResponsiveImageGroup}`

PR #1180, opened 2018-10-16
Put the request country at the top of the list of regions for Guardian Weekly

lmath:
Before we switched to use new plans after the Guardian Weekly relaunch, https://subscribe.theguardian.com/p/WWM99X?country=AU would shown Australia to be the first region in the list (and same for other countries that were previously zone C).

So, this makes it so that the request country (either passed as a query param, or else if we have country info from Fastly, or else the UK is default see [here](https://github.com/guardian/subscriptions-frontend/blob/master/app/controllers/PromoLandingPage.scala#L145)) is at the top of the list. 

We always show:
* United Kingdom
* United States
* Europe
* Australia
* New Zealand
* Canada

So, if the request country is any of these, then it is still the prioritised country but we make sure we don't display it twice: 
https://github.com/guardian/subscriptions-frontend/compare/lm-request-country-is-top?expand=1#diff-f5ccd2106078cafb58e6ec7887268e56R103 

This might be a convoluted way to go about it, so alternative approaches definitely welcome. 

e.g. before:
image

e.g. after:
image

### DIFF
--- a/app/model/GuardianWeeklyZones.scala
+++ b/app/model/GuardianWeeklyZones.scala
@@ -1,6 +1,6 @@
 package model
 
-import com.gu.i18n.CountryGroup
+import com.gu.i18n.{Country, CountryGroup}
 import com.gu.memsub.subsv2.{CatalogPlan, WeeklyPlans}
 
 object GuardianWeeklyZones {
@@ -8,19 +8,21 @@ object GuardianWeeklyZones {
 
   val zoneACountryGroups = Set(CountryGroup.UK, CountryGroup.US)
 
-  val domesticZoneCountryGroups = Set(
+  val domesticZoneCountryGroups = Seq(
     CountryGroup.UK,
     CountryGroup.US,
+    CountryGroup.Europe,
     CountryGroup.Australia,
-    CountryGroup.Canada,
-    CountryGroup.US,
     CountryGroup.NewZealand,
-    CountryGroup.Europe
+    CountryGroup.Canada
   )
-  val restOfWorldZoneCountryGroups = CountryGroup.allGroups.toSet.diff(domesticZoneCountryGroups)
+  val restOfWorldZoneCountryGroups = CountryGroup.allGroups.toSet.diff(domesticZoneCountryGroups.toSet)
 
   val domesticZoneCountries = domesticZoneCountryGroups.flatMap(_.countries)
-  val restOfWorldZoneCountries = allCountries.diff(domesticZoneCountries)
+  val restOfWorldZoneCountries = allCountries.diff(domesticZoneCountries.toSet)
+
+  def getDomesticCountryGroup(country: Country): Option[CountryGroup] =
+    domesticZoneCountryGroups.find(countryGroup => countryGroup.countries.contains(country))
 }
 
 object PurchasableWeeklyProducts {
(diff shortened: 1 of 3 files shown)


app/views/support/WeeklyPromotion.scala
```diff
@@ -72,43 +51,109 @@ object WeeklyPromotion {
...
+      CountryGroup.UK -> UKregion,
+      CountryGroup.US -> USregion,
+      CountryGroup.Europe -> EUregion,
+      CountryGroup.Australia -> AUSregion,
+      CountryGroup.NewZealand -> NZregion,
+      CountryGroup.Canada -> CAregion
+    )
+  }
+
+  //If the request country belongs to a domestic country group, we still still display that country at the top of the list
```
jacobwinch:
```suggestion
  //If the request country belongs to a domestic country group, we display that country at the top of the list
```

test/views/support/WeeklyPromotionTest.scala
```diff
@@ -0,0 +1,39 @@
+package views.support
+
+import com.gu.i18n.{Country, CountryGroup}
+import model.GuardianWeeklyZones
+import org.specs2.mutable.Specification
+
+class WeeklyPromotionTest extends Specification {
```
jacobwinch: 🏅

app/views/support/WeeklyPromotion.scala
```diff
@@ -72,43 +51,109 @@ object WeeklyPromotion {
...
+  //However, if it corresponds to a region that is always displayed then we need to ensure we don't show it twice
+  //e.g. if the request country is the US, then it is the prioritised country and we need to not show it again in the
+  //list of domestic countries. A rest of world country should always be at the top, and then all the domestic country groups after.
+  def domesticCountryGroupsToDisplay(requestCountry: Country): List[CountryGroup] = {
```
jacobwinch: One possible alternative to this de-duping would be to add a _prioritised_ field to the DiscountedRegion case class.

We'd then need to define a custom ordering based on this value, but once we had that it would allow us to sort the full list of discounted regions.

jacobwinch: Actually we could probably just use sortBy rather than bothering with an ordering...

lmath: Hmm, good idea. Let me give that a try.

app/model/GuardianWeeklyZones.scala
```diff
@@ -1,26 +1,28 @@
...
+import com.gu.i18n.{Country, CountryGroup}
 import com.gu.memsub.subsv2.{CatalogPlan, WeeklyPlans}
 
 object GuardianWeeklyZones {
   private val allCountries = CountryGroup.allGroups.flatMap(_.countries).toSet
 
   val zoneACountryGroups = Set(CountryGroup.UK, CountryGroup.US)
 
-  val domesticZoneCountryGroups = Set(
+  val domesticZoneCountryGroups = Seq(
```
LATaylor-guardian: Are each of the values here unique? I.e no duplicates and expected to never be no duplicates? If so, this should remain a `Set` which has these properties.

lmath: It's because there needs to be a deliberate order somewhere in the code. Yes, perhaps having that here is needlessly opaque -- any ideas where it would be good for the ordering to live more explicitly?

LATaylor-guardian: I understand. This is fine then.

PR #1181, opened 2018-10-17
Upgrade acquisition-event-producer library

rupertbates:
Upgrade to the latest version of the acquisition-event-producer library to fix the link between client-side and server-side tracking.


rupertbates:
😄

PR #1183, opened 2018-10-17
override jackson-databind to 2.9.7

lmath:
https://snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-31507

We are pulling in old versions of jackson-databind primarily via aws-sdks. It's suggested [here](https://github.com/aws/aws-sdk-java) that overriding jackson-databind shouldn't be a breaking change. Indeed, when running the project locally after the version bump, that seems to be the case. 

We are also pulling in an old version of jackson-databind to support sign in.

I bumped the version, signed in with a test user, then bought a digital pack locally and that worked fine.

Any other suggested checks?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -48,7 +48,7 @@ libraryDependencies ++= Seq(
     jodaForms,
     PlayImport.specs2 % "test",
     "com.gu" %% "membership-common" % "0.525",
-    "com.gu.identity" %% "identity-play-auth" % "2.4",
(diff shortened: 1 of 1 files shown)


lmath:
Gah, I'm looking into the failing test right now.

lmath:
OK, I've also bumped jackson-annotations and identity-play-auth and then bought a product with a test user and checked it appeared in Zuora. Please do take a look now @jranks123 @Ap0c @JustinPinner

PR #1185, opened 2018-10-17
Upgrade dispatch

jranks123:
https://app.snyk.io/vuln/SNYK-JAVA-ORGBEANSHELL-72452

Due to having an old version of dispatch, we had a dependency, specifically beanshell, which had the above security issue.

I have tested putting a guest user digipack sub through on CODE


johnduffell:
I don't know if we even need dispatch.  I have removed it from membership common as it is designed for a different http library from what we use.  We were only using a retry function. https://github.com/guardian/membership-common/pull/584

PR #1186, opened 2018-10-17
Provide a simple test to ensure getAddressIO api is accessible for a valid postcode.

LATaylor-guardian:


jacobwinch:
I think this will block deployment if getAddress are down / suffering from performance problems. I don't think we'd want that (although we should definitely know if we can't successfully call their API!).

We could consider running this test post-deployment instead, so that we get notified about problems, without being blocked?  We already do this with some other tests, see https://github.com/guardian/subscriptions-frontend/blob/master/build.sbt#L125.

LATaylor-guardian:
@jacobwinch Perfect, thanks that's exactly what I was looking for.

build.sbt
```diff
@@ -121,8 +121,9 @@ javaOptions in Universal ++= Seq(
 
 addCommandAlias("devrun", "run  9200")
 addCommandAlias("prodrun", "run 9200")
-addCommandAlias("fast-test", "testOnly -- -l Acceptance -l WeeklyCheckoutAcceptance -l WeeklyPromoAcceptance")
+addCommandAlias("fast-test", "testOnly -- -l Acceptance -l WeeklyCheckoutAcceptance -l WeeklyPromoAcceptance -l GetAddressIOServiceAcceptance")
 addCommandAlias("acceptance-test", "testOnly acceptance.CheckoutSpec")
```
jacobwinch: I think you'll want to make this change too, to ensure that Prout runs these tests (using Travis) after every deploy:

```suggestion
addCommandAlias("acceptance-test", "testOnly acceptance.CheckoutSpec acceptance.GetAddressIOServiceTest")
```

LATaylor-guardian: updated.

PR #1187, opened 2018-10-18
Improve error handling for postcode lookups

jacobwinch:
This should help us to spot problems in production faster, as common failures (authentication problems, rate limiting) should now be logged to Sentry (and included in application logs).

This will be particularly useful when this app is not being worked on frequently (the test added here should catch these problems during development / on deployment: https://github.com/guardian/subscriptions-frontend/pull/1186).

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -417,7 +417,11 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
       // should a client want not to proxy via this server.
       Ok(Json.obj("Addresses" -> result.Addresses))
     } recover {
-      case e => BadRequest(Json.obj("Message" -> e.getMessage))
+      case error if error.getMessage == "Bad Request" =>
+        BadRequest //The postcode was invalid
+      case error =>
+        SafeLogger.error(scrub"Failed to complete postcode lookup via getAddress.io due to: $error")
+        InternalServerError
     }
   }
 


app/controllers/Checkout.scala
```diff
@@ -417,7 +417,11 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
       // should a client want not to proxy via this server.
       Ok(Json.obj("Addresses" -> result.Addresses))
     } recover {
-      case e => BadRequest(Json.obj("Message" -> e.getMessage))
+      case error if error.getMessage == "Bad Request" =>
+        BadRequest //The postcode was invalid
```
jacobwinch: The client is not using the error response: https://github.com/guardian/subscriptions-frontend/blob/e057a5b8dfd8ec94c48ebae4455f0e1fcb6aed5c/assets/javascripts/modules/checkout/addressFinder.jsx#L42-L49

app/controllers/Checkout.scala
```diff
@@ -417,7 +417,11 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
       // should a client want not to proxy via this server.
       Ok(Json.obj("Addresses" -> result.Addresses))
     } recover {
-      case e => BadRequest(Json.obj("Message" -> e.getMessage))
+      case error if error.getMessage == "Bad Request" =>
```
LATaylor-guardian: Will the message equal "Bad Request"? Have you seen this?

LATaylor-guardian: Another case you may also add which we saw during the incident was the message being equal to: `Your account has expired. Please upgrade your account.`

jacobwinch: Yeah I tested it locally with a garbage postcode.

LATaylor-guardian: It's unfortunate here that the abstraction around the http call doesn't model all of the states as this would alleviate having to rely on a String.

LATaylor-guardian:
👍

app/controllers/Checkout.scala
jacobwinch: I'm not keen on the abstraction used here either (a 404 from getAddress shouldn't result in an Exception IMO, since it's perfectly valid for a user to look up a nonexistent postcode).

I'd prefer to handle the responses based on the HTTP status code, but since that has all been abstracted away, this seem like a quick fix.

LATaylor-guardian: Yeah completely agree -it's outside the scope of this.

PR #1188, opened 2018-10-18
Simplify acceptance tests.

LATaylor-guardian:
We want two scenarios:
1. Unit test suite to run and acceptances tests not to.
2. Acceptance tests to run and unit test suite not to.

By providing a single tag `AcceptanceTest`, we can run all tests without it in the first scenario, and all tests with it in the second scenario.
This also means that when adding a new acceptance test you simply add the same tag and don't have to modify the sbt commands as these are something we don't want to care about during development.

I've removed the individual commands to run individual acceptance tests as these are more boilerplate and unlikely to be used individually given there is a very small number of acceptance tests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -121,10 +121,8 @@ javaOptions in Universal ++= Seq(
 
 addCommandAlias("devrun", "run  9200")
 addCommandAlias("prodrun", "run 9200")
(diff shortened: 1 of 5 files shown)


jacobwinch:
^^^ The test failures seem to be due to Travis running some tests which have been invalid since we started selling the new Guardian Weekly plans. I'll have a tidy up...

PR #1192, opened 2018-10-18
Test against paid getAddress account

jacobwinch:
Follow up to: https://github.com/guardian/subscriptions-frontend/pull/1186

This ensures that we are using the same API key that is used by the production app. We store it in a private environment variable in Travis.


LATaylor-guardian:
👍 Perfect.

PR #1193, opened 2018-10-18
Fix broken build

jacobwinch:
My previous change (https://github.com/guardian/subscriptions-frontend/pull/1192) broke the TeamCity build, and meant that the Acceptance Tests would not run locally.

This fixes those issues.


johnduffell:
👍 good detective work!

PR #1195, opened 2018-10-19
Do not log errors if getAddress cannot find the postcode

jacobwinch:
We should only log errors in appropriate cases.

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -417,7 +417,7 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
       // should a client want not to proxy via this server.
       Ok(Json.obj("Addresses" -> result.Addresses))
     } recover {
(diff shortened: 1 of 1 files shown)


app/controllers/Checkout.scala
```diff
@@ -417,7 +417,7 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
       // should a client want not to proxy via this server.
       Ok(Json.obj("Addresses" -> result.Addresses))
     } recover {
-      case error if error.getMessage == "Bad Request" =>
+      case error if error.getMessage == "Bad Request" || error.getMessage == "Not Found" =>
```
jacobwinch: This error message came through in the logs this morning...

jacobwinch: (It's not what I saw when testing locally, I guess getAddress make a distinction between a garbage string and something which looks like a postcode, but isn't actually a postcode)

PR #1196, opened 2018-10-19
Stop serving the old Weekly checkout pages

jacobwinch:
It has been impossible to buy the old plans since https://github.com/guardian/subscriptions-frontend/pull/1179 was merged.

This PR means that we'll stop serving the checkout pages for the old plans too (and we will have less irrelevant code to read and maintain).

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -67,8 +67,6 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
 
       val matchingPlanList: Option[PlanList[ContentSubscription]] = {
 
-        val testOnlyPlans = if (tpBackend == fBackendFactory.Test) List(catalog.weekly.zoneB.plansWithAssociations) else List.empty
-
         val productsWithoutIntroductoryPlans = List(
           catalog.delivery.list.toList,
           catalog.voucher.list.toList,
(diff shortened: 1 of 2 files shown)


app/controllers/Checkout.scala
```diff
@@ -157,6 +136,7 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
                 getSettings(determinedCountryGroup.defaultCountry, determinedCountryGroup.currency)
               }
             }
+            case _ => getSettings(determinedCountryGroup.defaultCountry, determinedCountryGroup.currency)
```
jacobwinch: This seemed like a sensible default case to me

app/controllers/Checkout.scala
```diff
@@ -67,20 +67,16 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
 
       val matchingPlanList: Option[PlanList[ContentSubscription]] = {
 
-        val testOnlyPlans = if (tpBackend == fBackendFactory.Test) List(catalog.weekly.zoneB.plansWithAssociations) else List.empty
```
jacobwinch: I think this has been hanging around since the GW migration - I don't think anyone is testing with Guardian Weekly Zone B plans anymore.

app/controllers/Checkout.scala
lmath: 👍 agreed

PR #1191, opened 2018-10-18
Update footer text of checkout pages.

LATaylor-guardian:
Before: 
picture 215

After: 
picture 216

PR #1197, opened 2018-10-19
Update footer for all checkouts.

lmath:
Just picking up from Luke's PR https://github.com/guardian/subscriptions-frontend/pull/1191 but making it more product specific.

Re: the links, I've put in a link to the privacy policy as https://www.theguardian.com/help/privacy-policy, assuming it is the same for all products. 

existing footer:
image

updated footer:
image

This also adds a privacy policy and T&C link to the other products (Home Delivery, Voucher, Guardian Weekly). I've got a Q about the reference to T&Cs below.

### DIFF
--- a/app/views/fragments/global/footer.scala.html
+++ b/app/views/fragments/global/footer.scala.html
@@ -18,13 +18,13 @@
             By proceeding, you are agreeing to our
             <a href="@maybeProduct.getOrElse(Digipack).termsAndConditionsHref"
                 title="Terms and Conditions">
-                Terms and Conditions
-            </a>.
+                Terms and Conditions.
+            </a>
                 To find out what personal data we collect and how we use it, please visit our
                 <a href="https://www.theguardian.com/help/privacy-policy"
                     title="Privacy Policy">
-                    Privacy Policy
-                </a>.
+                    Privacy Policy.
+                </a>
             <p>
               For help with Guardian and Observer subscription services please email
                   <a href="mailto:@email">@email</a> or call @phone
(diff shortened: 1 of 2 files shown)


app/views/support/PlanOps.scala
```diff
@@ -126,6 +126,15 @@ object PlanOps {
         case _ => "https://www.theguardian.com/subscriber-direct/subscription-frequently-asked-questions"
       }
 
+    def termsAndConditionsHref: String =
+      product match {
+        case Delivery => "https://www.theguardian.com/subscriber-direct/subscription-terms-and-conditions"
+        case Voucher => "https://www.theguardian.com/subscriber-direct/subscription-terms-and-conditions"
+        case Product.Digipack => "https://www.theguardian.com/info/2014/aug/06/guardian-observer-digital-subscriptions-terms-conditions"
+        case _: Product.Weekly => "https://www.theguardian.com/info/2014/jul/10/guardian-weekly-print-subscription-services-terms-conditions"
+        case _ => "Sensible default" //TODO
```
lmath: Product also includes Contributions and Membership, so I'm undecided about what to link in that case. Any ideas?

app/views/fragments/global/footer.scala.html
```diff
@@ -23,8 +33,13 @@
                 You may also find help in our
                 <a href="@maybeProduct.getOrElse(Digipack).faqHref"
                    title="Frequently Asked Questions">
-                   Frequently Asked Questions.
+                   Frequently Asked Questions
                 </a>
+                and the @maybeProduct.getOrElse(Digipack).productType
```
lmath: The footer is shared among the products so we need to think about what to do for the other products when referencing the digital pack T&Cs. Current behaviour uses this mapping https://github.com/guardian/subscriptions-frontend/blob/master/app/views/support/PlanOps.scala#L131

lmath:
Oooops, closing this PR because the card was about the new digital pack checkout

PR #1206, opened 2018-10-29
snyk: stop using dispatch library for retries

johnduffell:
exactly the same description as https://github.com/guardian/membership-common/pull/584


app/services/ExactTargetService.scala
```diff
@@ -5,7 +5,6 @@ import com.amazonaws.services.sqs.AmazonSQSClient
 import com.amazonaws.services.sqs.model._
 import com.github.nscala_time.time.Imports._
 import com.gu.aws.CredentialsProvider
-import com.gu.lib.Retry
```
johnduffell: this is the only substantial change to the imports, the rest was just intellij's idea of optimal import order

johnduffell:
seems to be working OK from local testing guardian weekly plus running all the tests from sbt

lmath:
👍 Nice one spotting this is all we were using dispatch for!

PR #1209, opened 2018-10-31
Align the header text

walaura:
screenshot 2018-10-31 at 10 39 08 am

Applies a couple design requests from @zeftilldeath to the header, namely:

- align the copy
- change the font to text-sans

Also contains a tweaked stylesheet for the 1 million campaign


assets/stylesheets/million.scss
```diff
@@ -0,0 +1,111 @@
+// ----- Main Stylesheets ----- //
```
walaura: Colours aside this is a copypaste of `garnett.scss`. Any non-garnett changes to that stylesheet that were to happen before the campaign will need to be replicated here

After the campaign, there's some minimum health work needed to decommission one of them to avoid this duplication

PR #1212, opened 2018-11-02
Annualised value tracking

rupertbates:
Upgrade the `acquisition-event-producer` library to start sending Annualised Value to GA in the Ecommerce revenue metric.

While testing this I noticed that the configuration for the CODE environment was a bit messed up so there are some fixes for that as well.

### DIFF
--- a/cloud-formation/subscriptions-app.cf.yaml
+++ b/cloud-formation/subscriptions-app.cf.yaml
@@ -180,6 +180,12 @@ Resources:
           - Effect: Allow
             Action: sqs:*
             Resource: arn:aws:sqs:eu-west-1:865473395570:subs-holiday-suspension-email*
+      - PolicyName: InvokeAVLambda
+        PolicyDocument:
+          Statement:
+          - Effect: Allow
+            Action: sts:AssumeRole
+            Resource: arn:aws:iam::021353022223:role/support-invoke-value-calculator
       ManagedPolicyArns:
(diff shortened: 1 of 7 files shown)


conf/touchpoint.CODE.conf
```diff
@@ -0,0 +1,34 @@
...
+            paymentDelayGracePeriod = 2
+            api {
+                username = ""
+                password = ""
+            }
+        }
+        stripe {
+            api.key {
+                secret = ""
+                public = "pk_test_Qm3CGRdrV4WfGYCpm0sftR0f"
```
tomrf1: is it ok to check this in?

rupertbates: Yes that's a public key, the secret on is in private config

cloud-formation/subscriptions-app.cf.yaml
```diff
@@ -180,6 +180,12 @@ Resources:
           - Effect: Allow
             Action: sqs:*
             Resource: arn:aws:sqs:eu-west-1:865473395570:subs-holiday-suspension-email*
+      - PolicyName: InvokeAVLambda
+        PolicyDocument:
+          Statement:
+          - Effect: Allow
+            Action: sts:AssumeRole
+            Resource: arn:aws:iam::021353022223:role/support-invoke-value-calculator
```
tomrf1: it's good practice not to commit the arn

rupertbates: That's not the convention in this file though

PR #1213, opened 2018-11-07
add paypal method to trigger properties

QuarpT:
Required by new Braze email templates


app/model/exactTarget/DataExtensionRow.scala
```diff
@@ -165,7 +166,7 @@ object PaperFieldsGenerator {
         "mandate_id" -> mandateId
       )
       case _: PaymentCard => Seq("payment_method" -> card)
-      case _ => Seq()
+      case _: PayPalMethod => Seq("payment_method" -> paypal)
```
Mullefa: Just to check, is this an exhaustive match on payment methods?

PR #1217, opened 2018-12-04
Temporarily extend 6 for 6 plan to cover 7 weeks

jacobwinch:
We will not publish a Guardian Weekly magazine on 28th December, so as things stand we are actually selling 6 for 5. 

This change allows us to extend the 6 for 6 fulfilment period in order to correct this problem.

Note that I am planning to take the following steps to facilitate this change:

1. Update membership-common for all apps which use the catalog service, to pick up: https://github.com/guardian/membership-common/pull/585
1. Update catalog (all environments)
1. Merge this PR

If anybody signs up for 6 for 6 between steps 2 and 3 they will end up getting two magazines on the date of their 7th issue. As the likelihood of this happening is pretty small I think it's acceptable to introduce this bug briefly.

More context:
https://trello.com/c/2aCynmAW/2103-change-6-for-6-promo-to-give-7-issues-to-cover-missing-issue-over-christmas
https://trello.com/c/Img8mLoX/222-implement-workaround-for-guardian-weekly-6-for-6-fulfilment-issue-currently-due-to-affect-users-on-28th-december

### DIFF
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -101,7 +101,7 @@ class CheckoutService(
           ),
           RatePlan(recurringPlan.id.get, None)
         )
-        val updatedContractAcceptance = sixWeeksPlanStartDate.plusWeeks(6)
+        val updatedContractAcceptance = sixWeeksPlanStartDate.plusWeeks(7) // FIXME: edit rate plan charge in Zuora and make this 6 again after 28th December
         originalCommand.copy(ratePlans = replacementPlans, contractAcceptance = updatedContractAcceptance)
       }
       updatedCommand.getOrElse(originalCommand)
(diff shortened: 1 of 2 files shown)


app/services/CheckoutService.scala
```diff
@@ -101,7 +101,7 @@ class CheckoutService(
           ),
           RatePlan(recurringPlan.id.get, None)
         )
-        val updatedContractAcceptance = sixWeeksPlanStartDate.plusWeeks(6)
+        val updatedContractAcceptance = sixWeeksPlanStartDate.plusWeeks(7) // FIXME: edit rate plan charge in Zuora and make this 6 again after 28th December
```
rupertbates: Classy ;-)

jacobwinch: I told you it was a hack!

PR #1219, opened 2018-12-10
Update deliveryFields.jsx

akash1810:
[card](https://trello.com/c/UgBMvSnm/2099-reduce-waiting-time-calculation-in-voucher-checkout-to-3-weeks)


akash1810:
This is actually the wrong file 😊. Additionally, we're already using [3 weeks](https://github.com/guardian/subscriptions-frontend/blob/master/assets/javascripts/modules/checkout/voucherFields.jsx#L18) so no work needed. 🎉

PR #1222, opened 2018-12-13
Gifting for guardian weekly

paulbrown1982:
This change enables customers to buy Guardian Weekly subscriptions as a gift. It excludes Six-for-six subscriptions.

It depends on https://github.com/guardian/membership-common/pull/586

The main UI change is a checkbox on the checkout page that switches the Delivery Address section to become a Gift Recipient section, and takes 3 new fields: Giftee's name, last name and email address.

Behind the scenes, the Salesforce orchestration creates a Delivery / Recipient Contact (Representing the gift recipient) alongside or in addition to an existing Standard Contact (Representing the buyer). A logged in customer can buy as many gifts as they like, and it will create that many new Delivery / Recipient Contacts without interfering with the Standard Contact's delivery address.

Changes in how subs are read back out of Zuora (in /manage for example), have also been made to ensure either only the buyer contact is retrieved. This uses the new Membership Common SalesforceContactServiceUsingZuoraRest object.

**Screenshots:**

cc @jacobwinch @pvighi

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -152,7 +152,7 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
 
           val resolvedSupplierCode = supplierCode orElse request.session.get(SupplierTrackingCode).map(SupplierCode) // query param wins
           val supplierCodeSessionData = resolvedSupplierCode.map(code => Seq(SupplierTrackingCode -> code.get)).getOrElse(Seq.empty)
-          val productData = ProductPopulationData(user.map(_.address), planList)
+          val productData = ProductPopulationData(user.map(_.correspondenceAddress), planList)
           Ok(views.html.checkout.payment(
             personalData = personalData,
             productData = productData,
--- a/app/model/PurchaserIdentifiers.scala
+++ b/app/model/PurchaserIdentifiers.scala
@@ -3,7 +3,15 @@ package model
 import com.gu.identity.play.IdMinimalUser
 import com.gu.salesforce.ContactId
 
-case class PurchaserIdentifiers(contactId: ContactId, identityId: Option[IdMinimalUser]) {
-  val description = (Seq(s"SalesforceContactID - ${contactId.salesforceContactId}") ++ identityId.map(id => s"IdentityID - $id")).mkString(" ")
+case class PurchaserIdentifiers(buyerContactId: ContactId, recipientContactId: ContactId, identityId: Option[IdMinimalUser]) {
+  val description = s"""
+      |Salesforce data: Account: ${buyerContactId.salesforceAccountId},
+      |Buyer Contact: ${buyerContactId.salesforceContactId},
+      |Recipient Contact: ${recipientContactId.salesforceContactId},
+      |Identity ID: ${identityId.map(_.id).mkString},
+      |Is gift? ${isGift}
+    """.stripMargin.trim
+  val isGift = buyerContactId.salesforceAccountId == recipientContactId.salesforceAccountId &&
+    buyerContactId.salesforceContactId != recipientContactId.salesforceContactId
   override def toString: String = description
 }
--- a/app/model/SubscriptionData.scala
+++ b/app/model/SubscriptionData.scala
@@ -43,6 +43,7 @@ case class PersonalData(first: String,
                         telephoneNumber: Option[String] = None,
                         title: Option[Title] = None
                         ) extends FullName {
+
   def fullName = s"$first $last"
 
   private lazy val countryGroup = CountryGroup.byCountryNameOrCode(address.country.fold(UK.alpha2)(c => c.alpha2))
@@ -81,8 +82,8 @@ object PersonalData {
 
     val personalData = PersonalData(
       title = u.privateFields.flatMap(_.title).flatMap(Title.fromString),
-      first = u.privateFields.flatMap(_.firstName).getOrElse(""),
-      last = u.privateFields.flatMap(_.secondName).getOrElse(""),
+      first = u.privateFields.flatMap(_.firstName).mkString,
+      last = u.privateFields.flatMap(_.secondName).mkString,
       email = u.primaryEmailAddress,
       receiveGnmMarketing = u.statusFields.flatMap(_.receiveGnmMarketing).getOrElse(false),
       address = u.billingAddress,
@@ -95,4 +96,10 @@ object PersonalData {
 
 case class SubscribeRequest(genericData: SubscriptionData, productData: Either[PaperData, DigipackData]) {
   def productRatePlanId = productData.fold(_.plan.id, _.plan.id)
+}
+
+case class DeliveryRecipient(title: Option[Title], firstName: Option[String], lastName: Option[String], email: Option[String], address: Address) extends FullName {
+  val first = firstName.mkString
+  val last = lastName.mkString
+  val isGiftee = (Seq() ++ title.map(_.title) ++ firstName ++ lastName ++ email).exists(_.trim.nonEmpty)
 }
--- a/app/model/SubscriptionData.scala
+++ b/app/model/SubscriptionData.scala
@@ -64,7 +64,7 @@ case class DigipackData(
 
 case class PaperData(
   startDate: LocalDate,
-  deliveryAddress: Address,
+  deliveryRecipient: DeliveryRecipient,
   deliveryInstructions: Option[String],
   plan: CatalogPlan.Paper
 ) {
--- a/app/services/CheckoutService.scala
+++ b/app/services/CheckoutService.scala
@@ -121,7 +121,7 @@ class CheckoutService(
     val idMinimalUser = authenticatedUserOpt.map(_.user)
 
     val telephoneNumber = subscriptionData.productData match {
-      case Left(paperData) => NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, paperData.deliveryRecipient.address.country orElse personalData.address.country)
+      case Left(paperData) => NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, personalData.address.country orElse paperData.deliveryRecipient.address.country)
       case _ => NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, personalData.address.country)
     }
 
@@ -130,7 +130,8 @@ class CheckoutService(
         name = Some(paperData.deliveryRecipient).filter(_.isGiftee).getOrElse(personalData),
         address = paperData.deliveryRecipient.address,
         email = Some(paperData.deliveryRecipient).filter(_.isGiftee).map(_.email).getOrElse(personalData.email),
-        deliveryInstructions = paperData.sanitizedDeliveryInstructions
+        deliveryInstructions = paperData.sanitizedDeliveryInstructions,
+        phone = Some(paperData.deliveryRecipient).filterNot(_.isGiftee).flatMap(_ => telephoneNumber)
       ))
       case _ => None
     }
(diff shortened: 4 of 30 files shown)


app/controllers/Checkout.scala
```diff
@@ -152,7 +152,7 @@ class Checkout(fBackendFactory: TouchpointBackends, commonActions: CommonActions
 
           val resolvedSupplierCode = supplierCode orElse request.session.get(SupplierTrackingCode).map(SupplierCode) // query param wins
           val supplierCodeSessionData = resolvedSupplierCode.map(code => Seq(SupplierTrackingCode -> code.get)).getOrElse(Seq.empty)
-          val productData = ProductPopulationData(user.map(_.address), planList)
+          val productData = ProductPopulationData(user.map(_.correspondenceAddress), planList)
```
paulbrown1982: This was a bug, where we were pre-filling the wrong delivery address from Identity, we were using the Billing not the Correspondence.

app/services/SalesforceService.scala
```diff
@@ -36,15 +49,29 @@ object SalesforceService {
     Keys.ALLOW_GU_RELATED_MAIL -> personalData.receiveGnmMarketing
   ) ++ personalData.title.fold(Json.obj())(title => Json.obj(
     Keys.TITLE -> title.title
-  )) ++ paperData.map(_.deliveryAddress).fold(Json.obj())(addr => Json.obj(
+  )) ++ paperData.map(_.deliveryRecipient).filterNot(_.isGiftee).map(_.address).fold(Json.obj())(addr => Json.obj(
```
paulbrown1982: This stops us upserting the delivery recipient's address into the buyer's delivery address if they are buying a gift.

app/views/fragments/checkout/fieldsPersonal.scala.html
```diff
@@ -1,57 +1,62 @@
...
 @import model.PersonalData
 
-@(form: Option[PersonalData])
+@(form: Option[PersonalData], productSupportsGifting: Boolean)
 
 @signedInAttrs = @{ if (form.isDefined) "required readonly" else "" }
 
 @* ===== Name ===== *@
 <div class="form-field js-checkout-title">
-    <label class="label" for="title">Title</label>
```
paulbrown1982: This was pulled out into its own template used by the gift (delivery) and the personal sections.

app/model/SubscriptionData.scala
```diff
@@ -96,3 +97,9 @@ object PersonalData {
 case class SubscribeRequest(genericData: SubscriptionData, productData: Either[PaperData, DigipackData]) {
   def productRatePlanId = productData.fold(_.plan.id, _.plan.id)
 }
+
+case class DeliveryRecipient(title: Option[Title], firstName: Option[String], lastName: Option[String], email: Option[String], address: Address) extends FullName {
+  val first = firstName.mkString
+  val last = lastName.mkString
+  val isGiftee = (Seq() ++ title.map(_.title) ++ firstName ++ lastName ++ email).exists(_.trim.nonEmpty)
```
jacobwinch: I don't understand why this confirms that someone is a giftee?

app/services/CheckoutService.scala
```diff
@@ -121,17 +122,22 @@ class CheckoutService(
     val idMinimalUser = authenticatedUserOpt.map(_.user)
 
     val telephoneNumber = subscriptionData.productData match {
-      case Left(paperData) => NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, paperData.deliveryAddress.country orElse personalData.address.country)
+      case Left(paperData) => NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, personalData.address.country orElse paperData.deliveryRecipient.address.country)
```
jacobwinch: Looking at the form, I think the phone number always belongs to the buyer, so I don't understand why we would ever use data about the delivery recipient?

paulbrown1982: Because the personalData.address is null at this point if the user ticks the 'Use Delivery Address as billing' address. Hence the old logic was a bug giving higher precedence to the delivery country than the billing country when determining the phone number prefix.

app/services/SalesforceService.scala
```diff
@@ -5,26 +5,39 @@ import com.gu.memsub.NormalisedTelephoneNumber
...
   def isAuthenticated: Boolean
 }
 
 class SalesforceServiceImp(val repo: SimpleContactRepository) extends SalesforceService {
-  override def createOrUpdateUser(personalData: PersonalData, paperData: Option[PaperData], userId: Option[IdMinimalUser]): Future[ContactId] =
-    repo.upsert(userId.map(_.id), SalesforceService.createSalesforceUserData(personalData, paperData))
+  override def createOrUpdateBuyerAndRecipient(personalData: PersonalData, paperData: Option[PaperData], userId: Option[IdMinimalUser]): Future[PurchaserIdentifiers] = {
+    for {
+      buyerContact <- repo.upsert(userId.map(_.id), SalesforceService.getJSONForBuyerContact(personalData, paperData))
+      recipientContact <- { // Only create a Delivery / Recipient (Related) Contact if the recipient is a Giftee
```
jacobwinch: We could make this clearer (and testable) by pulling the code which decides to create a recipient into a separate function?

assets/javascripts/modules/checkout/gifting.js
```diff
@@ -0,0 +1,48 @@
...
+                        $CHECKOUT_DELIVERY_SAMES_AS_BILLING.attr('disabled', 'disabled');
+                        $DELIVERY_DETAILS_LEGEND.text('Gift recipient details');
+                        $GIFT_RECIPIENT_DETAILS.removeClass('is-hidden');
+                        $BILL_MY_DELIVERY_ADDRESS_SECTION.addClass('is-hidden');
+                        $BILLING_ADDRESS_SECTION.removeClass('is-hidden');
+                    } else {
+                        $CHECKOUT_DELIVERY_SAMES_AS_BILLING.attr('checked');
+                        $DELIVERY_DETAILS_LEGEND.text('Delivery address');
+                        $GIFT_RECIPIENT_DETAILS.addClass('is-hidden');
+                        //$BILL_MY_DELIVERY_ADDRESS_SECTION.removeClass('is-hidden');
```
jacobwinch: Can these commented out lines be deleted?

app/services/SalesforceService.scala
jacobwinch: I would add some logging here too

app/services/CheckoutService.scala
```diff
@@ -121,17 +122,22 @@ class CheckoutService(
...
       case _ => NormalisedTelephoneNumber.fromStringAndCountry(personalData.telephoneNumber, personalData.address.country)
     }
 
+    val readerType: ReaderType = subscriptionData.productData match {
+      case Left(paperData) if paperData.deliveryRecipient.isGiftee => Gift
```
jacobwinch: So technically we could do gifts for Voucher+HD but we have decided not to offer it yet?

app/views/checkout/payment.scala.html
```diff
@@ -169,7 +169,7 @@
                             <a href="#yourDetails" class="text-button u-button-reset js-edit-your-details" title="Edit your personal details">Edit</a>
                         </div>
                         <div class="field-panel__fields">
-                            @fragments.checkout.fieldsPersonal(personalData)
+                            @fragments.checkout.fieldsPersonal(personalData, productData.isGuardianWeekly && !productData.isSixForSix)
```
jacobwinch: Minor: I would add the productSupportsGifting parameter name here

app/model/PurchaserIdentifiers.scala
```diff
@@ -3,7 +3,15 @@ package model
 import com.gu.identity.play.IdMinimalUser
 import com.gu.salesforce.ContactId
 
-case class PurchaserIdentifiers(contactId: ContactId, identityId: Option[IdMinimalUser]) {
-  val description = (Seq(s"SalesforceContactID - ${contactId.salesforceContactId}") ++ identityId.map(id => s"IdentityID - $id")).mkString(" ")
+case class PurchaserIdentifiers(buyerContactId: ContactId, recipientContactId: ContactId, identityId: Option[IdMinimalUser]) {
```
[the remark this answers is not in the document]
jacobwinch: Also PurchaserIdentifiers no longer seems like an appropriate name for this (since it may now include the identifier of a giftee!)

app/model/SubscriptionData.scala
paulbrown1982: I'll refactor this in a subsequent PR.

jacobwinch:
Looks good.

Let's get this out and pick up some of the refactoring in subsequent PRs 👍

app/model/PurchaserIdentifiers.scala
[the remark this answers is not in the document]
paulbrown1982: I'll look into refactoring this from membership-common upwards, tomorrow.

app/services/CheckoutService.scala
paulbrown1982: Correct, the backend elements of this change will work for any paper product. The restriction/filter is limited to frontend elements.

jacobwinch: Cool, thanks for confirming 👍

assets/javascripts/modules/checkout/gifting.js
paulbrown1982: Whoops, deleting that now.

app/services/CheckoutService.scala
```diff
@@ -121,17 +122,22 @@ class CheckoutService(
...
-        name = personalData, // TODO once we have gifting change this to the Giftee's name
-        address = paperData.deliveryAddress,
-        email = personalData.email, // TODO once we have gifting change this to the Giftee's email address
-        phone = telephoneNumber,
-        deliveryInstructions = paperData.sanitizedDeliveryInstructions
+        name = if (readerType == Gift) paperData.deliveryRecipient else personalData,
+        address = paperData.deliveryRecipient.address,
+        email = if (readerType == Gift) paperData.deliveryRecipient.email else Some(personalData.email),
+        deliveryInstructions = paperData.sanitizedDeliveryInstructions,
+        phone = if (readerType == Direct) telephoneNumber else None
```
pvighi: this is minor so we could fix it in another PR if we need to release this, But maybe it would be more clear with different cases for the different reader types:

we could do: 
```scala
(subscriptionData.productData, readerType) match{
case (Left(paperData), Gift) => [gift stuff]
case (Left(paperData), Direct) => [direct stuff]
case (Left(paperData), Agent) => [agent  stuff] //maybe this case should not exist?
case _ => None
```
we could also do this if in the cases and just matching on the productData:
```scala 
 case  Left(PaperData) if (readerType == Gift) => do stuff
```

app/services/CheckoutService.scala
pvighi: I'm confused as to where the billing country will be at this point..
but if we should give higher precedence to the billing country why would we ever check the delivery country? isn't billing country supposed to always be there ?

app/services/SalesforceService.scala
pvighi: this line looks very similar to  [line 218 in CheckoutService](https://github.com/guardian/subscriptions-frontend/pull/1222/files#diff-68b2eee9ecf473f772caa4b38dbdaa17R218). 
if we could extract it to a common location we could remove the duplication.  we could call it correspondence address and put it DeliveryRecipient, but I don't know if conceptually it would make sense here to use the "correspondence address"

app/views/fragments/checkout/giftRecipientDetails.scala.html
```diff
@@ -0,0 +1,37 @@
...
+    Html(s"id='${(namePrefix + "-" + field).replace('.', '-')}' name='$namePrefix.$field'")
+}
+
+@labelFor(field: String) = @{
+    Html(s"for='${(namePrefix + "-" + field).replace('.', '-')}'")
+}
+
+<div class="js-gift-recipient-details is-hidden">
+    @********************************************************************************************
+    * A decision was ade not to collect the Title for gift recipients, but in the event that    *
```
pvighi: * made

app/services/SalesforceService.scala
paulbrown1982: I have an idea on how to improve this. I'll submit that in an additional PR later today.

app/services/SalesforceService.scala
paulbrown1982: Agreed, this is better off being in a separate function. I'll raise a separate PR later today.

app/services/CheckoutService.scala
paulbrown1982: Interesting, I'll have a think about this as I also don't want to add lots of repeated setters in the [stuff] section.

jacobwinch:
> Testing in PROD failed! [Details](https://travis-ci.org/guardian/subscriptions-frontend/builds/467903266?utm_source=github_status&utm_medium=notification)

The test failure seems to be genuine so we have backed this out whilst we investigate...

PR #1224, opened 2018-12-14
Fixed the voucher checkout JS error.

paulbrown1982:
Fixes JS error on voucher checkout: 
```
raven.js:351 Uncaught TypeError: Cannot read property 'length' of null
    at HTMLAnchorElement.handleBillingVisibility (deliveryDetails.js:75)
    at call (bean.js:254)
    at HTMLAnchorElement.handler (bean.js:269)
    at HTMLAnchorElement.rootListener (bean.js:434)
    at HTMLAnchorElement.wrapped (raven.js:347)
09:24:03.696
```
cc @jacobwinch

### DIFF
--- a/app/views/fragments/checkout/fieldsVoucher.scala.html
+++ b/app/views/fragments/checkout/fieldsVoucher.scala.html
@@ -5,7 +5,7 @@
 
 @_root_.views.html.fragments.checkout.address(
     address = data.deliveryAddress,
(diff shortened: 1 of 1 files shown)


jacobwinch:
This changes the ids which the Selenium tests depend on, which means that the tests are failing even thought the flow has been fixed... Will raise a fix PR in a few minutes.

PR #1235, opened 2019-01-25
Enable Optimize in Code environment

jacobwinch:
This caught me out for a while during some testing I was doing earlier today. I think it's helpful to have Optimise enabled in CODE environments so that we can check our work before setting up tests against PROD.


walaura:
i was able to use optimize in code for the DP landing page test so now I'm kinda confused as to how i did that ??

jacobwinch:
> i was able to use optimize in code for the DP landing page test so now I'm kinda confused as to how i did that ??

@walaura this change enables Optimize in subscriptions-frontend CODE, I think it's been enabled in support-frontend CODE for a while. Presumably your test only relates to the latter?

PR #1237, opened 2019-02-04
Rename 'Subscriber ID' to 'Subscription ID'

tjsilver:
A user can have more than one subscription so 'Subscriber ID' is no longer appropriate. It has been changed to 'Subscription ID' to match changes in manage-frontend.

*_Logged in_*:
image

*_Logged out_*:
image

*_Confirmation page_*:
image

### DIFF
--- a/app/views/digitalpack/info.scala.html
+++ b/app/views/digitalpack/info.scala.html
@@ -98,7 +98,7 @@
                 <li class="grid__item">
                     <span class="numbered">
                         <span class="numbered__bullet">1.</span>
-                        <span class="numbered__body">Order now and we will send you a subscriber ID</span>
+                        <span class="numbered__body">Order now and we will send you a subscription ID</span>
                     </span>
                 </li>
                 <li class="grid__item">
@@ -110,7 +110,7 @@
                 <li class="grid__item">
                     <span class="numbered">
                         <span class="numbered__bullet">3.</span>
-                        <span class="numbered__body">Enter your subscriber ID into the app, and enjoy</span>
+                        <span class="numbered__body">Enter your subscription ID into the app, and enjoy</span>
                     </span>
                 </li>
             </ol>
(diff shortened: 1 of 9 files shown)


app/views/digitalpack/info.scala.html
```diff
@@ -98,7 +98,7 @@ <h3 class="page-slice__title">How do I get it?</h3>
                 <li class="grid__item">
                     <span class="numbered">
                         <span class="numbered__bullet">1.</span>
-                        <span class="numbered__body">Order now and we will send you a subscriber ID</span>
+                        <span class="numbered__body">Order now and we will send you a subscription ID</span>
```
paulbrown1982: I think we should keep it consistent with a capital S on Subscription Id

app/views/digitalpack/info.scala.html
```diff
@@ -110,7 +110,7 @@ <h3 class="page-slice__title">How do I get it?</h3>
                 <li class="grid__item">
                     <span class="numbered">
                         <span class="numbered__bullet">3.</span>
-                        <span class="numbered__body">Enter your subscriber ID into the app, and enjoy</span>
+                        <span class="numbered__body">Enter your subscription ID into the app, and enjoy</span>
```
paulbrown1982: Same here, let's have a capital S.

app/views/fragments/promotion/subscriptionTermsAndConditions.scala.html
```diff
@@ -12,7 +12,7 @@ <h4>Digital subscription terms and conditions</h4>
     <p>
         Subscriptions available to people aged 18 and over with a valid email address. Free trial open to new digital pack subscribers only. Free trial period lasts
         @promotion.asFreeTrial.fold(defaultDigitalPackFreeTrialPeriod.getDays.toString)(_.promotionType.duration.getDays.toString)
-        days from receipt of subscriber ID, up to and including the day before your first payment falls due. At the end of the free trial,
+        days from receipt of subscription ID, up to and including the day before your first payment falls due. At the end of the free trial,
```
paulbrown1982: capital S

tjsilver: Good point - will do.

PR #1241, opened 2019-02-11
fix home delivery postcode validation

pvighi:
postcode validation is not working at all for home delivery subscriptions.

### DIFF
--- a/assets/javascripts/modules/checkout/deliveryDetails.js
+++ b/assets/javascripts/modules/checkout/deliveryDetails.js
@@ -134,8 +134,7 @@ define([
             }
 
             bean.on(formEls.$DELIVERY_DETAILS_SUBMIT[0], 'click', handleBillingVisibility);
-
-            if (formEls.DELIVERY.$POSTCODE_CONTAINER.data('validate-for') === 'delivery') {
+            if (formEls.DELIVERY.$POSTCODE_CONTAINER.data('validate-for') === 'delivery.address') {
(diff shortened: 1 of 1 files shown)


assets/javascripts/modules/checkout/deliveryDetails.js
```diff
@@ -134,8 +134,7 @@ define([
             }
 
             bean.on(formEls.$DELIVERY_DETAILS_SUBMIT[0], 'click', handleBillingVisibility);
-
-            if (formEls.DELIVERY.$POSTCODE_CONTAINER.data('validate-for') === 'delivery') {
+            if (formEls.DELIVERY.$POSTCODE_CONTAINER.data('validate-for') === 'delivery.address') {
```
pvighi: Is this the correct way to fix this ?
i see that a similar change was done recently for [vouchers](https://github.com/guardian/subscriptions-frontend/pull/1224/files)
After these changes the home delivery validation works for there is no voucher validation at all. I know the postcode whitelist should not be applied but it doesn't even check that the string entered is a real postcode.
is this the expected behaviour @paulbrown1982 ?

paulbrown1982: Yeah this is the correct way to fix. My GW Gifting work nested the name prefix one level deeper for delivery address, and I missed updating this selection.

pvighi:
[fix postcode validation in subscriptions frontend](https://trello.com/c/TYzPl2ei/193-fix-postcode-validation-in-subscriptions-frontend)

pvighi:
the post deploy test fails for digipack paid by direct debit, but testing in UAT this seems to work fine.

PR #1242, opened 2019-02-12
fix postcode address lookup

pvighi:
Co-Authored-By: Tom Richards <twrichards@users.noreply.github.com>

Postcode lookup was only working for the first postcode entered.
While fixing it we noticed that the react component in charge of this didn't need any state so we refactored it.

### DIFF
--- a/assets/javascripts/modules/react/addressChooser.jsx
+++ b/assets/javascripts/modules/react/addressChooser.jsx
@@ -2,7 +2,9 @@ import React from 'react'
 import PropTypes from 'prop-types';
 
 function chooseAddress(addressSection, addressStr) {
-    if (!addressSection) { return; }
+    if (!addressSection) {
+        return;
+    }
     var address = addressStr.split(', ');
     if (address.length === 7) {
         addressSection.$ADDRESS1.val((address[0] + ' ' + address[1]).trim());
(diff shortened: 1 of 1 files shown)


assets/javascripts/modules/react/addressChooser.jsx
```diff
@@ -12,55 +14,37 @@ function chooseAddress(addressSection, addressStr) {
...
+
+export default ChooseAddress;
 
-    render() {
-        let createItem = (item, key) => {
-            return <option key={key} value={item.value}>{item.name}</option>;
-        };
-        return (
-            <select className="select select--wide" onChange={this.handleChange} value={this.state.value}>
-                {this.options.map(createItem)}
```
pvighi: the cause of the bug was that this.options was never updated outside the constructor

paulbrown1982: Has this bug always been there then?

pvighi: I don't know when it was introduced but it doesn't seem to be related to the validation bug we fixed yesterday

pvighi:
[Fix home delivery address lookup on subscriptions frontend](https://trello.com/c/NFRaaBxj/194-fix-home-delivery-address-lookup-on-subscriptions-frontend)

PR #1243, opened 2019-02-15
Update newspaperLegalTerms.scala.html

SarahLaughton:
Bullet point about staff not being eligible for the print offer has been removed.


paulbrown1982:
@SarahLaughton do you want to update the neighbouring /digitalpackLegalTerms.scala.html file too?

SarahLaughton:
Yep, just double-checked with Charles and he's said it needs to go from there too. I'll do it now :)

PR #1246, opened 2019-02-18
reveal Guardian Weekly Price Rise on /manage - only within 30days

twrichards:
To coordinate with the Guardian Weekly Price Rise letters, this PR ensures that the price rise is only visible online for 30 days in advance, and that the ugly internal name for the product rate plan is no longer exposed.

| Before | After |
| --- | --- |
| image | image | 

If a price rise is inferred (product rate plans don't match on future plans) and the price rise is due more than 30 days ahead, in addition to hiding the 'Upcoming Plan', the current plan is passed into the `billingSchedule` and is then compared with each bill amount, if the bill amount is larger than the currentPlan amount then the current plan amount is used instead, which should support the temporary discounts without revealing the price rise before the 30 days

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,7 @@ By default, the setup script will hash file assets and generate a `conf/assets.m
 which in turn will cause Play to render assets with their hashed path. Use the `grunt compile --dev`
 task in order to have Play to render assets without hashing them.
 
-Finally, the grunt file provides a
-convenient `watch` task, which will dynamically recompile assets as they get edited.
+Alternatively, run `grunt watch --dev`, to dynamically recompile assets as they get edited.
 
 ## Test execution
 
--- a/app/model/SubscriptionOps.scala
+++ b/app/model/SubscriptionOps.scala
@@ -66,6 +66,14 @@ object SubscriptionOps extends LazyLogging {
     def oneOffPlans = subscription.plans.list.filterNot(p => p.end.isBefore(now) || p.charges.billingPeriod.isRecurring)
     def introductoryPeriodPlan = oneOffPlans.find(_.charges.billingPeriod == SixWeeks)
     def hasIntroductoryPeriod = introductoryPeriodPlan.isDefined
+
+    def shouldHideUpcomingPlan = (currentPlans.headOption, futurePlans.headOption) match {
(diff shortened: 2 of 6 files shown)


jacobwinch:
> the billing schedule uses the price of the first invoice for all the other invoices (this is messy but since it's limited to Guardian Weekly it should be fine)

I've been thinking about this a little more... I doubt there will be many real examples of this, but if there is a temporary discount on the subscription this display could be a little misleading. As an alternative, we could hide all scheduled payments that are >= price rise date?

twrichards:
> > the billing schedule uses the price of the first invoice for all the other invoices (this is messy but since it's limited to Guardian Weekly it should be fine)
> 
> I've been thinking about this a little more... I doubt there will be many real examples of this, but if there is a temporary discount on the subscription this display could be a little misleading. As an alternative, we could hide all scheduled payments that are >= price rise date?

@jacobwinch my only fear with that is that, as the price rise approaches the billing schedule will have only one or two entries, implying that their subscription is coming to an end (imagine it on the screenshot above) this would be misleading for a lot of users rather than only a few in the case of temp. discounts.
I will see if I can do some sort of Hybrid...

twrichards:
> > > the billing schedule uses the price of the first invoice for all the other invoices (this is messy but since it's limited to Guardian Weekly it should be fine)
> > 
> > 
> > I've been thinking about this a little more... I doubt there will be many real examples of this, but if there is a temporary discount on the subscription this display could be a little misleading. As an alternative, we could hide all scheduled payments that are >= price rise date?
> 
> @jacobwinch my only fear with that is that, as the price rise approaches the billing schedule will have only one or two entries, implying that their subscription is coming to an end (imagine it on the screenshot above) this would be misleading for a lot of users rather than only a few in the case of temp. discounts.
> I will see if I can do some sort of Hybrid...

@jacobwinch I've improved how this works... now the current plan is passed into the `billingSchedule` if an upcoming rate plan change is more than 30days away, and is then compared with each bill amount, if the bill amount is larger than the currentPlan amount then the current plan amount is used instead, which should support the temporary discounts without revealing the price rise before the 30 days

jacobwinch:
New approach to handling the billing schedule sounds good 👍

PR #1247, opened 2019-02-25
Use latest membership common to avoid MatchError

jacobwinch:
Pulls in the fix added here: https://github.com/guardian/membership-common/pull/589/files#diff-2d005ddfcd75abd06f447d0a030cc0afR71


jacobwinch:
Thanks @paulbrown1982 & @lmath!

PR #1256, opened 2019-03-22
Remove paper delivery UI

walaura:
Removes the switch to paper delivery buttons and redirects the delivery URLs

Screenshot 2019-03-22 at 09 04 50

[Trello](https://trello.com/c/22A6TAOb/2314-print-home-delivery-old-checkout-switch-off)

### DIFF
--- a/app/controllers/Shipping.scala
+++ b/app/controllers/Shipping.scala
@@ -50,11 +50,11 @@ class Shipping(touchpointBackend: TouchpointBackend, commonActions: CommonAction
   }
 
   def viewDeliveryPaperDigital() = NoCacheAction {
-    Redirect("/collection/paper-digital")
+    Redirect("/")
   }
 
   def viewDeliveryPaper() = NoCacheAction {
-    Redirect("/collection/paper")
+    Redirect("/")
   }
 
 }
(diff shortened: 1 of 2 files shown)


app/controllers/Shipping.scala
```diff
@@ -49,30 +49,12 @@ class Shipping(touchpointBackend: TouchpointBackend, commonActions: CommonAction
...
-        title = "Paper + digital home delivery subscription",
-        description =
-          """|If you live within the M25 you can have your papers delivered by 7am Monday - Saturday and 8.30am on Sunday.
-             |Plus you can start using the daily edition and premium live news app immediately.""".stripMargin,
-        altPackagePath = s"/collection/$segment",
-        options = catalog.delivery.list.toList.filter(_.charges.digipack.isDefined).map(planToOptions).sortBy(_.weeklyPrice).reverse
-      ), segment)
-    }
+  def viewDeliveryPaperDigital() = NoCacheAction {
+    Redirect("/collection/paper-digital")
```
jacobwinch: As discussed, perhaps it would be better to redirect these to the subs landing page? (If there are any links which push people straight to home delivery we don't want them to end up buying vouchers by mistake)

PR #1264, opened 2019-07-16
update versions for dependencies

johnduffell:
Updated jackson core for a snyk vulnerability.
https://app.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-450917

I have also updated sbt and play as they were both way behind.
Seems to work the tests pass (apart from the ones that seem to hit address services etc.)
and I got most of the way through a digital pack checkout, however the UAT and DEV salesforce credentials are not up to date so I couldn't actually create the subscription.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,5 @@
 import play.sbt.PlayImport
+import scala.sys.process._
 
 name := "frontend"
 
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -1,4 +1,5 @@
 import play.sbt.PlayImport
+import scala.sys.process._
```
johnduffell: needed for the sbt update, to get the `!!` function on a string.

build.sbt
```diff
@@ -73,7 +74,7 @@ libraryDependencies ++= Seq(
     "net.logstash.logback" % "logstash-logback-encoder" % "4.11",
     "com.fasterxml.jackson.dataformat" % "jackson-dataformat-cbor" % "2.9.2",
     "com.google.guava" % "guava" % "25.0-jre", //-- added explicitly - snyk report avoid logback vulnerability
-    "com.fasterxml.jackson.core" % "jackson-databind" % jacksonVersion, //added explicitly to avoid snyk vulnerability
+    "com.fasterxml.jackson.core" % "jackson-databind" % (jacksonVersion+".1"), //added explicitly to avoid snyk vulnerability
```
johnduffell: the other jackson libs didn't get the bump, only this one.  Doing it this way will force it to break once the next bump happens at compile time, which will trigger the next person to update it here.  If I hard code the version it may get missed and only changed for the others.

PR #1265, opened 2019-07-29
Authentication via identity API

Mullefa:
Identity are wanting to centralise user authentication i.e. an application should authenticate a user by making a call to the authentication endpoint in identity API (please talk to us in person for more context).

This changes the type signature of user authentication from:
```scala
RequestHeader => Option[U]
```
to:
```scala
RequestHeader => Future[Option[U]]
```
To lay the ground-work for this change, define and use the class `AsyncAuthenticatedBuilder` to support this change in type signature.  

__Notes:__
- this PR does not change the authentication mechanism used; authenticating via the identity API will be done in a subsequent PR
- apologies for the savage diffs in this PR; refactoring was quite awkward in some places since the ubiquitous:
    ```scala
    implicit val resolution: TouchpointBackend.Resolution = touchpointBackends.forRequest(...)
    implicit val tpBackend: TouchpointBackend = resolution.backend
    ```
    is now asynchronous (`forRequest()` relies on authenticating a user); I will annotate some of the diffs to hopefully make them easier to digest.

PR #1266, opened 2019-07-30
user models from identity-model rather than identity-play-auth

Mullefa:
#1265 introduces the types for authentication using the identity API; in a subsequent PR [`identity-auth-play`](https://github.com/guardian/identity/tree/master/identity-auth-play/src) will be used to implement this authentication. 

Since this library will remove the user models currently being used, in favour for the user models in `identity-model`, I have split this out into a separate PR since there were [issues](https://github.com/guardian/support-frontend/pull/1984) when I previously tried to make analogous changes in support frontend in a single PR.

### DIFF
--- a/app/model/SubscriptionData.scala
+++ b/app/model/SubscriptionData.scala
@@ -86,7 +86,7 @@ object PersonalData {
       first = u.privateFields.firstName.mkString,
       last = u.privateFields.secondName.mkString,
       email = u.primaryEmailAddress,
-      receiveGnmMarketing = false,
+      receiveGnmMarketing = false, // Deprecated, unused and not a GDPR compliant consent
       address = u.billingAddress,
       telephoneNumber = phoneNumber
     )
--- a/app/model/SubscriptionData.scala
+++ b/app/model/SubscriptionData.scala
@@ -1,7 +1,8 @@
 package model
 import com.gu.i18n.Country.UK
 import com.gu.i18n.{CountryGroup, Currency, Title}
-import com.gu.identity.play.IdUser
+import com.gu.identity.model.{User => IdUser}
+import com.gu.memsub.Subscription.ProductRatePlanId
 import com.gu.memsub._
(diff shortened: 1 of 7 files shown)


app/model/SubscriptionData.scala
```diff
@@ -69,23 +70,23 @@ case class PaperData(
...
     val personalData = PersonalData(
-      title = u.privateFields.flatMap(_.title).flatMap(Title.fromString),
-      first = u.privateFields.flatMap(_.firstName).mkString,
-      last = u.privateFields.flatMap(_.secondName).mkString,
+      title = u.privateFields.title.flatMap(Title.fromString),
+      first = u.privateFields.firstName.mkString,
+      last = u.privateFields.secondName.mkString,
       email = u.primaryEmailAddress,
-      receiveGnmMarketing = u.statusFields.flatMap(_.receiveGnmMarketing).getOrElse(false),
+      receiveGnmMarketing = false,
```
johnduffell: just to clarify, this change is related to the whole thing about the field that we're reading from isn't actually set from anywhere anyway, so it was already effectively hard coded to false

johnduffell:
great, thanks for splitting out the changes in to smaller PR! much easier to review and lower risk (and easier to roll back!)
👍

app/model/SubscriptionData.scala
paulbrown1982: It's also deprecated, and unused - it's not a GDPR-compliant consent. So in theory, it should be null, or defaulted within the case class constructor.

Mullefa: @johnduffell yeah - correct and I meant to put a comment but forgot. Put I think Paul's point now takes precedent, so I'll add that instead.

Mullefa:
Have tested locally and on code and all seems to be working - I was able to purchase a number of products as both a guest and registered user. I'll merge now and keep and eye on the post merge production tests.

PR #1267, opened 2019-07-31
Use dev-nginx to configure Nginx for subscriptions-frontend locally

Mullefa:
Thanks @akash1810 for talking me through this; really interesting to understand how the new approach works. 

The updated setup script has been run locally; can confirm that it's working.


README.md
```diff
@@ -1,27 +1,19 @@
 # Subscriptions frontend
 
-## NGinx
+## Nginx
 
-#### Setup Nginx for `Identity-Platform`
+#### Setup Nginx for identity-platform
 
 Subscriptions depends on Nginx config and SSL certs from Identity, so you'll need to perform the
```
akash1810: I think we can remove this block once all projects have been updated.

PR #1274, opened 2019-08-01
Synchronous authentication via identity API

Mullefa:
#1265 introduced authentication via the identity API; the `Future` data type was introduced to handle asynchronous network requests. However, this resulted in a PR with a very complicated diff. This PR should be seen as an alternative to #1265, which keeps user authentication synchronous by executing blocking requests on a dedicated thread pool. Reviewers free to compare and contrast both approaches, but my preference would be for this PR to be the one that's merged. Reasons for this preference have been added as inline code comments; I've paraphrased them below for convenience:

It was deemed that the risk of a bug being introduced in #1265 was greater than the risk of a user getting a slower than normal response (as a result of all threads available to execute this method call being blocked). The latter was considered low risk, since the subscriptions-frontend app is low traffic and a dedicated thread pool of 30 threads has been reserved to execute these blocking calls (c.f. `Authentication.unsafeInit()`). The fact that subscriptions-frontend is a deprecated application also factored into the decision.


app/controllers/CheckoutHandler.scala
```diff
@@ -30,7 +29,7 @@ import views.support.{BillingPeriod => _}
 
 import scala.concurrent.{ExecutionContext, Future}
 
-class CheckoutHandler(fBackendFactory: TouchpointBackends, commonActions: CommonActions, implicit val executionContext: ExecutionContext, override protected val controllerComponents: ControllerComponents) extends BaseController with LazyLogging {
+class CheckoutHandler(fBackendFactory: TouchpointBackends, authenticationService: AuthenticationService, testUsers: TestUsers, commonActions: CommonActions, implicit val executionContext: ExecutionContext, override protected val controllerComponents: ControllerComponents) extends BaseController with LazyLogging {
```
johnduffell:
```suggestion
class CheckoutHandler(
  fBackendFactory: TouchpointBackends,
  authenticationService: AuthenticationService,
  testUsers: TestUsers,
  commonActions: CommonActions,
  implicit val executionContext: ExecutionContext,
  override protected val controllerComponents: ControllerComponents
) extends BaseController with LazyLogging {
```

app/services/AuthenticationService.scala
```diff
@@ -1,8 +1,87 @@
...
+  // as changing the result type to Future resulted in a PR with a very complicated diff:
+  // https://github.com/guardian/subscriptions-frontend/pull/1265
+  //
+  // It was deemed that the risk of a bug being introduced in #1265
+  // was greater than the risk of a user getting a slower than normal response
+  // (as a result of all threads available to execute this method call being blocked).
+  // The latter was considered low risk, since the subscriptions-frontend app is low traffic
+  // and a dedicated thread pool of 30 threads has been reserved
+  // to execute these blocking calls (c.f. Authentication.unsafeInit()).
+  // The fact that subscriptions-frontend is a deprecated application also factored into the decision.
```
johnduffell: sounds very fair, even though it offends my functional purity mind 🙈 
given how limited future development this app will get, the argument in favour of maintainability is reduced a lot!

app/services/AuthenticationService.scala
```diff
@@ -1,8 +1,87 @@
...
+  def authenticatedUserFor(request: RequestHeader): Option[AuthenticatedIdUser] = {
+    identityPlayAuthService.getUserFromRequest(request)
+      .map { case (credentials, user) => buildAuthenticatedUser(credentials, user) }
+      .attempt
+      .unsafeRunTimed(limit = 5.seconds) // fairly arbitrary
```
johnduffell: is this the "time out and throw an exception" limit per single call? so longer would be more likely to block up the thread pool, and shorter would be more likely to fail when identity is going slowly?

app/services/IdentityService.scala
```diff
@@ -22,6 +24,51 @@ import scala.concurrent.{ExecutionContext, Future}
...
+object CookieBuilder {
+
+  def cookiesFromDescription(
+    cookieDescriptionList: CookieDescriptionList,
+    domain: Option[String] = None
+  )(implicit clock: Clock = Clock.systemUTC()): Seq[Cookie] = {
+
+    val maxAge = Duration.between(clock.instant(), cookieDescriptionList.expiresAt).getSeconds.toInt
+
+    for (cookieDescription <- cookieDescriptionList.values) yield {
```
johnduffell: what does `description` mean? could this be just `cookie`?  as it seems toinlcude the key as well

johnduffell: or `identityCookie` if it's the one for the ideneity side rather than the play side

app/services/IdentityService.scala
```diff
@@ -22,6 +24,51 @@ import scala.concurrent.{ExecutionContext, Future}
...
+      val isSecure = cookieDescription.key.startsWith("SC_")
+      val maxAgeOpt = if (cookieDescription.sessionCookie.contains(true)) None else Some(maxAge)
+      Cookie(
+        cookieDescription.key,
+        cookieDescription.value,
+        maxAge = maxAgeOpt,
+        secure = true, // as of https://github.com/guardian/identity-frontend/pull/196
+        httpOnly = isSecure, // ideallycd . this would come from the Cookie Description
```
johnduffell: "cd ." 😆

johnduffell:
looks good, a few comments but I think overall it's fair to do it in this slightly hacky way!
👍

Mullefa:
Thanks for reviewing John. Will address your comments. Glad you're ok with this approach cos there's another one coming for membership-frontend 😅

app/services/IdentityService.scala
Mullefa: 😂

app/services/AuthenticationService.scala
Mullefa: Yes, although given identity API's response time, I don't think this limit would ever be reached. The return type of `unsafeRunTimed()` is an option, so if the limit is reached (before a response is ready), a `None` will be returned.

app/services/IdentityService.scala
Mullefa: Code was (mainly) copy and pasted from identity-play-auth - presumably the names were influenced by the types of the values they refer to 🤷‍♂️ 

I think given the names of the types, the variable names are reasonable, but they could be another change to make in (c.f.) identity 4.x

I'll put a comment though to make the code clearer.

Mullefa:
Tested on CODE

PR #1277, opened 2019-08-28
Update subs checkout to have more home delivery postcodes

arausuy:
As a result of changing suppliers we now have a wider delivery zone.

This is to align subscriptions-frontend with the new postcodes in guardian/support-frontend#2038


arausuy:
Sorry! its postcodes though :'(

PR #1279, opened 2019-10-10
snyk: update jackson and scala versions

johnduffell:
good old snyk.
tested on local and seems to work! checked out a guardian weekly.  Hopefully it will work in prod.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,11 +37,11 @@ lazy val root = (project in file(".")).enablePlugins(
       "com.gu.memsub.Subscription.ProductRatePlanId"
   ))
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -37,11 +37,11 @@ lazy val root = (project in file(".")).enablePlugins(
       "com.gu.memsub.Subscription.ProductRatePlanId"
   ))
 
-scalaVersion := "2.12.6"
+scalaVersion := "2.12.10"
```
johnduffell: this was not required by snyk, but it's always a good idea

PR #1282, opened 2019-10-17
Final copy change for DP rename

rupertbates:
Final copy change for DP rename


app/controllers/Shipping.scala
```diff
@@ -30,7 +30,7 @@ class Shipping(touchpointBackend: TouchpointBackend, commonActions: CommonAction
     catalog.map { catalog =>
       index(CollectionSubscriptionProduct(
         title = "Paper + digital voucher subscription",
-        description = "Save money on your newspapers and digital content. Plus start using The Guardian Daily and premium access to the Live app immediately.",
+        description = "Save money on your newspapers and digital content. Plus start using The Guardian Daily and premium access to the Guardian Live app immediately.",
```
lucymonie: Sorry to be annoying but I think it's The Guardian (capital T)

PR #1284, opened 2019-11-05
Update USA & Canada's phone line opening times

arausuy:
In this PR there is a small change to the copy relating the the USA & Canada's phone lines opening times.

image

This has been done to reflect that the contact centre closes at 5pm(EST) rather than the original 6:00(EST) 

Trello Ticket: https://trello.com/c/7IZ6kisi/631-update-opening-hours-on-all-mma-platforms

### DIFF
--- a/app/views/support/ContactCentreOps.scala
+++ b/app/views/support/ContactCentreOps.scala
@@ -12,7 +12,7 @@ object ContactCentreOps {
 
   def phone(contactUsCountry: Option[Country] = None): String = {
     if (contactUsCountry exists countriesHandledByUSCallCentre)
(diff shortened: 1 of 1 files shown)


app/views/support/ContactCentreOps.scala
```diff
@@ -12,7 +12,7 @@ object ContactCentreOps {
 
   def phone(contactUsCountry: Option[Country] = None): String = {
     if (contactUsCountry exists countriesHandledByUSCallCentre)
-      s"1-844-632-2010 (toll free); ${directLine(contactUsCountry)} (direct line). Lines are open 9:15am-6pm, Monday to Friday (EST/EDT)"
```
twrichards: is `EDT` no longer a thing?

arausuy: Although it is missing from the copy I think its pragmatic to add it - we have BST for the UK number

PR #1285, opened 2019-11-28
Bump membership-common to support GW 3 month plan

twrichards:
Similar to https://github.com/guardian/members-data-api/pull/413 we want to support the **new GW 3 month plan** on this legacy platform since some customers will have received links in welcome emails and/or already be familiar with `/manage` for viewing these subs.


twrichards:
Closing since the UI doesn't currently respect fixed-term and so displays the new 3month fixed term subs as indefinite (which we've decided is worse than not being able to log in).
image
**There will be some following work to correctly respect fixed-term which will likely branch off this.**

PR #1286, opened 2019-12-16
New three month plan

rupertbates:
We have added a new three month fixed term Guardian Weekly plan which is available to buy as a gift. This PR updates the /manage page to handle renewals for these subs.

sub thegulocal com_manage

### DIFF
--- a/app/model/ContentSubscriptionPlanOps.scala
+++ b/app/model/ContentSubscriptionPlanOps.scala
@@ -2,7 +2,7 @@ package model
 
 import com.gu.i18n.Currency._
 import com.gu.i18n._
-import com.gu.memsub.BillingPeriod.{OneYear, Quarter, SixWeeks, Year}
+import com.gu.memsub.BillingPeriod.{OneYear, Quarter, SixWeeks, ThreeMonths, Year}
 import com.gu.memsub.Product
 import com.gu.memsub.subsv2.CatalogPlan
 import model.BillingPeriodOps._
@@ -56,7 +56,8 @@ object ContentSubscriptionPlanOps {
     }
 
     def availableForCheckout: Boolean = in.charges.billingPeriod.isRecurring || in.charges.billingPeriod == SixWeeks
-    def availableForRenewal: Boolean = in.charges.billingPeriod == Quarter || in.charges.billingPeriod == OneYear || in.charges.billingPeriod == Year
+
+    def availableForRenewal: Boolean = in.charges.billingPeriod == Quarter || in.charges.billingPeriod == OneYear || in.charges.billingPeriod == Year || in.charges.billingPeriod == ThreeMonths
   }
 
 }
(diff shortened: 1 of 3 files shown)


app/model/ContentSubscriptionPlanOps.scala
```diff
@@ -2,7 +2,7 @@ package model
 
 import com.gu.i18n.Currency._
 import com.gu.i18n._
-import com.gu.memsub.BillingPeriod.{OneYear, Quarter, SixWeeks, Year}
+import com.gu.memsub.BillingPeriod.{OneYear, Quarter, SixWeeks, ThreeMonths, Year}
```
lucymonie: Should the finite ones be named so as to make it totally clear? Something like 'ThreeMonthsFinite' or 'ThreeMonthsGift' or 'ThreeMonthsOnly'?

app/model/ContentSubscriptionPlanOps.scala
```diff
@@ -56,7 +56,8 @@ object ContentSubscriptionPlanOps {
     }
 
     def availableForCheckout: Boolean = in.charges.billingPeriod.isRecurring || in.charges.billingPeriod == SixWeeks
-    def availableForRenewal: Boolean = in.charges.billingPeriod == Quarter || in.charges.billingPeriod == OneYear || in.charges.billingPeriod == Year
+
+    def availableForRenewal: Boolean = in.charges.billingPeriod == Quarter || in.charges.billingPeriod == OneYear || in.charges.billingPeriod == Year || in.charges.billingPeriod == ThreeMonths
```
jfsoul: Consider `List(Quarter, OneYear, Year, ThreeMonths) contains in.charges.billingPeriod`?

app/model/Email.scala
```diff
@@ -157,7 +157,7 @@ object PaperFieldsGenerator {
                  promotionDescription: Option[String]
                ): Seq[(String, String)] = {
     val paymentFields = paymentMethod match {
-      case GoCardless(mandateId, accName, accNumber, sortCode) => Seq(
+      case GoCardless(mandateId, accName, accNumber, sortCode, _, _) => Seq(
```
lucymonie: What are the underscores for? Are they wildcard arguments?

app/model/ContentSubscriptionPlanOps.scala
rupertbates: That would be a massive change unfortunately as these are defined in membership-common which is a separate library and is consumed by a lot of other projects

app/model/Email.scala
rupertbates: It's a match pattern so we're saying that we want to match a case where it is of type `GoCardless` and we want to capture the the first four fields of that object to named variables which we can use below, but we don't care about the last two fields because we're not going to use them.

app/model/ContentSubscriptionPlanOps.scala
lucymonie: 👍

app/model/Email.scala
lucymonie: Ah, thanks!

PR #1287, opened 2019-12-18
Remove Krux as we no longer use it.

paulbrown1982:
I've also searched Github for use of string `Jglpp88U` and cannot find it in any active app, just a very old legacy version of this site.


paulbrown1982:
Tested in CODE.

PR #1297, opened 2020-04-21
Remove extra trailing slash on identity base

jfsoul:
Neither CODE nor DEV have a trailing slash, so this seems fairly safe - but anyway the motivation for this is that Identity appear to be rejecting requests with the extra slash with a 404, which is new.

### DIFF
--- a/conf/PROD.public.conf
+++ b/conf/PROD.public.conf
@@ -4,7 +4,7 @@ include "touchpoint.UAT.conf"
 stage="PROD"
 
 identity {
-  baseUri = "https://idapi.theguardian.com/"
+  baseUri = "https://idapi.theguardian.com"
   production.keys=true
   webapp.url="https://profile.theguardian.com"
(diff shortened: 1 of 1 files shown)


johnduffell:
re overridden in private conf, the only things are:
api key
test.users.secret

johnduffell:
we always add a slash - this file: https://github.com/guardian/subscriptions-frontend/blob/bfaccbdc825c6863585bd156963f6aa29c2440c8/app/services/IdentityService.scala#L236

Also we pass into the IdentityPlayAuthService which uses http url builder to put the path together with the base url so that should cope alright.  That would affect things you need to be logged into.

paulbrown1982:
re overridden in private conf - I just did that and redeployed not knowing about this PR, sorry! So please remove that line just before you merge this!

johnduffell:
thanks @paulbrown1982  you must have done it the moment after I checked! I will do as you suggest, I was thinking the same.

PR #1305, opened 2020-09-07
TCFv2 - privacy settings link in footer

i-hardy:
## What does this change?
Adds a link to the footer which surfaces the privacy settings modal, allowing the user to adjust their privacy preferences at any time.

## How to test
Run the site as normal. With JS enabled, a 'Privacy Settings' link should be visible in the footer, and should open the privacy settings manager modal

## Images

**Mobile**
Screenshot 2020-09-07 at 15 14 56

**Tablet**
Screenshot 2020-09-07 at 15 15 41

**Desktop**
Screenshot 2020-09-07 at 15 12 52


assets/javascripts/modules/analytics/cmp.js
```diff
@@ -34,4 +37,10 @@ const checkCCPA = () => new Promise((resolve) => {
 
 const registerCallbackOnConsentChange = (fn) => onConsentChange(fn);
 
-export { getConsentForVendors, checkAllTCFv2PurposesAreOptedIn, checkCCPA, registerCallbackOnConsentChange };
+const createPrivacySettingsLink = () => {
+    const privacySettingsButton = document.querySelector(PRIVACY_SETTINGS_SELECTOR);
+    privacySettingsButton.classList.remove(HIDDEN_CLASS);
```
GHaberis, line 42: Can we wrap the `classList.remove` and `addEventListener` calls in a conditional `if (privacySettingsButton) { ... }`? Just in case someone removes the link from the template, or updates the class names...

assets/javascripts/modules/analytics/setup.js
```diff
@@ -44,7 +44,10 @@ define([
          * GA and other tags often prefer Ophan to have bootstrapped and set window.guardian state,
          * so load them after Ophan has loaded if thirdPartyTrackingEnabled.
          */
-        ophan.loaded.finally(() => cmp.registerCallbackOnConsentChange(loadGA));
+        ophan.loaded.finally(() => {
+            cmp.createPrivacySettingsLink();
```
GHaberis, line 48: I think this `cmp.createPrivacySettingsLink();` would be better placed in the `main.js` file, after we've initiliased the cmp: https://github.com/guardian/subscriptions-frontend/blob/f85be04da5f3bd60eb5a97683e173d8ad80747ee/assets/javascripts/main.js#L56

i-hardy: 👍 Both changes made, thanks!

assets/javascripts/modules/analytics/cmp.js
```diff
@@ -34,4 +37,12 @@ const checkCCPA = () => new Promise((resolve) => {
 
 const registerCallbackOnConsentChange = (fn) => onConsentChange(fn);
 
-export { getConsentForVendors, checkAllTCFv2PurposesAreOptedIn, checkCCPA, registerCallbackOnConsentChange };
+const createPrivacySettingsLink = () => {
```
GHaberis, line 40: I'm not sure `createPrivacySettingsLink` should remain in the `analytics.js` module now we're calling it from `main.js`, I think it makes sense to move the entire function into `main.js`. `analytics.js` is primarily concerned with making tracking-analytics calls.

i-hardy: Makes sense, we have the CMP module in scope there too. I'll make the change.

GHaberis (approved):
Looks good 👍

PR #1302, opened 2020-08-06
TCFv2 CMP

GHaberis:
## Why are you doing this?

This PR enabled the TCFv2 compliant CMP for all non-US users, this is implemented using https://www.npmjs.com/package/@guardian/consent-management-platform.

[**Trello Card**](https://trello.com/c/zmyZDXp9/3243-upgrade-to-tcfv2-cmp-on-membershiptheguardiancom)

## Changes

- Add `@guardian/consent-management-platform`  v4 (TCFv2 compliant support) dependency.
- Update `main.js` to call `cmp.init` using the new interface exposed by `@guardian/consent-management-platform`  v4.
- Update `thirdPartyTrackingConsent.js` to use the new `onConsentChange` function exposed by `@guardian/consent-management-platform`.  `onConsentChange` takes a callback that it executes with consent state, passed as the parameter `state`.  The value of `state` will differ when in CCPA mode (US users) and TCFv2 mode (everyone else), so the callback determines which mode we're in, and returns either "OptedIn" or "OptedOut". There is no longer an "Unset" (null) state, so this has been removed from the code base.
- Remove logic used to read/write the `GU_TK` cookie from thirdPartyTrackingConsent.js`.
- Remove the old-CMP component (`consentBanner.scala.html`), it's associated styles.

I also had to add the following dependencies, as these are peerDepenedencies of `@guardian/consent-management-platform`, they're mostly used if rendering the old TCFv1 CMP. Unfortunately without them the build failed:

"@babel/runtime": "^7.2.0",
"@emotion/core": "^10.0.21",
"@guardian/src-button": "^2.0.0",
"@guardian/src-foundations": "^2.0.0",
"@guardian/src-icons": "^2.0.0",
"emotion-theming": "^10.0.27",
"react": "^16.10.2"

## Screenshots

tcvf2_subs_mobile

tcvf2_subs_desktop

PR #1309, opened 2020-09-21
Fix post deploy tests

GHaberis:
## What does this change?

This PR reintroduces the conditional check which prevents the CMP from showing during Post Deploy Tests.

When we switched from the old CMP to the new TCFv2 compliant CMP https://github.com/guardian/subscriptions-frontend/pull/1302 we removed the logic that prevents the CMP from showing during Post Deploy Tests. Preventing the CMP from showing is necessary for the tests to proceed.

### DIFF
--- a/assets/javascripts/main.js
+++ b/assets/javascripts/main.js
@@ -1,5 +1,6 @@
 require([
     'utils/ajax',
+    'utils/cookie',
     'modules/raven',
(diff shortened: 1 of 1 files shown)


rupertbates (approved):
Well done for picking this up!

i-hardy (approved):
👍

PR #1311, opened 2020-10-21
Bump membership-common to v0.589

kelvin-chappell:
This is to support the new Patron reader type: guardian/membership-common#633

There's been a namespace change in one of the transitive dependencies since v0.553.
`io.lemonlabs.uri.typesafe.dsl._` operators build `Uri` instances rather than Strings.  But the `toString` implementation seems to do the expected thing.

### DIFF
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -1,13 +1,12 @@
 package configuration
 
-
 import com.github.nscala_time.time.Imports._
 import com.gu.cas.PrefixedTokens
 import com.gu.config._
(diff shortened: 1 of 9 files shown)


rupertbates (approved):
Is this just to keep Subs up to date? I can't see any reason why it needs to know about Patrons

kelvin-chappell:
> Is this just to keep Subs up to date? I can't see any reason why it needs to know about Patrons

@rupertbates Yes it's just to keep it on latest version in this case.

PR #1319, opened 2021-01-15
Remove deprecated IDAPI user model fields 

QuarpT:
## What does this change?

- Remove deprecated IDAPI user model fields - these fields are now represented by items in the consents array on the user model.

## How can we measure success?

Avoid confusion of unused deprecated fields.

### DIFF
--- a/app/services/IdentityService.scala
+++ b/app/services/IdentityService.scala
@@ -201,9 +201,8 @@ object PersonalDataJsonSerialiser {
           "postcode" -> addr.postCode,
           "country" -> addr.country.fold(addr.countryName)(_.name)
         )
-      }),
-      "statusFields" ->
-        Json.obj("receiveGnmMarketing" -> personalData.receiveGnmMarketing))
+      })
+     )
   }
 }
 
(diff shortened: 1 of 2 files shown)


app/services/IdentityService.scala
```diff
@@ -201,9 +201,8 @@ object PersonalDataJsonSerialiser {
           "postcode" -> addr.postCode,
           "country" -> addr.country.fold(addr.countryName)(_.name)
         )
-      }),
```
QuarpT, line 204: Do we need to update the IDAPI consent array instead?
I noticed we're still getting `receiveGnmMarketing` coming through in a POST form request and being set in SalesForce.

johnduffell: good spot, since this subs frontend project is going away "soon" I would assume it's not a big issue.
Only paper+digital, and assisted sale would go through this route.  Paper+digital will be removed soon.  So if we miss out on consents via this route I don't think it would be a significant issue.

johnduffell (approved):
thanks for this 👏

PR #1329, opened 2021-04-20
stop using bintray for acquisition-event-producer

johnduffell:
This PR stops using the above library from bintray.  This is needed becuase bintray will go away on 1st may and we still need to be able to build this project.

I have chosen to build a copy locally from support-frontend, hacking it to produce a scala 2.12 version (as support repo is on scala 2.13 now)
I then checked that version into github as an unmanaged dependency (all 40MB of it!), because it seemed easier than trying to publish to sonatype.  If subscribe was a supported repo that would be actively developed, I might have tried other ways, but hopefully this will be the last version bump.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,10 +37,10 @@ lazy val root = (project in file(".")).enablePlugins(
       "com.gu.memsub.Subscription.ProductRatePlanId"
   ))
 
-scalaVersion := "2.12.10"
+scalaVersion := "2.12.13"
 scalacOptions ++= Seq("-feature")
 
-val scalatestVersion = "3.0.4"
+val scalatestVersion = "3.1.1"
 val jacksonVersion = "2.10.0"
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -37,10 +37,10 @@ lazy val root = (project in file(".")).enablePlugins(
       "com.gu.memsub.Subscription.ProductRatePlanId"
   ))
 
-scalaVersion := "2.12.10"
+scalaVersion := "2.12.13"
```
johnduffell, line 40: opportunistic version bump

build.sbt
```diff
@@ -37,10 +37,10 @@ lazy val root = (project in file(".")).enablePlugins(
...
 scalacOptions ++= Seq("-feature")
 
-val scalatestVersion = "3.0.4"
+val scalatestVersion = "3.1.1"
```
johnduffell, line 43: I got MethodNotFound errors because there was a version of scalatest imported from the event producer, and it clashed with this old one.  Now I have matched the versions, so it can work.

build.sbt
```diff
@@ -56,13 +56,13 @@ libraryDependencies ++= Seq(
...
     "com.gu" %% "content-authorisation-common" % "0.4",
     "com.gu" %% "tip" % "0.1.1",
-    "com.gu" %% "acquisition-event-producer-play26" % "4.0.26",
     "com.github.nscala-time" %% "nscala-time" % "2.16.0",
     "io.sentry" % "sentry-logback" % "1.7.5",
     "com.typesafe.scala-logging" %% "scala-logging" % "3.7.2",
     "org.scalatest" %% "scalatest" % scalatestVersion % "test",
     "org.scalactic" %% "scalactic" % scalatestVersion % "test",
-    "org.seleniumhq.selenium" % "selenium-java" % "3.7.1" % "test",
+    "org.scalatestplus" %% "selenium-3-141" % (scalatestVersion + ".0") % "test",
```
johnduffell, line 64: the 3.141 has to match the `selenium-java` version, and the actual version needs to match scalatest, with a point release appended (which happens to be `0`)

rupertbates (approved):
Good work 👍

PR #1333, opened 2021-05-05
pull in the whole acquisition event producer

johnduffell:
We've had quite a few goes at adding the acquisition event producer jar as an unmanaged dependency (due to the bintray switchoff) however this causes the instances to cycle as the dynamodb2 classes are somehow not found.

Since we have no reall hypothesis on why this is happening, this is a copy and paste job to resolve this.


rupertbates (approved):
🥳

PR #1336, opened 2021-05-07
Bump lodash and ua-parser sub dependencies

i-hardy:
## What does this change?
This fixes the two current high severity security issues in our JS dependencies, and all but two medium severity issues. Of the remaining issues, one doesn't have a possible resolution at the moment, and the other would require a major version upgrade of `node-fetch`, which is a sub-dependency that we're not actually using as there's no server-side JS.


rupertbates (approved):
👍

PR #1340, opened 2021-06-15
Remove Tip verification experiment

jacobwinch:
## What does this change?

Removes some experimental monitoring that was added in https://github.com/guardian/subscriptions-frontend/pull/901.

> Note that this requires us to add a GitHub access token to our private .conf file. For the purposes of testing out TiP, I suggest that we use my personal access token for this (temporarily).

Unfortunately this is currently still the case, so I'd like to remove this functionality and tidy things up!

## How to test

I've [deployed this to `CODE`](https://riffraff.gutools.co.uk/deployment/view/5a501b21-e587-4229-98c5-26fabea0b3e0) and confirmed that I can still buy a Digital Pack / reach the Thank You page.

## How can we measure success?

* We'll be able to delete a personal access token
* This repository will have less dependencies

## Have we considered potential risks?

Before deploying this change, I will also remove some private configuration related to TiP. This is a `PROD`-only operation, which takes place outside of version control, so it's a bit risky. After deploying I will buy something as a test user just to be on the safe side.

## To do

- [x] Remove private configuration

### DIFF
--- a/app/controllers/Checkout.scala
+++ b/app/controllers/Checkout.scala
@@ -17,7 +17,6 @@ import com.gu.memsub.subsv2.{CatalogPlan, PlansWithIntroductory}
 import com.gu.memsub.{Product, SupplierCode}
 import com.gu.monitoring.SafeLogger
 import com.gu.monitoring.SafeLogger._
-import com.gu.tip.Tip
 import com.gu.zuora.soap.models.errors._
 import com.typesafe.scalalogging.LazyLogging
 import configuration.Config.Identity.webAppProfileUrl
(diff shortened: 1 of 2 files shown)


johnduffell (approved):
I was just reading something on chat about tip the other day, and thinking it's just the kind of thing we decide we need when we have an incident retro.  It lets us move fast and break things more easily.  Then the rest of the time, we don't care enough to maintain it (this was in the context of membership-workflow where it hadn't been kept up to date)  This is really where having a SETI in the team helped to advocate this kind of thing!

jacobwinch:
> ❌ Testing in PROD failed! [Details](https://travis-ci.org/github/guardian/subscriptions-frontend/builds/774624968?utm_source=github_status&utm_medium=notification)

I have tested this manually and it looks like these tests have been failing for some time, so I don't think we need to be concerned about this.

PR #1345, opened 2021-08-16
Upgrade play from 2.6 to 2.8

johnduffell:
This PR updates play from 2.6 to 2.8.  This means we will be on a supported version again.

### DIFF
--- a/app/MyApplicationLoader.scala
+++ b/app/MyApplicationLoader.scala
@@ -31,7 +31,7 @@ class MyApplicationLoader extends ApplicationLoader {
       new MyComponents(context).application
     } catch {
       case e: Throwable =>
-        SafeLogger.error(scrub"Failed to load application due to $e")
+        SafeLogger.error(scrub"Failed to load application due to", e)
         throw e
     }
   }
--- a/app/configuration/Config.scala
+++ b/app/configuration/Config.scala
@@ -17,8 +17,6 @@ object Config {
   val appName = "subscriptions-frontend"
   val config = ConfigFactory.load()
 
-  val playSecret = config.getString("play.crypto.secret")
-
   lazy val googleGroupChecker = googleGroupCheckerFor(config)
 
   val stage = config.getString("stage")
--- a/build.sbt
+++ b/build.sbt
@@ -61,7 +61,6 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-databind" % "2.10.5.1"
 )
 libraryDependencies ++= Seq(
-    cache,
     ws,
     filters,
     jodaForms,
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -1,15 +1,6 @@
 # This is the main configuration file for the application.
 # ~~~~~
 
-# Secret key
-# ~~~~~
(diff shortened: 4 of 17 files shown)


conf/routes
```diff
@@ -6,15 +6,15 @@ GET         /robots.txt                      controllers.CachedAssets.at(path="/
 GET         /favicon.ico                     controllers.CachedAssets.busted(path="images/favicons/32x32.ico")
 
 # Editions homepage
-GET         /                                controllers.Homepage.index
+GET         /                                controllers.Homepage.index()
```
johnduffell, line 9: I have added the brackets as per https://github.com/guardian/support-frontend/pull/3122#discussion_r653618017
allthough I can't remember the exact reason I went for adding the brackets here in place of removing them at the point of use.

app/MyApplicationLoader.scala
```diff
@@ -31,7 +31,7 @@ class MyApplicationLoader extends ApplicationLoader {
       new MyComponents(context).application
     } catch {
       case e: Throwable =>
-        SafeLogger.error(scrub"Failed to load application due to $e")
+        SafeLogger.error(scrub"Failed to load application due to", e)
```
johnduffell, line 34: this didn't provide a stacktrace before, so when it failed , I couldn't tell it was failing while loading the play config.

app/configuration/Config.scala
```diff
@@ -17,8 +17,6 @@ object Config {
   val appName = "subscriptions-frontend"
   val config = ConfigFactory.load()
 
-  val playSecret = config.getString("play.crypto.secret")
```
johnduffell, line 20: deleted because it was unused and causing an exception on startup

conf/application.conf
```diff
@@ -1,15 +1,6 @@
...
 # ~~~~~
 
-# Secret key
-# ~~~~~
-# The secret key is used to secure cryptographics functions.
-#
-# This must be changed for production, but we recommend not changing it in this file.
-#
-# See http://www.playframework.com/documentation/latest/ApplicationSecret for more details.
-play.crypto.secret="Is]F;gOuRS]X[eTv4j9k;_jC3Tc>2Xv<D_On/CRr/MBoJAM5yBB_w2Fp]1c7xDpX"
```
johnduffell, line 11: there was no need to have this default value, in fact it makes it more likely to accidentally use this (public) value in PROD. 
So i have removed and added the correctly named one for play 2.8 in the relevant secret files.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.2
+sbt.version=1.5.5
```
johnduffell, line 1: always good to keep sbt bumped regularly

build.sbt
```diff
@@ -61,7 +61,6 @@ libraryDependencies ++= Seq(
   "com.fasterxml.jackson.core" % "jackson-databind" % "2.10.5.1"
 )
 libraryDependencies ++= Seq(
-    cache,
```
johnduffell, line 64: doesn't seem to exist any more, but removing it doesn't break anything obvious

tomrf1 (approved):

PR #1350, opened 2021-12-16
add basic snyk action

johnduffell:
This PR adds a basic action to push issues into snyk

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,5 +26,5 @@ jobs:
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
-          args: --org=the-guardian-cuu --project-name=${{ github.repository }} --file=./frontend/yarn.lock
+          args: --org=the-guardian-cuu --project-name=${{ github.repository }} --file=./yarn.lock
           command: monitor
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,30 @@
+name: Snyk
+
+on:
+  push:
+    branches: [ main, jd-snyk-action ]
+  workflow_dispatch:
+
+jobs:
+  snyk:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      - name: SBT - Run Snyk to check for vulnerabilities
+        uses: snyk/actions/scala@0.3.0
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          args: --org=the-guardian-cuu --project-name=${{ github.repository }} --file=./build.sbt
+          command: monitor
+
+      - name: Yarn - Run Snyk to check for vulnerabilities
+        uses: snyk/actions/node@0.3.0
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,30 @@
+name: Snyk
+
+on:
+  push:
+    branches: [ main, jd-snyk-action ]
+  workflow_dispatch:
```
kelvin-chappell, line 6: We've added a schedule trigger as well to our workflows.  But I'm not sure if Snyk does scheduled runs anyway so could be redundant

.github/workflows/main.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+
+      - name: SBT - Run Snyk to check for vulnerabilities
+        uses: snyk/actions/scala@0.3.0
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+        with:
+          args: --org=the-guardian-cuu --project-name=${{ github.repository }} --file=./build.sbt
```
kelvin-chappell, line 21: I think the `file` argument is redundant as that's the default value for SBT

kelvin-chappell (approved):
👍

.github/workflows/main.yml
johnduffell: if you miss it out, does it not go off and find some other stuff?  I did have to add that to the promos one but maybe I was mistaken

.github/workflows/main.yml
johnduffell: it doesn't but is there a benefit to scheduled runs?  I would have thought they could only change if a commit is made?

kelvin-chappell: Discovery of a vulnerability in a dependency that's already in use?

.github/workflows/main.yml
johnduffell: "When omitted Snyk will try to detect the appropriate file for your project."

I think it can work if you only have a build.sbt but no other package.json etc, but I prefer to be explicit, I think in this case it won't work right.

https://docs.snyk.io/features/snyk-cli/guides-for-our-cli/cli-reference#:~:text=When%20omitted%20Snyk%20will%20try%20to%20detect%20the%20appropriate%20file%20for%20your%20project.

.github/workflows/main.yml
kelvin-chappell: Actually does that make sense? Presumably Snyk would just update its list so it is redundant.

.github/workflows/main.yml
kelvin-chappell: Perhaps it searches for build files if you don't tell it.  I'm not sure so ignore this comment.

.github/workflows/main.yml
johnduffell: yeah I think (although happy to be corrected) the monitor is just uploading the deps tree and I think the comparing with new vulns will happen regardless

johnduffell: It's not 100% clear but the docs suggest that's ok

> Use the snyk monitor CLI command to create a snapshot of a project on the Snyk website that will be continuously monitored for new vulnerabilities.

https://docs.snyk.io/introducing-snyk/snyks-core-concepts/running-tests#:~:text=Use%20the%20snyk%20monitor%20CLI%20command%20to%20create%20a%20snapshot%20of%20a%20project%20on%20the%20Snyk%20website%20that%20will%20be%20continuously%20monitored%20for%20new%20vulnerabilities.

PR #1353, opened 2021-12-21
Bump runtime lodash dep

TomPretty:
## What does this change?

Bump runtime lodash deps to fix https://github.com/guardian/subscriptions-frontend/security/dependabot/yarn.lock/lodash/open.

There were three runtime packages:

- `@emotion/core`
- `redux`
- `react-redux`

For the emotion I fixed the version in the `resolutions` field. `redux` and `redux-react` are unused, so I removed those dependencies.


GHaberis (approved):
👍

PR #1357, opened 2022-02-11
Fix review section on paper checkout

rupertbates:
## What does this change?
This fixes a bug we found whilst testing the newspaper checkout for the Waitrose campaign which was causing the applied promo code to be prepended to the delivery address in the review section.

Strangely this was being called by the `getGiftDetails()` function which seemed to be just picking up the promo code and returning it as though it was a gift recipient. 

We are not selling gifts through this site any more and are about to turn it all off anyway so I've just removed the function rather than spending too much time trying to work out what's going on.

### DIFF
--- a/assets/javascripts/modules/checkout/reviewDetails.js
+++ b/assets/javascripts/modules/checkout/reviewDetails.js
@@ -26,14 +26,6 @@ define([
         }
     }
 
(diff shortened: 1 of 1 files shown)


i-hardy (approved):
👍

PR #1358, opened 2022-02-14
Update the packshot image on the newspaper checkout

rupertbates:
## What does this change?
The packshot image on subscribe hasn't been updated for years so we're updating it for the Waitrose promotion.
Screen Shot 2022-02-14 at 11 42 31


marialani (approved):
👍

